PR #8, opened 2025-05-29
small fix for non-streaming responses

GitHappens2Me:
Just a small fix for non-streaming responses. 

A few days ago, I started getting Errors when sending requests with `"stream": false` to the proxy.
The nginx error message was: `upstream sent "Content-Length" and "Transfer-Encoding" headers at the same time`.
This was likely the result of some changes from 0b97839, but I am not sure what exactly caused the change in behaviour. 

Removing the header fixes the bug.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,7 @@ wallet.sqlite3
 
 # Development 
 .notes
-.keys.db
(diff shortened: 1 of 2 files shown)


shroominic (approved):
LGTM

PR #9, opened 2025-05-30
Added required dependency for coverage, improved gitignore and Test-Readme.md

GitHappens2Me:
`pytest-cov` was missing from pyproject.toml, so running `pytest --cov=router --cov-report=html` would fail.
 
Also, installing the test dependencies with `uv pip install -e ".[dev]"`, did not work. 
I tried `uv pip install --group dev`, which is usually used for the `[dependency-groups]`-property, and it works. 

I changed the command in the readme, but would ask you to test the new command on your machine as well. Maybe there are some differences in our setup. In this case, we should probably support both and adapt the readme accordingly.

### DIFF
--- a/router/proxy.py
+++ b/router/proxy.py
@@ -189,17 +189,10 @@ async def proxy(
                         key, response_json, session
                     )
                     response_json["cost"] = cost_data
-
-                    response_headers = dict(response.headers)
-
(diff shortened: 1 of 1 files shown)


GitHappens2Me:
Okey, thanks. I will look into that command.
Just to make sure, does that mean using this command, `pytest-cov` does not need to be specified in the `pyproject.toml`?

PR #10, opened 2025-06-02
Added the prepaid api key feature for self/testing.

routstrdev:
Now any provider can set up their own API key to use via Routstr instead of topping up with Cashu tokens all the time. 

Updated .env.example

Note: If you want a new prepaid api key with new balance, just change it in .env and restart the server.


shroominic:
the problem is we need to exclude this api key then from the balance calculations and usage tracking because currently it is summing up all user balances and substracting that from the cashu wallet to calculate payouts

PR #44, opened 2025-06-19
X cashu refund

shroominic:


9qeklajc:
maybe/not sure if it makes sense to also set the `X-Cashu` header when request failed:
* fail to redeem
* llm provider is down
* ...

basically always send `X-Cashu` header even with status code different than 200

PR #46, opened 2025-06-24
Fixed the internal server error for endpoints without a body

routstrdev:
Models endpoint and many endpoints without a body were failing. To reproduce: get /v1/models


9qeklajc:
fixed in other issue

Issue #49, opened 2025-06-26
/admin does not return the admin route / interface


Sebastix:
When I open http://localhost:8000/admin while the proxy is running (using Docker):

```json
{
  "detail": {
    "error": {
      "message": "API key or Cashu token required",
      "type": "invalid_request_error",
      "code": "missing_api_key"
    }
  }
}
```

It seems the URI is handled as a regular API request.

9qeklajc:
as a workaround u could add the / to the end of the url
http://localhost:8000/admin/

that should do

PR #50, opened 2025-06-26
#49 fix admin page navigation

9qeklajc:
solution is hacky. though, not other way to do it, except we move the endpoint to '/' instead of '/admin' and remove the info page


sh1ftred:
No needed anymore.

PR #60, opened 2025-06-30
X cash refac

9qeklajc:


shroominic (commented):
whats the difference between CostData | MaxCostData?
because form the class its the same
class MaxCostData(CostData):
    pass

so why this match?
case MaxCostData() as cost:
         return cost.dict()
case CostData() as cost:

because its a different class it matches different?

CURRENCY = cast(CurrencyUnit, os.environ.get("CURRENCY", "sat"))
not 100% sure but i think cast is not type validation but just overriding the type for type checking so mybe better Pydantic validation?

or cast just to get rid of the typing error

9qeklajc:
> whats the difference between CostData | MaxCostData? because form the class its the same class MaxCostData(CostData): pass
> 
> so why this match? case MaxCostData() as cost: return cost.dict() case CostData() as cost:
> 
> because its a different class it matches different?


it is basically a type check to move value check consern to different level, so the function where the classes are used, will not have to care about checking values to decide when and what to return ...

9qeklajc:
> CURRENCY = cast(CurrencyUnit, os.environ.get("CURRENCY", "sat")) not 100% sure but i think cast is not type validation but just overriding the type for type checking so mybe better Pydantic validation?
> 
> or cast just to get rid of the typing error


not realy validating in that case, I'm not sure if we need to validate (we could) .. just making sure that the user, that uses the redeem function know what to expect

shroominic (approved):
LGTM!

PR #79, opened 2025-07-26
add local relay for faster NIP-60

shroominic:


sh1ftred:
Not needed right now. Moved to a local wallet.

PR #92, opened 2025-08-04
Fix context length cost calculation

shroominic:
Fixes the context length cost calculation logic in the payment system.


sh1ftred (approved):
LGTM

PR #182, opened 2025-09-24
feat(wallet) added /deduct balance endpoint

abvstudio-net:
This pull request adds a /deduct endpoint which can be used to charge the user for web searches, vector-store searches or other costly / compute-intensive use-cases.  This will allow routstr to be better integrated into agentic frameworks like LangGraph.


shroominic:
Hey this implementation is currently a way to donate a partial balance to the roustr node provider, is your goal to use this as a hosted wallet so you can payout a partial amount in cashu tokens to pay for other services without implementing a cashu wallet yourself?

PR #233, opened 2025-11-18
Fixed false positives in SecurityFilter 

GitHappens2Me:
The regex in SecurityFilter was too broad, which resulted in redactions like:

`No token pricing configured, using base cost` -> `No token: [REDACTED] configured, using base cost`

This PR fixes the regex and cleans up the code a bit.
It also adds unit tests for the SecurityFilter, which can't hurt.


shroominic (approved):
LGTM! thanks

PR #234, opened 2025-11-18
PPQ.ai provider integration

shroominic:
image

easy account creation by just clicking a button:
image

easy topup over the dashboard:
image

account creation and topup can be implemented by using the same abstraction functions on other classes to achieve the same UI support


routstr/upstream/base.py
```diff
@@ -37,6 +38,17 @@
 logger = get_logger(__name__)
 
 
+class TopupData(BaseModel):
```
9qeklajc: it might be better to have a topup provider abstraction, so we keep the payment and the llm interaction separated ...

PR #239, opened 2025-11-22
gemini upstream

9qeklajc:


routstr/upstream/clients/base.py
```diff
@@ -0,0 +1,40 @@
...
+
+class BaseAPIClient(ABC):
+    """Base class for AI provider API clients."""
+
+    def __init__(self, api_key: str, base_url: str | None = None):
+        self.api_key = api_key
+        self.base_url = base_url
+
+    @abstractmethod
+    async def generate_content(
```
shroominic: i would maybe call it chat_completions and chat_completions_stream? so then we implement responses, embeddings, each endpoint seperatly?

PR #260, opened 2025-12-15
added reponses api

9qeklajc:
some refactoring for cleaner interface & added response endpoint


routstr/upstream/processing/http_client.py
```diff
@@ -0,0 +1,200 @@
...
+                    "url": url,
+                    "path": path,
+                    "query_params": dict(request.query_params),
+                    "traceback": tb,
+                },
+            )
+
+            raise
+
+
```
shroominic: proper cleanup got removed from the place needed, put here but then this part of the code is not used anywhere?

routstr/upstream/processing/http_client.py
```diff
@@ -0,0 +1,200 @@
...
+from ...core import get_logger
+
+if TYPE_CHECKING:
+    from ...payment.models import Model
+
+logger = get_logger(__name__)
+
+
+class HttpForwarder:
```
shroominic: i think this doesn't need to be a class, static function should be simpler and does the same

9qeklajc:
reopened https://github.com/Routstr/routstr-core/pull/278

PR #287, opened 2026-01-03
Disable provider discovery by default

shroominic:
Sets default `PROVIDERS_REFRESH_INTERVAL_SECONDS` to 0 and raises 404 in `get_providers` if it is 0, effectively disabling discovery by default.


9qeklajc:
why, do we need this ? the default behaviour is ok ?

shroominic:
> why, do we need this ? the default behaviour is ok ?

this setting should also prob be called different since the feature is more a help for clients so they dont need to discover nostr for providers, its not meaning the upstream provider but other routstr nodes

PR #360, opened 2026-02-11
Custom models were only showing up in the DB but now in the v1/models output

sh1ftred:


routstr/upstream/base.py
```diff
@@ -3023,18 +3021,45 @@ def _match_model(self, model_id: str, or_models: list[dict]) -> dict | None:
...
+                
+                # .first() returns the object or None if not found
+                provider = result.first()
+                if not provider or not provider.id:
+                    raise HTTPException(status_code=404, detail="Provider not found")
+
+                db_models = await list_models(
+                    session=session,
+                    upstream_id=provider.id,
+                    include_disabled=True,
```
9qeklajc: should this be False ?

9qeklajc:
LGTM

PR #389, opened 2026-03-03
UI update v0.4.0 + analytics sharing

fanyiy:


9qeklajc:
image

unit doesn't match and pirces are too high

PR #489, opened 2026-05-06
fix: refund users when model returns zero tokens with non-zero cost

sh1ftred:
## Problem

When a model produces empty output (zero tokens reported) but the cost calculation returns a non-zero USD cost, the previous behavior was to charge the full amount as `output_msats`. This means users were being billed for responses where the model produced nothing.

## Fix

In the `else` branch of `calculate_cost()` (when no tokens are reported):
- **Before:** `output_msats = cost_in_msats` (user gets charged the full amount)
- **After:** Set `input_msats = 0`, `output_msats = 0`, `cost_in_msats = 0` — issuing a **full refund** to the user
- Added a `logger.warning` with `usd_cost` and `model` details for debugging/monitoring

## File Changed

- `routstr/payment/cost_calculation.py`

## Testing

The change is a simple conditional branch update. When `usage.input_tokens` and `usage.output_tokens` are both 0 but `usd_cost > 0`, the user now receives a full refund instead of being charged.

### DIFF
--- a/routstr/payment/cost_calculation.py
+++ b/routstr/payment/cost_calculation.py
@@ -180,7 +180,17 @@ async def calculate_cost(  # todo: can be sync
                     input_msats = int(cost_in_msats * input_ratio)
                     output_msats = cost_in_msats - input_msats
                 else:
-                    output_msats = cost_in_msats
+                    # No tokens reported — model produced empty output, issue full refund
+                    logger.warning(
+                        "Zero tokens with non-zero USD cost — issuing full refund",
+                        extra={
+                            "usd_cost": usd_cost,
+                            "model": response_data.get("model", "unknown"),
+                        },
+                    )
+                    input_msats = 0
+                    output_msats = 0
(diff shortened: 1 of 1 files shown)


9qeklajc:
some upstream will have the problem of not setting the token count for some reasons even when the user gets a response which in that case implies that the request is free

shroominic:
Heads up: I opened a clean re-implementation of this fix against current `main` — **https://github.com/Routstr/routstr-core/pull/561**. This PR is 110 commits behind a refactored `main` (conflicting) and ships no tests; the new one re-applies the fix to the current `_calculate_from_usd_cost` code, **tightens the gate** so legitimately-billed cache-read-only turns aren't refunded (this PR refunds whenever prompt+completion==0, which over-fires on cache-only turns), and adds regression tests (truly-empty → refunded, verified fail-without-fix; cache-only → still billed). Suggest closing this in favour of https://github.com/Routstr/routstr-core/pull/561 once reviewed.

PR #514, opened 2026-05-17
add provider slug

9qeklajc:
This helps agents using routstr-cli update providers by stable slug as well as numeric ID.

### DIFF
--- a/migrations/versions/c6d7e8f9a0b1_add_slug_to_upstream_providers.py
+++ b/migrations/versions/c6d7e8f9a0b1_add_slug_to_upstream_providers.py
@@ -1,0 +1,55 @@
+"""add slug to upstream_providers
+
+Revision ID: c6d7e8f9a0b1
+Revises: b5e7c9d1f3a2
+Create Date: 2026-06-29 00:00:00.000000
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "c6d7e8f9a0b1"
+down_revision = "b5e7c9d1f3a2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+    columns = {c["name"] for c in inspector.get_columns("upstream_providers")}
+
+    if "slug" not in columns:
+        op.add_column(
+            "upstream_providers",
+            sa.Column("slug", sa.String(), nullable=True),
+        )
+
+    op.execute(
+        "UPDATE upstream_providers "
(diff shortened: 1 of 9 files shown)


jeroenubbink:
Nice work! This makes it a lot easier than working with ID's.

2 suggestions:

1. The PATCH to /api/upstream-providers already exists here: https://github.com/Routstr/routstr-core/blob/a16bc1220cbbe7ebe40bb31abbb420c3467880ab/routstr/core/admin.py#L815-L817. I suggest to make all endpoints that now take a provider id also take the identifier.
2. I would rename "identifier" to "slug" which is a more conventional name for these things and a well known term for api's and other web applications.

jeroenubbink:
Nice — this is a genuinely useful handle to have, and I'd like to build on it, so keen to see it land. One gap worth closing first:

The slug is only generated on the admin create path (`create_upstream_provider`). Providers seeded from settings/env go through `_seed_providers_from_settings` in `routstr/upstream/helpers.py`, which constructs `UpstreamProviderRow(...)` directly (around lines 280 / 300 / 321 / 343) and isn't touched here — so those rows land with `slug = NULL`.

That means the providers most single-node operators actually run (one env-configured upstream) can't be addressed by slug at all: on a fresh install the migration backfill runs before any rows exist, then seeding inserts NULL-slug rows; on an existing install the one-time backfill covers current rows but any newly-added settings provider still seeds as NULL. The unique index tolerates the NULLs (SQLite allows multiple), so nothing crashes — the feature just silently doesn't cover the seed path.

The fix wants to be **deterministic**, which is the crux: the admin path's `secrets.token_hex(3)` suffix can't be reused in the seed path, because a random slug regenerated per boot would either churn or defeat the `(base_url, api_key)` dedup. A slug derived deterministically from the provider (e.g. `provider_type`, then `provider_type-2`/`-3` for collisions at a base_url) stays stable across re-seeds and gives seeded providers a predictable, scriptable handle — which is the whole point of the feature. Same derivation could back the admin default too, so a hand-created `openai` provider reads as `openai` rather than `openai-a1b2c3`.

Happy to help wire the seed path through the same generator if useful. Either way, would love to get this merged — it's a clean addition and a couple of us are waiting to build on top of it.

jeroenubbink (approved):
LGTM — this fully addresses both my earlier comments: seeded providers now get deterministic, reseed-stable slugs (no more `slug = NULL` on the seed path), the random suffix is gone in favour of the same generator everywhere, and every provider endpoint accepts id-or-slug while staying back-compat with numeric IDs. Clean addition, keen to build on it. 🚀

PR #516, opened 2026-05-18
fix: SSE parsing, encoding fixes, and tee GET passthrough

sh1ftred:
### SSE parsing rewrite (4a8d0e1)
- Replace regex-based SSE splitting with proper buffered `\n\n` delimiter parsing
- Fix [DONE] detection to match `data: [DONE]` instead of bare `[DONE]`
- Add debug logging for SSE buffer state, parsed events, and stream lifecycle
- Distinguish billing-only (usage) chunks from content-bearing chunks; hold back usage-only chunks for later cost metadata injection
- Extract JSON payload from `data: ...` prefix explicitly
- Gracefully pass through non-JSON SSE events

### Encoding fix (4406a9e)
- Fix encoding bug in upstream response handling

### Tee GET passthrough (e6cfdb3)
- Pass through tee GET requests properly

### DIFF
--- a/routstr/proxy.py
+++ b/routstr/proxy.py
@@ -162,6 +162,7 @@ _API_PATH_PREFIXES = (
     "images/",
     "moderations",
     "providers",
+    "tee/",
 )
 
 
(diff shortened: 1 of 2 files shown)


sh1ftred:
Splitting into separate PRs: tee fixes and SSE refactor

PR #519, opened 2026-05-19
refactor: rewrite SSE parsing with buffered double-newline delimiter

### LINKED ISSUE #505: Gemini models fail when upstream is Gemini (Invalid JSON payload: Unknown name 'store')
When using Gemini models and the upstream is also Gemini, the request fails with the following error:

```json
[
  {
    "error": {
      "code": 400,
      "message": "Invalid JSON payload received. Unknown name \"store\": Cannot find field.",
      "status": "INVALID_ARGUMENT",
      "details": [
        {
          "@type": "type.googleapis.com/google.rpc.BadRequest",
          "fieldViolations": [
            {
              "description": "Invalid JSON payload received. Unknown name \"store\": Cannot find field."
            }
          ]
        }
      ]
    }
  }
]
```

It appears that a `store` parameter is being sent which the Gemini API doesn't recognize.

sh1ftred:
## Changes

- Replace regex splitting on `data: ` with proper SSE buffering using `\n\n` as the event separator, correctly handling partial chunks across transport boundaries
- Fix `[DONE]` detection to match `data: [DONE]` instead of bare `[DONE]`
- Add debug logging for SSE buffer state, parsed events, and stream lifecycle
- Distinguish billing-only (usage) chunks from content-bearing chunks; hold back usage-only chunks for later cost metadata injection
- Extract JSON payload from `data: ...` prefix explicitly rather than relying on raw part parsing
- Gracefully pass through non-JSON SSE events instead of dropping them

### DIFF
--- a/routstr/upstream/base.py
+++ b/routstr/upstream/base.py
@@ -701,54 +701,100 @@ class BaseUpstreamProvider:
                         pass
 
             try:
+                sse_buffer = b""
                 async for chunk in response.aiter_bytes():
-                    # Split chunk into SSE events
-                    parts = re.split(b"data: ", chunk)
-                    for i, part in enumerate(parts):
-                        if not part:
+                    sse_buffer += chunk
+                    logger.debug(
+                        "[chat] SSE chunk from upstream",
+                        extra={"chunk_size": len(chunk), "buffer_size": len(sse_buffer)},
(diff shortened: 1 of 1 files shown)


9qeklajc:
build failed

PR #525, opened 2026-05-25
Feature: Web Search / RAG Integration for LLM Context Enhancement

GitHappens2Me:
### Overview
This PR adds a complete **Retrieval-Augmented Generation** (RAG) pipeline to Routstr, enabling AI requests to be enhanced with real-time web search context. When a user (or their client) sends a request with the HTTP request parameter `enable_web_search` set to `true`, the system performs a web search, retrieves relevant content and injects it as context into the LLM prompt, before forwarding the request to the upstream provider. 

### Why This Is Useful
- All models can gain access to current knowledge regardless of provider, even if the provider does not offer RAG natively.
- Users gain more options. The RAG system in this PR is completely optional but can be used instead of the upstream provider's system.
- Routstr operators can choose which RAG-implementations to offer. Currently, Tavily and Exa (both all-in-one paid services) and a customizable pipeline are supported.

### Architecture
Except for some integrations in, for example, `base.py` and `cost_calculation`, all of the web search logic is located in the `routstr-core/routstr/websearch` folder. The only module which interacts with the existing code is the `web_manager.py`. This handles the initialization (depending on the settings) and the orchestration of the web search itself. The architecture heavily uses abstract base classes to allow many different alternative implementations without code duplication. A complete module overview is below:
```
routstr/websearch/
├── types.py                 # Data types: WebPage, SearchResult, WebSearchContext, provider enums
├── web_manager.py           # Orchestrator: provider init and caching, query extraction, content injection
├── http_client.py           # Shared async HTTP client with connection pooling
├── mock_utils.py            # Debug utilities for saving/loading API responses
│
├── base_web_rag.py          # ABC for all RAG providers
├── tavily_web_rag.py        # Tavily API implementation
├── exa_web_rag.py           # Exa API implementation
├── custom_web_rag.py        # Customizable multi-step pipeline
│
├── base_web_searcher.py     # ABC for search providers (necessary for custom_web_rag)
├── serper_web_searcher.py   # Serper (Google) search implementation
│
├── base_web_scraper.py      # ABC + orchestrator for concurrent scraping (necessary for custom_web_rag)
├── http_web_scraper.py      # HTTP + Trafilatura content extraction
│
├── base_chunker.py          # ABC for text chunking (necessary for custom_web_rag)
├── fixed_size_chunker.py    # Fixed-size character chunking with overlap
├── recursive_chunker.py     # Semantic recursive chunking (paragraphs → sentences → words)
│
├── base_ranker.py           # ABC for result ranking (necessary for custom_web_rag)
└── bm25_ranker.py           # BM25 relevance ranking with local + global pruning
```

### "All-in-one Providers" vs. Custom Pipeline
The simplest setup is to create an account with one of the two providers currently supported ([Tavily](https://www.tavily.com/) and [Exa](https://exa.ai/)). Then the feature can be activated by setting `WEB_RAG_PROVIDER` to the chosen provider and providing its API key. For example: 
```env
WEB_RAG_PROVIDER=tavily
TAVILY_API_KEY=tvly-xxxxx
```
For more control, the custom pipeline can be activated instead, which requires Routstr operators to choose an implementation for the four RAG stages (Searching, Scraping, Chunking, Ranking). Sensible defaults are available. The custom pipeline does most work locally, except for the searching stage which currently relies on the service [Serper](https://serper.dev/):
```env
WEB_RAG_PROVIDER=custom
WEB_SEARCH_PROVIDER=serper
SERPER_API_KEY=xxxxx
WEB_SCRAPER_PROVIDER=http       (simple async HTTP scraping + extraction using trafilatura)
WEB_CHUNKER_PROVIDER=recursive  (recursive chunking)
WEB_RANKING_PROVIDER=bm25       (BM25-Ranking Algorithm using rank-bm25)
```
By default web searches are disabled:
```env
WEB_RAG_PROVIDER=disabled
```

### 1. Request Flow

When a client sends a request with `"enable_web_search": true`, the `web_manager.py` module does the following:

1. **`extract_web_search_parameter()`** strips the flag from the request body so it never reaches the upstream provider
2. **`extract_query_from_request_body()`** pulls the last user message as the search query
3. **`retrieve_context()`** calls the configured RAG provider
4. **`generate_xml_context()`** formats results into structured XML including a list of sources
5. **`inject_web_context_into_request()`** inserts the XML as a system message just before the last user message. The XML also includes instructions for the LLM, on how to use the information.
6. The enhanced request is forwarded to the upstream provider, where inference is done

Any errors during this process are caught gracefully and logged, before returning control back to the original code, so the inference request to the upstream provider is still sent.  

### 2. Cost Integration

Web search cost flows through the existing cost calculation system:

- **`web_search_msats`** is added to the response cost when `web_search_executed` is true
- Cost is either a fixed amount (setting: `WEB_SEARCH_FIXED_COST`) or model-specific (`model.sats_pricing.web_search`)
- The `WebSearchContext` object carries an `executed`-flag and optional `sources` through the entire request pipeline
- Both streaming and non-streaming responses include `web_search_executed` and `sources` metadata. The web search cost is returned alongside the inference cost. The sources are returned so any frontend can choose to display them next to the response. 

### Unit Tests

I also added unit and integration tests for the `web_manager.py` as the main orchestrator and to test the web search feature with all possible request paths (streaming, non-streaming, Anthropic's format etc.).

---

As this is a large and opinionated PR, I welcome any questions, discussion or general feedback. 
In the meantime, one can test this feature at https://chat.githappens.space/, where I added a simple toggle to perform a web search for a given query. Currently web searches are performed using Tavily's service.

I am opening this as a Draft PR as I am sure there will be many suggestions, further questions and discussions before merging would make sense.


sh1ftred:
Hey, wouldn't it be better to have this on the client side rather than on Routstr node?

9qeklajc:
@GitHappens2Me thnx for the PR... the diff is currently too big as the linter and formatter didn't run ..

I'm also not sure if this not better be an extension that client can use to do the web search and modify the extracted data (enhance, filter, etc.. ) then do a request to the node as there a lot more parameter that a client can configure to enhance the search, by setting the number of web searches, setting the urls , etc .. which makes it more complex and not easy to maintain .. 

or can u explain ur use case

GitHappens2Me:
Thanks for looking into the PR and giving feedback. 

The original idea was to add RAG to the proxy so **one** account/subscription/access etc. to a RAG provider can be used by multiple users of that proxy. This removes complexity from users, as they can just rely on the Routstr operator to configure the search, who in turn gets rewarded for offering this service with additional fees. Using this approach would also anonymize the user's search queries as they can only be traced back to the Routstr server (not the individual user), as is already the case with inference requests.

But you raise valid points. This PR adds a lot of additional complexity (and code to maintain), which might not be desirable for a feature that might not actually be used / offered by many Routstr operators.

The biggest downsides, I see, of moving the RAG logic to the client are:
- Every user needs to provide their own access to a RAG API service.
- Users loose the privacy benefits of proxying the web requests through Routstr, which I believe undermines one of Routstr's advantages.

@9qeklajc  What would an extension to the client look like? Do you mean moving this PR's logic to the routstr-chat repo?

9qeklajc:
@GitHappens2Me I understand and I agree 100% about this
> Users loose the privacy benefits of proxying the web requests through Routstr, which I believe undermines one of Routstr's advantages.

For now, it makes sense that the client makes use of the existing provider with an OpenAI-compatible API, as they support setting a request parameter to trigger the web search — for example, here: https://developers.openai.com/api/docs/guides/tools-web-search — or just use Perplexity models to do that.
I'm also not really sure if this still has a use case other than the one you mentioned above about privacy, since agents are now widely used and have their own tools to get the data they want.

GitHappens2Me:
Thanks again for the feedback. I totally see your points! 
While I still believe there's value in the privacy aspect of this PR, I agree the added maintenance burden and complexity don't justify merging this into Routstr right now.
I appreciate the discussion though, happy to close this.

PR #527, opened 2026-05-28
Fix/fix dockerfile and secp256k1 git repo

jeroenubbink:
Properly fixes a couple of issues with secpk256k1 python package and properly builds a docker image from `Dockerfile` again.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,6 @@
 FROM ghcr.io/astral-sh/uv:python3.11-alpine
 
-# Install system dependencies required for secp256k1
+# Install system dependencies
 RUN apk add --no-cache \
     pkgconf \
     build-base \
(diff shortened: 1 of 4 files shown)


9qeklajc:
thx for the PR..
after thorough search I think we do not even need this dep. anymore, was previously added to add support to nip-60 wallet which was dropped later and replaced with a local wallet impl.... so dropping it would be safe
https://github.com/Routstr/routstr-core/pull/530

jeroenubbink:
the nostr package requires this.

```
$ uv pip show secp256k1
Name: secp256k1
Version: 0.14.0
Location: /home/jeroen/dev/routstr-core/.venv/lib/python3.11/site-packages
Requires: cffi
Required-by: nostr, routstr
```

9qeklajc:
right but only nostr needs it and already declares it.. shouldn't be declared twice

```
$ uv pip show secp256k1
Name: secp256k1
Version: 0.14.0
Location: /Users/molk/projects/routstr-core/.venv/lib/python3.11/site-packages
Requires: cffi
Required-by: nostr
```

jeroenubbink:
> right but only nostr needs it and already declares it.. shouldn't be declared twice
> 
> ```
> $ uv pip show secp256k1
> Name: secp256k1
> Version: 0.14.0
> Location: /Users/molk/projects/routstr-core/.venv/lib/python3.11/site-packages
> Requires: cffi
> Required-by: nostr
> ```

Yes, but it needs binary code and the pypi release only ships for a few libc versions and architectures, there is no version for musl published so it needs to be compiled when building the docker file to run with alpine, which uses musl and not glibc.

Hence we need to provide it with the source so it can be compiled during docker build.

jeroenubbink:
i see you changed the docker image to one with glibc, that works as well. I'll close the PR

PR #528, opened 2026-05-28
Remove deprecated cashu-refund-by-hash endpoint

bilthon:
### What this does

Removes a refund-recovery endpoint that has never actually worked, plus the dead code path that fed it.

### Why

While poking at the x-cashu flow locally, I sent a paid request without `-i`, lost track of the `X-Cashu` response header, and went looking for the recovery endpoint hinted at in `routstr/balance.py`. The endpoint exists:

```python
@router.get("/cashu-refund/{payment_token_hash}")
async def get_cashu_refund(payment_token_hash: str, ...):
    result = await session.get(CashuTransaction, payment_token_hash)
    if result is None:
        raise HTTPException(404, "Refund not found")
```

It looks up `CashuTransaction` by primary key, where the key is meant to be a SHA-256 hash of the original payment token. The trouble is that `CashuTransaction.id` is a random UUID (`routstr/core/db.py:141`), and the hash is never stored on the row, in any column. So the lookup cannot match, ever. The endpoint returns 404 for every valid input, which is what I hit.

The hash itself is computed at two places in `routstr/upstream/base.py` and threaded through about eleven function signatures and roughly thirty four call sites, but nothing else consumes it. It is pure dead weight.

A working alternative already exists right next door at `routstr/balance.py:246`:

```
POST /v1/wallet/refund
x-cashu: <original-token>
```

That handler finds the inbound `type='in'` row by matching the token string, follows the `request_id` to the corresponding `type='out'` row, and returns the refund token. It is also a bit better on the privacy side (token in a header rather than a hash in the URL path), and it does not need a schema change to keep working.

Given that, the cleanest fix is to delete the broken endpoint and the parameter that fed it, and to point users at the endpoint that does work. No schema migration, no behavioural change to anything that was working before.

Tests and docs are updated alongside the code. Five unit test invocations that passed `payment_token_hash=` as a kwarg were trimmed so the suite still compiles and passes, and `docs/api/overview.md` gains a single sentence pointing readers at `POST /v1/wallet/refund` as the recovery path when the `X-Cashu` response header is missed.

Verified locally: make lint, make type-check, make test-unit all green

### DIFF
--- a/docs/api/overview.md
+++ b/docs/api/overview.md
@@ -155,7 +155,7 @@ The response includes your change in the same header:
 X-Cashu: cashuA7k2mNp4...
 ```
 
-This is fully stateless—no session, no `/v1/balance/refund` call needed. However, **streaming does not work with `X-Cashu`** because the refund can only be calculated after the full response is generated.
+This is fully stateless—no session, no `/v1/balance/refund` call needed. However, **streaming does not work with `X-Cashu`** because the refund can only be calculated after the full response is generated. If you lose the `X-Cashu` response header before claiming your change, you can reclaim the refund via `POST /v1/wallet/refund` by supplying the original payment token in the `x-cashu` header.
 
 ## Response Headers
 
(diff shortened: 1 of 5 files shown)


9qeklajc:
nice, will take a look into it

9qeklajc:
LGTM

PR #540, opened 2026-06-03
fix: make child key balance deduction atomic

jeroenubbink:
The previous in-memory deduction (key.balance -= cost) was a read-modify- write on stale state, allowing two concurrent create_child_key() calls to both pass the balance check and both succeed, effectively charging the parent only once for two child keys.

Replace with an atomic UPDATE ... WHERE balance - reserved_balance >= cost and check rowcount, matching the pattern already used in pay_for_request. Also adds a concurrent integration test that reproduces the race and confirms the fix holds.


9qeklajc:
LGTM

PR #541, opened 2026-06-03
fix: persist and propagate key constraints from Lightning invoices

jeroenubbink:
LightningInvoice had no columns for balance_limit, balance_limit_reset, or validity_date. SQLModel silently dropped these constructor kwargs, so create_api_key_from_invoice always produced an unconstrained key.

Add the three columns to LightningInvoice with a migration, and wire them through to the ApiKey in create_api_key_from_invoice, matching the pattern already used in the child key creation path.


9qeklajc:
LGTM

PR #547, opened 2026-06-09
fix: enforce balance_limit atomically on child key requests

jeroenubbink:
The Python pre-check on balance_limit provided no concurrency guarantee — two concurrent requests could both pass the check on stale in-memory state and both proceed to reserve, exceeding the limit.

Add the balance_limit guard to the child key UPDATE's WHERE clause so the enforcement is atomic. The HTTP 402 is raised without an explicit rollback: since session.commit() was never called, the uncommitted billing key update is discarded when the session context manager closes on exception exit.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -545,6 +545,28 @@ async def pay_for_request(
     )
     result = await session.exec(stmt)  # type: ignore[call-overload]
 
+    if result.rowcount == 0:
+        logger.error(
+            "Concurrent request depleted balance",
+            extra={
+                "key_hash": key.hashed_key[:8] + "...",
+                "billing_key_hash": billing_key.hashed_key[:8] + "...",
+                "required_cost": cost_per_request,
+                "current_balance": billing_key.balance,
+            },
+        )
+
+        raise HTTPException(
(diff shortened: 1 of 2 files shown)


jeroenubbink:
had this one still lying around.

tests/integration/test_key_logic.py
```diff
@@ -120,6 +122,56 @@ async def test_periodic_key_reset_job(integration_session: AsyncSession) -> None
     assert key2.total_spent == 0
 
 
+@pytest.mark.asyncio
+async def test_balance_limit_enforced_atomically_under_concurrency(
```
9qeklajc, line 126: the more tests the better, can u add some tests to test every case, for instance:
* maybe some tests to make parallel payment with parent and child key and run them in parallel 
* different tests with total_spend set, with reserve_balance set (just in case)

routstr/auth.py
9qeklajc: I think it makes sense that this block be moved above, so when it fails it just returns without executing the child key logic

9qeklajc:
solid PR

9qeklajc:
LGTM

PR #549, opened 2026-06-11
fix: retry foreign mint swaps with observed fees instead of trusting fee_reserve

### LINKED ISSUE #468: When a cashu token of an unsupported mint is sent to the nodes, it sometimes leads to this error.
`{"detail":{"error":{"message":"Invalid or expired Cashu key: Failed to melt token from foreign mint https://mint.cubabitcoin.org: could not pay invoice: Mint Error: not enough inputs provided for melt. Provided: 179, needed: 180 (Code: 11000)","type":"invalid_request_error","code":"invalid_api_key"}},"request_id":"34db7574-f00a-48fc-a25c-a65bb6e871ae"}`

jeroenubbink:
## Problem

A foreign mint's melt `fee_reserve` is a non-binding estimate (NUT-05): the mint may demand more when re-quoting or at melt execution. `swap_to_primary_mint` had zero tolerance for this, so topups failed outright when a mint under-quoted — observed in production with mint.cubabitcoin.org ("provided 179, needed 180", #468). The same swap path serves both the topup endpoint and cashu tokens used directly as API keys, which is where #468 surfaced; the verbatim production error string (including cashu-py's "could not pay invoice" wrapper) is pinned as retryable in the classifier tests.

Closes #468

## Approach: reactive retry, no safety buffer

Padding the estimate with a buffer (flat or percentage) was considered and rejected: any margin is a guess parameter and gets stranded at the foreign mint on every swap, including at honest mints. Instead the mint-quote/melt-quote/melt cycle retries (max 3 attempts) with the amount recomputed from the fees the mint actually demands. Honest mints pay no cost: the happy path is unchanged (two quotes, full net amount credited).

Melt failures are classified by the NUT-00 error code, parsed from cashu-py's stable `(Code: NNNN)` suffix:
- **11005** `TransactionUnbalanced` (registered, sent by cdk) and **11000** (nutshell's generic `TransactionError`) → retry with a smaller invoice
- anything else (e.g. **20004** Lightning payment failed, **11001** already spent) → surface immediately; a smaller invoice would not help and the payment state may be unknown

No NUT standardizes amounts in error responses, so nutshell's `Provided: X, needed: Y` text only refines the retry step when present; otherwise the amount shrinks one unit at a time.

## Tests

- Unit + integration tests written first (TDD); the retry scenarios reproduce the cubabitcoin incident exactly, plus a cdk-style rejection
- Every statement and branch of the topup path is covered, including the previously untested post-melt mint-failure/recovery block (recovery must never credit proofs the wallet does not hold; post-melt failures propagate unwrapped since the foreign proofs are already spent)
- Mocks sit only at the external mint boundary (the cashu `Wallet`); orchestration is tested through the public functions, pure leaf helpers (error classifier, unit conversion) directly
- `make test`: 243 passed, `make lint`: ruff + mypy clean

### DIFF
--- a/routstr/wallet.py
+++ b/routstr/wallet.py
@@ -127,10 +127,12 @@ _MAX_SWAP_ATTEMPTS = 3
 _MINT_ERROR_CODE_RE = re.compile(r"\(Code: (\d+)\)")
 _MELT_SHORTFALL_RE = re.compile(r"Provided: (\d+), needed: (\d+)")
 
-# Insufficient melt inputs across implementations: 11005 is the registered
-# "Transaction is not balanced" code (used by cdk); 11000 is nutshell's
-# generic TransactionError, which is unregistered but what nutshell sends.
-_RETRYABLE_MELT_CODES = frozenset({"11000", "11005"})
+# Insufficient-melt-inputs failures differ across mint implementations. 11005 is
+# the registered "Transaction is not balanced" code (cdk), specific enough to
+# trust on the code alone. 11000 is nutshell's generic, unregistered
+# TransactionError covering many unrelated failures, so it only counts as a fee
+# shortfall alongside the "not enough inputs" detail text. With no code suffix at
+# all, that same text is the only signal.
 
 
 def _net_minted_amount(amount_msat: int, token_unit: str, fees: int) -> int:
@@ -160,11 +162,19 @@ def _melt_insufficient_shortfall(error: Exception) -> int | None:
     """
     message = str(error)
     code_match = _MINT_ERROR_CODE_RE.search(message)
-    if code_match is not None:
-        if code_match.group(1) not in _RETRYABLE_MELT_CODES:
+    code = code_match.group(1) if code_match is not None else None
+    has_shortfall_text = "not enough inputs" in message.lower()
+
+    match code:
+        case "11005":  # registered TransactionUnbalanced: trust the code
+            pass
+        case "11000" if has_shortfall_text:  # generic nutshell error: needs the text
+            pass
+        case None if has_shortfall_text:  # no code suffix: text is the only signal
+            pass
+        case _:  # other codes, a bare 11000, or no signal: must not retry
             return None
-    elif "not enough inputs" not in message.lower():
-        return None
+
     amounts = _MELT_SHORTFALL_RE.search(message)
     if amounts is not None:
         provided, needed = int(amounts.group(1)), int(amounts.group(2))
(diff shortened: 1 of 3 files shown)


jeroenubbink:
@9qeklajc  i noticed that make lint runs mypy only on source code while the backend pipelines run mypy on everything (including tests) by running uv run mypy. Personally i do not care so much about mypy in tests, so i am ok if we drop this from the pipeline and use make lint there as well.

9qeklajc:
@jeroenubbink maybe it makes sense to update make to do the same as the build server, I know tests are not really important in terms of types and style but I'd rather enforce it then drop it

jeroenubbink:
converting to draft to do some manual testing.

9qeklajc:
Looks good overall. I left one small concern around treating any Code 11000 melt error as retryable — might be worth tightening that classifier or adding a regression test before merge.

9qeklajc:
> Looks good overall. I left one small concern around treating any Code 11000 melt error as retryable — might be worth tightening that classifier or adding a regression test before merge.

this ?

shroominic (approved):
LGTM ✅ — verified by building the branch and running the suite, plus a real 3-way merge into current `main`.

- **Fixes #468:** the retry loop is load-bearing — neutralizing `_melt_insufficient_shortfall` reproduces the exact `not enough inputs provided for melt. Provided: 179, needed: 180 (Code: 11000)` failure. `tests/unit/test_wallet.py` → 44 passed; `tests/integration/test_swap_fee_retry.py` → 3 passed.
- **Conservatively scoped classifier:** only fee-shortfall (`11000`/`11005`, "not enough inputs") retries; spent-proof (`11001`), Lightning (`20004`), and timeouts never retry.
- **No #556 regression on merge:** this branch predates #556, so in isolation `wallet.py` *looks* like the pre-fee-deduction version — but a real 3-way merge into current `main` is clean and **preserves** `_redeem_same_mint` (`amount - input_fees`) and the `997` guard test. Merged tree: 341 passed, 1 skipped; ruff + mypy clean.

One operational note: please land this via a normal 3-way **merge/rebase**, not by overwriting `main`'s `wallet.py` wholesale (that would drop the #556 deduction).

jeroenubbink:
> > Looks good overall. I left one small concern around treating any Code 11000 melt error as retryable — might be worth tightening that classifier or adding a regression test before merge.
> 
> this ?

fixed now, made it less generic for nutshell's 11000

9qeklajc:
LGTM

PR #550, opened 2026-06-11
fix: bill cached input tokens at their real rates across vendor dialects

jeroenubbink:
## Problem

Cached prompt tokens were billed at the full input rate. For a DeepSeek upstream this overcharges heavily: DeepSeek's caching is automatic and always on, cache hits cost 10x less than misses upstream, and agentic/multi-turn workloads run at 80–95% cache hit rates — a 10k-token prompt with 90% hits was charged 11,000 msats where honest pricing is 2,900. OpenAI cached reads were silently mispriced the same way, and Anthropic cache writes (1.25x input) were *under*charged.

## Root causes

1. **Unparsed usage dialect** — DeepSeek reports `prompt_cache_hit_tokens`/`prompt_cache_miss_tokens` (no OpenAI-style `prompt_tokens_details`), which billing never read, so every prompt token billed as regular input.
2. **Missing cache rates** — model pricing comes from OpenRouter's feed, which omits `input_cache_read`/`input_cache_write` for most DeepSeek models (and e.g. `openai/gpt-4o`); billing then fell back to the full input rate.

## Changes

- **`payment/usage.py`** — canonical `NormalizedUsage` + a union parser over the known, non-colliding dialects (OpenAI details, Anthropic additive fields, DeepSeek hit/miss), with single-subtraction precedence. Exposed as an overridable `BaseUpstreamProvider.normalize_usage` hook (the escape hatch for future vendors whose fields genuinely conflict); every settlement call site passes the provider's result through, so `calculate_cost` holds **zero vendor knowledge**. This keeps vendor dialects inside `upstream/` — adding a new vendor never touches `payment/`.
- **`backfill_cache_pricing`** — missing cache rates are filled from litellm's bundled cost map (already a dependency) before the provider fee applies; OpenRouter-provided rates stay authoritative; the input-rate fallback remains only as the documented last resort (never undercharges).
- **`refactor:` commit** — drops the `session` parameter from `calculate_cost`: needed when pricing lived in the DB (73d3613), dead since the in-memory model map (0da08fb). Also removes the per-request DB session `get_x_cashu_cost` opened solely to feed it. Separately revertable if it raises eyebrows.
- **`build:` commit** — `make lint`/`type-check`/`ci-lint` now run `mypy .` matching CI, so test files can't pass locally and fail the pipeline.

## Tests

36 cache-billing tests written first (TDD): verbatim DeepSeek payloads from their API reference, dialect precedence/double-subtraction guards, msats-level assertions for the headline scenario (2,900 vs 11,000), Anthropic write premium, litellm backfill (asserted dynamically against litellm's map so version bumps don't break), provider-fee application to backfilled rates, and an `ArtificialDumbnessProvider` proving a fictional vendor's dialect flows through billing without generic code knowing its field names. Full suite green, `mypy .` clean.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -772,7 +772,7 @@ async def adjust_payment_for_tokens(
                     extra={"error": str(e), "fee_msats": fee_msats},
                 )
 
-    match await calculate_cost(response_data, deducted_max_cost, session, usage=usage):
+    match await calculate_cost(response_data, deducted_max_cost, usage=usage):
         case MaxCostData() as cost:
             logger.debug(
                 "Using max cost data (no token adjustment)",
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -20,6 +20,7 @@ from .payment.cost_calculation import (
     MaxCostData,
     calculate_cost,
 )
+from .payment.usage import NormalizedUsage
 from .wallet import credit_balance, deserialize_token_from_string
 
 logger = get_logger(__name__)
@@ -675,12 +676,20 @@ async def revert_pay_for_request(
 
 
 async def adjust_payment_for_tokens(
-    key: ApiKey, response_data: dict, session: AsyncSession, deducted_max_cost: int
+    key: ApiKey,
+    response_data: dict,
+    session: AsyncSession,
+    deducted_max_cost: int,
+    usage: NormalizedUsage | None = None,
 ) -> dict:
     """
     Adjusts the payment based on token usage in the response.
     This is called after the initial payment and the upstream request is complete.
     Returns cost data to be included in the response.
+
+    ``usage`` carries the upstream provider's normalized token usage (its
+    ``normalize_usage`` hook); when omitted, the response's usage object is
+    normalized with the default union parser.
     """
     billing_key = await get_billing_key(key, session)
     model = response_data.get("model", "unknown")
@@ -763,7 +772,7 @@ async def adjust_payment_for_tokens(
                     extra={"error": str(e), "fee_msats": fee_msats},
                 )
 
-    match await calculate_cost(response_data, deducted_max_cost, session):
+    match await calculate_cost(response_data, deducted_max_cost, session, usage=usage):
         case MaxCostData() as cost:
             logger.debug(
                 "Using max cost data (no token adjustment)",
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -20,7 +20,6 @@ from .payment.cost_calculation import (
     MaxCostData,
     calculate_cost,
 )
-from .payment.usage import NormalizedUsage
 from .wallet import credit_balance, deserialize_token_from_string
 
 logger = get_logger(__name__)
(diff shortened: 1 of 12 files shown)


routstr/auth.py
```diff
@@ -675,12 +676,20 @@ async def revert_pay_for_request(
 
 
 async def adjust_payment_for_tokens(
-    key: ApiKey, response_data: dict, session: AsyncSession, deducted_max_cost: int
+    key: ApiKey,
+    response_data: dict,
+    session: AsyncSession,
+    deducted_max_cost: int,
+    usage: NormalizedUsage | None = None,
```
9qeklajc: can't see why `usage` is added here as a parameter, as I can tell, it's always bound to the `response_data` or I'm I missing something

9qeklajc:
did u test the PR against routstrd, it already display CR/CW when available ?

routstr/upstream/base.py
```diff
@@ -101,6 +103,17 @@ def __init__(self, base_url: str, api_key: str, provider_fee: float = 1.01):
         self._models_cache = []
         self._models_by_id = {}
 
+    def normalize_usage(self, usage_data: object) -> NormalizedUsage | None:
```
9qeklajc: I think it make sense to have the impl. here instead of having it in `usage.py` .. because as soon we need to override this function we shall not create a new file but rather add the concrete impl. into the provider specific code .. or what do u think ?

jeroenubbink:
thanks for the feedback, i put it back to draft and will come back to this later.

routstr/auth.py
9qeklajc: cleaned this up and let `usage.py` as is, I think it's cleaner even if (in the future) we need to support different other provider's usages .. it's not worth splitting

routstr/upstream/base.py
9qeklajc: the usage normalization happens now only in `usage.py`

9qeklajc:
@jeroenubbink added support to openrouter and updated some logic

routstr/upstream/base.py
```diff
@@ -501,6 +519,27 @@ def prepare_request_body(
...
+            and self._upstream_accepts_cache_control()
+            and is_explicit_cache_model(
+                model_obj.id,
+                model_obj.forwarded_model_id,
+                model_obj.canonical_slug,
+            )
+        ):
+            if inject_anthropic_cache_breakpoints(data):
+                changed = True
+
```
jeroenubbink, line 542: My only concern here is that routstr now takes control over caching, which is fine except that Anthropic cache writes cost 1.25x input_tokens so it will cost the user extra, even if he will not use the cache. It is maybe not a super big concern. What do you think?

jeroenubbink (commented):
single questions because anthropic cache writing costs extra and we decide for the user now.

routstr/upstream/base.py
sh1ftred: Most anthropic models are being used in agents right now. It absolutely makes sense to pay 1.25x because you are saving so much more on subsequent requests. 

We can give this option to the client (as it already is, i think? don't we just pass through the cache control flag?)

sh1ftred: even if we hard code it for the time being, it is still worth it because majority of anthropic models are being used in agents

jeroenubbink: Ok, my thought is that we should not have an opinion on how a user wants to use an upstream provider, maybe they just do 1 chat request and they still pay 1.25x token for it.

Anyway it is fine for now but something we should then think about in the future.

9qeklajc:
LGTM

 to not hold this for long, deepseek cost issue will be handled in a second issue

PR #556, opened 2026-06-17
fix: deduct mint NUT-02 input fee when crediting trusted-mint topups

jeroenubbink:
## Problem
When a topup token comes from a **trusted** mint (one in `CASHU_MINTS`), `recieve_token` swaps the incoming proofs at that same mint via `wallet.split(..., include_fees=True)` — which pays the mint's **NUT-02 per-proof input fee** — but then returns `token_obj.amount` (the full face value). `credit_balance` credits that full amount.

For any mint that charges an input fee (e.g. Cuba mint, Mountain Lake at 100 ppk; Minibits is 0), routstr therefore credits **more than it actually holds**: it keeps `face − input_fee` in fresh proofs but books `face`. Repeated across topups, the node's wallet drifts toward insolvency.

The foreign-mint path (`swap_to_primary_mint` → `_calculate_swap_amount`) already subtracts `get_fees_for_proofs`; the trusted-mint path just never got the same treatment.

## Fix
Subtract `wallet.get_fees_for_proofs(token_obj.proofs)` from the credited amount in the trusted-mint branch of `recieve_token`. `credit_balance`'s existing `amount <= 0` guard already rejects dust tokens whose fee meets/exceeds their value.

## Tests
- New `test_recieve_token_trusted_mint_deducts_input_fee` (unit): a 1000-sat token with a 3-sat input fee credits 997 — verified fail-first (`1000 != 997`) before the fix.
- Updated `test_recieve_token_valid` to mock the fee call (fee-free mint → full credit, e.g. Minibits).
- Full suite: 293 passed / 1 skipped; ruff + mypy clean.

E2E coverage (real fee-charging mint through `/topup`, proven RED on `main` and GREEN with this fix) is in **Routstr/routstr-testing** (companion PR).

### DIFF
--- a/routstr/wallet.py
+++ b/routstr/wallet.py
@@ -35,6 +35,24 @@ async def get_balance(unit: str) -> int:
     return wallet.available_balance.amount
 
 
+async def _redeem_same_mint(
+    wallet: Wallet, token_obj: Token
+) -> tuple[int, str, str]:  # amount, unit, mint_url
+    """Redeem proofs at their own issuing mint (no cross-mint swap).
+
+    split() re-mints the incoming proofs into fresh ones we own so the sender
+    can't double-spend them. With include_fees=True the mint deducts its NUT-02
+    per-proof input fee, so we end up holding only `amount - input_fees`. Credit
+    that, not the face value, or routstr over-credits the user and its wallet
+    drifts insolvent.
+    """
+    await wallet.load_mint(keyset_id=token_obj.keysets[0])
+    wallet.verify_proofs_dleq(token_obj.proofs)
+    input_fees = wallet.get_fees_for_proofs(token_obj.proofs)
+    await wallet.split(proofs=token_obj.proofs, amount=0, include_fees=True)
+    return int(token_obj.amount) - input_fees, token_obj.unit, token_obj.mint
+
+
 async def recieve_token(
     token: str,
 ) -> tuple[int, str, str]:  # amount, unit, mint_url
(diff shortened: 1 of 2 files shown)


9qeklajc:
This fix makes sense for the trusted-mint path. If `split(..., include_fees=True)` burns the input fee, crediting `amount - input_fees` is the right accounting model.

One adjacent path may still have the same issue: when the token mint is not in `CASHU_MINTS`, `recieve_token` routes to `swap_to_primary_mint`. In the `token mint == primary_mint` shortcut there, it also does a same-mint `split(..., include_fees=True)`, but returns the full `token_amount`.





Small test nit: the new regression test is good, but the comment says “21 proofs @ 100 ppk” while the mocked token has one proof and the fee is mocked as 3. Maybe simplify that to “Mock a 3-sat input fee from the Cashu wallet API.”

jeroenubbink:
> One adjacent path may still have the same issue: when the token mint is not in `CASHU_MINTS`, `recieve_token` routes to `swap_to_primary_mint`. In the `token mint == primary_mint` shortcut there, it also does a same-mint `split(..., include_fees=True)`, but returns the full `token_amount`.

Small refactor and included now in same mint as well.

> Small test nit: the new regression test is good, but the comment says “21 proofs @ 100 ppk” while the mocked token has one proof and the fee is mocked as 3. Maybe simplify that to “Mock a 3-sat input fee from the Cashu wallet API.”

Yes this does not lie about the proofs. I updated it.

9qeklajc:
LGTM

PR #561, opened 2026-06-20
fix(payment): refund truly-empty upstream responses billed a non-zero USD cost

shroominic:
## Summary

When an upstream reports a non-zero USD cost but the response carries **no tokens at all** — `input`, `output`, `cache_read` and `cache_creation` all zero — the USD-cost path in `calculate_cost` billed the user the full USD-derived amount for an empty response. This patch returns an all-zero cost (full refund) for that case only.

**Supersedes #489**, which is 110 commits behind a refactored `main`, conflicts, and ships no tests. This is a clean re-implementation against current `main`.

## The bug (confirmed on current `main`)

In `routstr/payment/cost_calculation.py`, the `usd_cost > 0` branch logs a warning when `input_tokens == 0 and output_tokens == 0` but then unconditionally falls through to `_calculate_from_usd_cost`, which sets `total_msats = cost_in_msats`. A truly empty response is therefore charged the full USD cost.

## The fix, and the tightened gate vs #489

PR #489 refunded whenever prompt + completion tokens were zero. That **over-fires**: a cache-read-only (or cache-creation-only) turn legitimately reports zero prompt/completion tokens with a real cost and must still be billed — #489 would have refunded those legitimate charges for free.

This patch refunds **only when every token bucket is zero**:

```python
truly_empty = (
    input_tokens == 0
    and output_tokens == 0
    and cache_read_tokens == 0
    and cache_creation_tokens == 0
)
```

A cache-only turn (`cache_read_tokens > 0` or `cache_creation_tokens > 0`) is unaffected and still billed.

## Tests

Added to `tests/unit/test_cost_calculation_caching.py`:

- `test_truly_empty_usd_cost_response_is_refunded` — 0 input + 0 output + 0 cache tokens with a non-zero USD cost → `total_msats == 0` (full refund). **Fails without the fix** (bills 200000 msats) — independently confirmed by reverting the source hunk.
- `test_cache_read_only_usd_cost_response_is_billed` — `cache_read_tokens > 0`, 0 prompt/completion, non-zero cost → still billed. Permanent guard against re-introducing #489's over-refund.

Verification: `pytest tests/unit` → **352 passed, 1 skipped**; `ruff check .` → clean; `mypy` (changed file) → clean.

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,8 +38,3 @@ proof_backups
 
 *.todo
 ui_out
-
-# local cashu wallet state (never commit)
-.wallet/
-*.sqlite3-shm
-*.sqlite3-wal
--- a/.gitignore
+++ b/.gitignore
@@ -38,3 +38,8 @@ proof_backups
 
 *.todo
 ui_out
+
+# local cashu wallet state (never commit)
+.wallet/
+*.sqlite3-shm
+*.sqlite3-wal
--- a/routstr/payment/cost_calculation.py
+++ b/routstr/payment/cost_calculation.py
@@ -109,6 +109,39 @@ async def calculate_cost(
     # Try USD cost first
     usd_cost = _resolve_usd_cost(usage_data, response_data)
     if usd_cost > 0:
+        truly_empty = (
+            input_tokens == 0
+            and output_tokens == 0
+            and cache_read_tokens == 0
+            and cache_creation_tokens == 0
+        )
+        if truly_empty:
+            logger.warning(
+                "Upstream reported a USD cost but the response carries no "
+                "tokens at all (input, output, cache-read and cache-creation "
+                "are all zero) — refunding in full rather than billing the "
+                "USD-derived cost for an empty response.",
+                extra={
+                    "model": response_data.get("model", "unknown"),
+                    "usd_cost": usd_cost,
+                    "usage_keys": sorted(usage_data.keys())
+                    if isinstance(usage_data, dict)
+                    else None,
+                },
+            )
+            return CostData(
+                base_msats=0,
+                input_msats=0,
+                output_msats=0,
+                total_msats=0,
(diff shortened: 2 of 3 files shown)


sh1ftred:
Fully vibed PR, check carefully @9qeklajc haha

routstr/payment/cost_calculation.py
```diff
@@ -109,6 +109,39 @@ async def calculate_cost(
...
     usd_cost = _resolve_usd_cost(usage_data, response_data)
     if usd_cost > 0:
+        truly_empty = (
+            input_tokens == 0
+            and output_tokens == 0
+            and cache_read_tokens == 0
+            and cache_creation_tokens == 0
+        )
+        if truly_empty:
+            logger.warning(
```
9qeklajc, line 129: nice.. just one small thing ...
would be better to put this in a function and use it for both paths here and here https://github.com/Routstr/routstr-core/pull/561/changes#diff-d39788ce27d4a012d8bac7f6cde0d170dcfff2336d44954fe396a2c0a77a65a8L72

.gitignore
```diff
@@ -38,3 +38,8 @@ proof_backups
 
 *.todo
 ui_out
+
```
9qeklajc: I'd rather have this done in a separate PR

9qeklajc:
Addressed both comments in `ee5ced1`:

- **`_empty_cost` helper** — extracted the all-zero refund cost object into `_empty_cost(cls=CostData)` and reused it in both paths: the no-usage-data branch (`_empty_cost(MaxCostData)`) and the truly-empty USD-cost branch (`_empty_cost()`). No more duplicated zero-object construction.
- **`.gitignore`** — reverted; will land in a separate PR.

`ruff` + `mypy` clean, `pytest tests/unit` → 353 passed.

PR #566, opened 2026-06-21
fix: always charge balance on reservation overrun even after stale sweep

9qeklajc:
When actual cost exceeds the discounted reservation, the finalize UPDATE previously guarded on `reserved_balance >= deducted_max_cost`. If the stale-reservation sweeper had already released the reservation, that guard failed and the charge was silently dropped.

Replaces the WHERE guard with a clamping `CASE` so the charge always lands, and `reserved_balance` never goes negative — for both the billing key and the child key.

Test: `tests/integration/test_free_response_stale_reservation.py`.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -1044,7 +1044,7 @@ async def adjust_payment_for_tokens(
                         total_spent=col(ApiKey.total_spent) + chargeable,
                     )
                 )
-                result = await session.exec(finalize_stmt)  # type: ignore[call-overload]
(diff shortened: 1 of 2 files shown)


routstr/auth.py
jeroenubbink: rowcount is now always 1 so we can drop this statement

jeroenubbink: Because `result.rowcount` is now always 1 this is dead code.

jeroenubbink (changes requested):
We should remove the no longer needed conditional

9qeklajc:
good catch

PR #567, opened 2026-06-21
feat: prune dead zero-balance API keys with background janitor

9qeklajc:
Adds a background janitor that deletes provably-dead API keys — the 0/0/0 rows left behind when a funded key is fully refunded or expires without ever being used.

A key is pruned only if it has zero balance, reservation, spend and requests, no parent, no children, no pending invoice, and is older than a grace period. Cashu transaction rows are unlinked (not deleted) first, so the financial audit trail survives. The eligibility predicate runs inside both the unlink and delete statements, so a key that gains balance mid-run is left untouched.

New settings (env): `DEAD_KEY_PRUNE_INTERVAL_SECONDS` (default 3600, set 0 to disable) and `DEAD_KEY_MIN_AGE_SECONDS` (grace period, default 1 week).

Test: `tests/integration/test_prune_dead_api_keys.py`.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -1305,6 +1305,35 @@ async def periodic_key_reset() -> None:
             logger.error(f"Error in periodic_key_reset: {e}")
 
 
+async def periodic_dead_key_prune() -> None:
+    """Periodically prune dead API keys. Interval <= 0 disables it.
+
+    See ``prune_dead_api_keys`` for eligibility.
+    """
+    from .core.db import create_session, prune_dead_api_keys
+
+    interval = settings.dead_key_prune_interval_seconds
+    if interval <= 0:
+        logger.info("Dead-key pruning disabled (interval <= 0)")
+        return
+
+    while True:
+        try:
+            await asyncio.sleep(interval)
+        except asyncio.CancelledError:
+            break
+
+        try:
+            async with create_session() as session:
+                await prune_dead_api_keys(
+                    session, settings.dead_key_min_age_seconds
+                )
+        except asyncio.CancelledError:
+            break
+        except Exception as e:
+            logger.error(f"Error in periodic_dead_key_prune: {e}")
+
+
 STALE_RESERVATION_SWEEP_INTERVAL_SECONDS: int = 60
 
 
--- a/routstr/core/db.py
+++ b/routstr/core/db.py
@@ -9,9 +9,10 @@ from typing import AsyncGenerator
 from alembic import command
 from alembic.config import Config
 from alembic.util.exc import CommandError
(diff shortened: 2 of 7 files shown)


routstr/core/db.py
```diff
@@ -125,6 +126,63 @@ async def release_stale_reservations(
...
+    # statements so a key that gained balance mid-run is left untouched.
+    await session.exec(  # type: ignore[call-overload]
+        update(CashuTransaction)
+        .where(col(CashuTransaction.api_key_hashed_key).in_(eligible_hashes))
+        .values(api_key_hashed_key=None)
+    )
+
+    result = await session.exec(  # type: ignore[call-overload]
+        delete(ApiKey).where(col(ApiKey.hashed_key).in_(eligible_hashes))
+    )
```
jeroenubbink, line 175: This DELETE introduces a new invariant break. Until now keys were never deleted, so any code that authenticated a key could assume it still exists for the rest of the request. The janitor breaks that.
  
The Lightning topup path is fine,b ut the cashu POST /topup path isn't: `wallet.credit_balance` redeems the token at the mint, then does `UPDATE api_keys SET balance = balance + amt WHERE hashed_key = ...` without checking `rowcount`. If this prune deletes a dormant 0/0/0 key (older than the grace period) in that window, the token is already spent at the mint but the credit matches 0 rows so the user's value is captured uncredited.

It's a really small chance this can happen but it'd be good to pair it with a `rowcount` guard in `credit_balance`:

```python
result = await session.exec(stmt)  # the balance UPDATE
if (result.rowcount or 0) == 0:
    raise ValueError("API key no longer exists; token redeemed but not credited")
```

routstr/auth.py
```diff
@@ -1305,6 +1305,35 @@ async def periodic_key_reset() -> None:
...
+    """Periodically prune dead API keys. Interval <= 0 disables it.
+
+    See ``prune_dead_api_keys`` for eligibility.
+    """
+    from .core.db import create_session, prune_dead_api_keys
+
+    interval = settings.dead_key_prune_interval_seconds
+    if interval <= 0:
+        logger.info("Dead-key pruning disabled (interval <= 0)")
+        return
```
jeroenubbink, line 1318: I don't think this is under test coverage, but it's minor.

jeroenubbink (approved):
LGTM.

Super small chance somebody decides to topup with cashu and janitor runs while it's happening. Chances of this happening are really low so up to you if you feel like adding a guard for it.

Also found a small untested code branch while running the tests, but seems very unimportant.

9qeklajc:
thnks for the review

PR #569, opened 2026-06-25
docs(ui): document dev/build/serve workflow; make missing-bundle warning actionable

jeroenubbink:
## What

- Replace `ui/README.md` (still stock create-next-app boilerplate) with the actual workflow for this app:
  - It's a **Next.js static export** served by FastAPI from `ui_out/` — same origin as the API in production, no separate web server.
  - The **dev loop**: backend on `:8000` + `make ui-dev` (Next dev server, hot reload) on `:3000`, which auto-targets `http://127.0.0.1:8000` when `NEXT_PUBLIC_API_URL` is unset. No `make ui-build` loop while developing.
  - The build/integration commands (`make ui-build` / `make ui-build-docker`) and the cross-origin CORS note.
- Make the startup "`ui_out` not found" log **actionable**: it now points to `make ui-build` / `make ui-dev` instead of just saying it skipped serving.

## Why

The README described none of how the UI is actually built or served, which makes the dev setup hard to discover (running the backend alone serves API-only and a UI route just falls back to JSON). No behavior change beyond the log message wording.

### DIFF
--- a/routstr/core/main.py
+++ b/routstr/core/main.py
@@ -369,7 +369,10 @@ if UI_DIST_PATH.exists() and UI_DIST_PATH.is_dir():
 
 else:
     logger.warning(
-        f"UI dist directory not found at {UI_DIST_PATH}, skipping static file serving"
+        "UI dist directory not found at %s; serving API only. Run `make ui-build` "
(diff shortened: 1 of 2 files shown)


9qeklajc:
LGTM

Issue #553, opened 2026-06-13
Config ownership: env vs DB precedence, secret storage, and first-run admin bootstrap


jeroenubbink:
## Problem

`SettingsService.initialize()` snapshots **all** settings into the `settings` DB row on first boot, then lets the DB win on every later boot. Two consequences:

1. Changing `.env` after first boot (`ADMIN_PASSWORD`, `UPSTREAM_*`, …) is **silently ignored** — the first-boot DB snapshot shadows it. A stale row with `admin_password: "admin"` can even override a strong env password.
2. Secrets (`admin_password`, `nsec`, `upstream_api_key`) sit in **plaintext** in the DB.

There's also no defined story for how the *first* admin gets into the UI on a fresh node.

## Principle

Two roles, not necessarily the same person:

- the **installer** — installs the node (once, deploy-level infrastructure);
- the **operator** — runs it day-to-day (upstream providers, pricing, mints, identity).

The operator must never need shell access + a fastapi restart to keep the node running. So: `.env` holds only what the installer sets and that genuinely can't change on a running node; everything an operator manages lives in the DB and is editable in the admin UI. No setting is read from two live sources, so nothing can silently overwrite anything. **There is exactly one way to configure each thing** — no "if X is set, otherwise fall back to Y" modes.

## Proposed: per-field ownership

**`.env`-only (install-time, immutable without redeploy)**
`DATABASE_URL`, listen host/port, `models_path`, `reset_reserved_balance_on_startup` (a horizontal-scaling flag), `tor_proxy_url`, `ROUTSTR_SECRET_KEY` (see Secret storage).

**DB + admin UI (operator-managed)**
Upstream providers (the existing providers table), pricing (`fixed_*`, `exchange_fee`, `upstream_provider_fee`, `tolerance_percentage`, `child_key_cost`, `min_request_msat`), payout (`min_payout_sat`, `payout_interval_seconds`), cashu (`cashu_mints`, `primary_mint`, `receive_ln_address`), `name`/`description`, `relays`, the `*_refresh_interval*` / `enable_*_refresh` flags, `enable_analytics_sharing`.

Env may **seed** these on first boot (so existing compose/`.env` setups keep working); thereafter the DB is authoritative and any divergence between env and the persisted value is **logged, never silently applied**.

**Secrets** — `admin_password` and `nsec` — live only in the DB, in dedicated encrypted columns (see below), never in the `settings` blob and never in plaintext.

## Secret storage

[...]

PR #570, opened 2026-06-26
feat(config): store node secrets in an encrypted vault

jeroenubbink:
## What

First step toward issue #553 (config ownership): move node **secrets** out of
the settings JSON blob and into a dedicated, encrypted store, so a secret has
exactly one home and can't be silently shadowed by a stale env value or blob row.

- New `vault` module: Fernet encryption for secrets at rest and scrypt hashing
  for the admin password, keyed by a new `ROUTSTR_SECRET_KEY`.
- New singleton `Secret` table (hand-written migration) holding the admin
  password **hash**, the **encrypted** nsec, and an `nsec_managed` flag marking
  that the vault has taken ownership of the nsec — credentials, not config, so a
  typed table rather than the settings blob.
- Boot-time bootstrap: load/persist secrets on startup, migrate any legacy
  plaintext (from env or the old blob) into the encrypted store, and generate a
  temporary admin password on first run. Once the vault owns the nsec, legacy
  env/blob copies are never re-applied: a stale `NSEC` can't override a rotated
  key, and an nsec cleared via the admin API stays cleared across reboots.
- Admin API: a dedicated endpoint to set/rotate/clear the nsec, password
  verification against the stored scrypt hash (replacing a non-constant-time
  compare), and redaction of secret values in settings responses.
- The settings blob no longer persists secret fields; they live decrypted in
  memory at runtime only.
- `scripts/reset_admin_password.py`: offline recovery for a locked-out operator.

## ⚠️ Breaking changes for operators

1. **`ROUTSTR_SECRET_KEY` is now required.** A node that needs to read or write a
   stored secret will refuse to proceed without it and prints the command to
   generate one:
   `python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"`
   Keep this key stable — if it changes, the encrypted nsec can no longer be
   decrypted (see "Known trade-offs" #4). Added to `.env.example` and the README.

2. **The admin password is generated, not configured.** On first start the node
   creates a temporary admin password and logs it **once**, with the `/admin` URL;
   log in with it and change it from there (there's no setup screen). `ADMIN_PASSWORD`
   has been dropped from `.env.example` and isn't something to configure going forward.
   Any value already present in the environment is still honoured once as a seed, so
   existing nodes keep their password — after which it's ignored. An operator who
   loses access can recover at any time with `scripts/reset_admin_password.py`
   (`--password` to set a known one, or `--regenerate` to get a fresh password logged
   on the next start).

## Known trade-offs (intentional, called out for review)

These were found during self-review and deliberately left for follow-up rather
than fixed here:

1. **Generated admin password is logged in plaintext.** Standard headless-
   bootstrap pattern (k8s/postgres images); it's the only way to surface a
   bootstrap secret with no UI. Shown once; operator is told to rotate.
2. **scrypt now runs on the unauthenticated `/api/login`** with no rate limiting,
   turning login into a CPU/memory amplification primitive under flood. Rate
   limiting is pre-existing scope; flagging it.
3. **`ADMIN_PASSWORD` env is ignored after first boot** (by design per #553).
4. **No `ROUTSTR_SECRET_KEY` rotation path.** Changing the key bricks the
   encrypted nsec — the node detects this and fails fast rather than booting with
   an unreadable identity. The issue mentions decrypt-old/re-encrypt-new rotation;
   this PR only detects and stops. Recovery today is re-entering the nsec.
5. **Minor nits**: `admin_login` reads the ORM object after its session block
   closes (safe — `expire_on_commit=False` — but inconsistent with the sibling
   handlers); the reset CLI's two flags are mutually exclusive only at the argparse
   layer; `--regenerate` has no confirm prompt.

## Two deviations from the issue worth noting

- **Fail-fast is lazy, not eager.** The node only requires `ROUTSTR_SECRET_KEY`
  when a secret actually has to be en/decrypted (password hashing is key-
  independent). So a node with no nsec and no key still boots. The issue's wording
  is "refuses to start without it"; in practice it refuses the moment a secret
  operation needs the key. Still safe — no plaintext is ever written — but it
  fails on first secret use, not unconditionally at start.
- **Key rotation is detect-and-stop, not re-encrypt** (same as trade-off #4).

## API compatibility

Removing the admin password from the settings model also removes it from the
`GET /admin/api/settings` response. This is **not** a backward-incompatible API
change. The endpoint declares no `response_model` — it returns a free-form
`dict` — so `admin_password` was never part of its OpenAPI schema. The OpenAPI
schema is the contract, and the field was not in the contract, so dropping it
doesn't break it. In practice the field only ever serialized as the constant
`"[REDACTED]"` (the real password was never exposed), so no consumer could have
read a usable value from it. The PATCH side is likewise unaffected: a stray
`admin_password` in a settings update is silently ignored, exactly as before.

## Scope

This is the first of two stacked changes for #553's secret handling. A follow-up
will encrypt the per-provider upstream API keys (`upstream_providers.api_key`),
which need a blind index and so are kept separate. Field-ownership cleanup
(env-vs-DB precedence) and the admin-UI rework are tracked as later work in #553.

## Testing

- `make lint` (ruff + mypy) clean; `make test` green.
- New unit tests for the vault primitives and the blob-sunsetting logic; new
  integration tests for the Secret model, boot-time bootstrap (including startup
  ordering, legacy-plaintext migration, stale-env-doesn't-override-vault, and
  cleared-nsec-stays-cleared), admin auth, the nsec endpoint, the
  settings-redaction endpoint, and the recovery CLI.
- End-to-end coverage of the whole-system behaviours (real boot, real env
  secrets, on-disk SQLite, fail-fast, first-run password, key-change brick,
  nsec-survives-reboot) is added in the routstr-testing harness in a companion PR.

Closes part of #553.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,11 +55,18 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
 
 1. **Prepare your `.env`**:
    ```bash
-   ADMIN_PASSWORD=mysecretpassword
+   # Required: encrypts secrets at rest. The node won't start without it.
+   ROUTSTR_SECRET_KEY=<generated-key>
    NAME="My AI Node"
    DESCRIPTION="Fast access to models"
    NSEC=yournsec
    RECEIVE_LN_ADDRESS=yourname@wallet.com
+   ```
+
+   Generate `ROUTSTR_SECRET_KEY` once and keep it stable — changing it makes
+   previously encrypted secrets unreadable:
+   ```bash
+   python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
    ```
 
 2. **Start the services**:
@@ -67,7 +74,15 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
    docker compose up -d
    ```
 
-3. **Configure**:
+3. **Get your admin password**:
+   On first start the node generates an admin password and logs it once with the
+   `/admin` URL. Read it from the logs:
+   ```bash
+   docker compose logs routstr | grep -i admin
+   ```
+   (Lost it? Reset with `python scripts/reset_admin_password.py --regenerate`.)
+
+4. **Configure**:
    Open [http://localhost:8000/admin/](http://localhost:8000/admin/) to connect your AI providers and set pricing.
 
 For full instructions, see the **[Provider Quick Start Guide](https://docs.routstr.com/provider/quickstart/)**.
--- a/README.md
+++ b/README.md
@@ -55,7 +55,10 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
 
 1. **Prepare your `.env`**:
    ```bash
-   # Required: encrypts secrets at rest. The node won't start without it.
+   # Optional: encrypts node secrets at rest. If unset, the node generates a key
+   # on first start, writes it to routstr_secret.key, and prints it once — back
+   # up that file. Set it explicitly to manage the key yourself (recommended in
+   # production).
    ROUTSTR_SECRET_KEY=<generated-key>
    NAME="My AI Node"
    DESCRIPTION="Fast access to models"
@@ -63,10 +66,11 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
    RECEIVE_LN_ADDRESS=yourname@wallet.com
    ```
 
-   Generate `ROUTSTR_SECRET_KEY` once and keep it stable — changing it makes
-   previously encrypted secrets unreadable:
+   If you don't set one, a key is generated and printed on first start — save it
+   somewhere safe (losing it makes previously encrypted secrets unreadable). To
+   supply your own, generate it once and keep it stable:
    ```bash
-   python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
+   uv run python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
    ```
 
 2. **Start the services**:
@@ -80,7 +84,7 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
    ```bash
    docker compose logs routstr | grep -i admin
    ```
-   (Lost it? Reset with `python scripts/reset_admin_password.py --regenerate`.)
+   (Lost it? Reset with `docker compose exec routstr /.venv/bin/python scripts/reset_admin_password.py --regenerate`.)
 
 4. **Configure**:
    Open [http://localhost:8000/admin/](http://localhost:8000/admin/) to connect your AI providers and set pricing.
--- a/README.md
+++ b/README.md
@@ -62,9 +62,12 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
    ROUTSTR_SECRET_KEY=<generated-key>
    NAME="My AI Node"
    DESCRIPTION="Fast access to models"
-   NSEC=yournsec
    RECEIVE_LN_ADDRESS=yourname@wallet.com
    ```
+
+   Your Nostr identity (`nsec`) is not set in `.env` — configure it from the admin
+   UI after first start, where it's stored encrypted in the database. (`NSEC` in
+   `.env` is still read once as a legacy seed for existing deployments.)
 
    If you don't set one, a key is generated and printed on first start — save it
    somewhere safe (losing it makes previously encrypted secrets unreadable). To
--- a/docs/provider/deployment.md
+++ b/docs/provider/deployment.md
@@ -87,6 +87,8 @@ services:
       - ./logs:/app/logs
     environment:
       - TOR_PROXY_URL=socks5://tor:9050
+      # Keep the database (and the key file generated beside it) on the volume.
+      - DATABASE_URL=sqlite:////app/data/routstr.db
     depends_on:
       - tor
 
@@ -134,6 +136,9 @@ services:
       
       # Lightning withdrawals
       - RECEIVE_LN_ADDRESS=me@walletofsatoshi.com
+
+      # Keep the database (and the key file generated beside it) on the volume.
+      - DATABASE_URL=sqlite:////app/data/routstr.db
     volumes:
       - ./data:/app/data
 ```
@@ -156,9 +161,11 @@ Example `.env`:
 UPSTREAM_BASE_URL=https://api.openai.com/v1
 UPSTREAM_API_KEY=sk-proj-...
 ADMIN_PASSWORD=change-me
+# Keep the database (and the key file generated beside it) on the mounted volume.
+DATABASE_URL=sqlite:////app/data/routstr.db
 # Encrypts node secrets at rest. Optional — if unset, a key is generated next to
-# your database (on the same volume) and printed once. Set it explicitly to
-# manage the key yourself.
+# your database (on the same volume) and its file is named once for backup. Set
+# it explicitly to manage the key yourself.
 ROUTSTR_SECRET_KEY=
 NAME=My Provider Node
 RECEIVE_LN_ADDRESS=me@walletofsatoshi.com
@@ -177,11 +184,13 @@ See [Configuration](configuration.md) for all available options.
 
 ## Persistence
 
-Routstr stores all data in `/app/data`:
+Point `DATABASE_URL` inside `/app/data` (as the examples above do) so everything
+Routstr persists lands on the mounted volume:
 
 | Path | Contents |
 |------|----------|
-| `keys.db` | SQLite database (settings, API keys, sessions) |
+| `routstr.db` | SQLite database (settings, API keys, sessions) |
+| `routstr_secret.key` | Auto-generated master key, written beside the database when `ROUTSTR_SECRET_KEY` is unset |
 | `.wallet/` | Cashu wallet data (your Bitcoin!) |
 
 !!! warning "Back Up Your Data"
--- a/docs/provider/deployment.md
+++ b/docs/provider/deployment.md
@@ -38,7 +38,6 @@ services:
       - routstr-data:/app/data
     environment:
       DATABASE_URL: "sqlite:////app/data/routstr.db"
-      ADMIN_KEY: "your-secure-admin-key"
       LOG_LEVEL: "info"
 
 volumes:
@@ -127,8 +126,8 @@ services:
       - UPSTREAM_BASE_URL=https://api.openai.com/v1
       - UPSTREAM_API_KEY=sk-proj-...
       
-      # Secure the dashboard (recommended)
-      - ADMIN_PASSWORD=your-secure-password
+      # The admin password is generated and logged once on first start; set
+      # ADMIN_PASSWORD here only as a legacy seed for an existing deployment.
       
       # Node identity
       - NAME=My Provider Node
@@ -160,7 +159,6 @@ Example `.env`:
 ```bash
 UPSTREAM_BASE_URL=https://api.openai.com/v1
 UPSTREAM_API_KEY=sk-proj-...
-ADMIN_PASSWORD=change-me
 # Keep the database (and the key file generated beside it) on the mounted volume.
 DATABASE_URL=sqlite:////app/data/routstr.db
 # Encrypts node secrets at rest. Optional — if unset, a key is generated next to
--- a/docs/provider/deployment.md
+++ b/docs/provider/deployment.md
@@ -156,9 +156,20 @@ Example `.env`:
 UPSTREAM_BASE_URL=https://api.openai.com/v1
 UPSTREAM_API_KEY=sk-proj-...
 ADMIN_PASSWORD=change-me
+# Encrypts node secrets at rest. Optional — if unset, a key is generated next to
+# your database (on the same volume) and printed once. Set it explicitly to
+# manage the key yourself.
+ROUTSTR_SECRET_KEY=
 NAME=My Provider Node
 RECEIVE_LN_ADDRESS=me@walletofsatoshi.com
 ```
+
+!!! note "Secret key persistence"
+    If you leave `ROUTSTR_SECRET_KEY` unset, the node generates one and stores it
+    as `routstr_secret.key` **next to your database**, so it persists on the same
+    volume as your data — just include that volume in your backups. For stronger
+    isolation (keeping the key off the data volume), set `ROUTSTR_SECRET_KEY` from
+    a secrets manager instead.
 
 See [Configuration](configuration.md) for all available options.
 
--- a/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
+++ b/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
@@ -1,8 +1,8 @@
 """add secrets table
 
 Revision ID: c6f8d2e4a1b3
-Revises: b5e7c9d1f3a2
-Create Date: 2026-06-24 00:00:00.000000
+Revises: c6d7e8f9a0b1
+Create Date: 2026-07-07 00:00:00.000000
 
 Creates the node-level singleton secret store (issue #553). Schema only; moving
 any legacy plaintext into the encrypted/hashed columns happens at bootstrap,
@@ -14,7 +14,7 @@ import sqlmodel
 from alembic import op
 
 revision = "c6f8d2e4a1b3"
-down_revision = "b5e7c9d1f3a2"
+down_revision = "c6d7e8f9a0b1"
 branch_labels = None
 depends_on = None
 
--- a/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
+++ b/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
@@ -1,7 +1,7 @@
 """add secrets table
 
 Revision ID: c6f8d2e4a1b3
-Revises: c6d7e8f9a0b1
+Revises: d7e8f9a0b1c2
 Create Date: 2026-07-07 00:00:00.000000
 
 Creates the node-level singleton secret store (issue #553). Schema only; moving
@@ -17,7 +17,7 @@ import sqlmodel
 from alembic import op
 
 revision = "c6f8d2e4a1b3"
-down_revision = "c6d7e8f9a0b1"
+down_revision = "d7e8f9a0b1c2"
 branch_labels = None
 depends_on = None
 
--- a/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
+++ b/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
@@ -1,0 +1,42 @@
+"""add secrets table
+
+Revision ID: c6f8d2e4a1b3
+Revises: b5e7c9d1f3a2
+Create Date: 2026-06-24 00:00:00.000000
+
+Creates the node-level singleton secret store (issue #553). Schema only; moving
+any legacy plaintext into the encrypted/hashed columns happens at bootstrap,
+where the live ROUTSTR_SECRET_KEY is available.
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+revision = "c6f8d2e4a1b3"
+down_revision = "b5e7c9d1f3a2"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "secrets",
+        sa.Column("id", sa.Integer(), nullable=False),
+        sa.Column(
+            "admin_password_hash",
+            sqlmodel.sql.sqltypes.AutoString(),
+            nullable=True,
+        ),
+        sa.Column(
+            "encrypted_nsec",
+            sqlmodel.sql.sqltypes.AutoString(),
+            nullable=True,
+        ),
+        sa.Column("updated_at", sa.Integer(), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+    )
+
+
+def downgrade() -> None:
+    op.drop_table("secrets")
--- a/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
+++ b/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
@@ -6,7 +6,9 @@ Create Date: 2026-07-07 00:00:00.000000
 
 Creates the node-level singleton secret store (issue #553). Schema only; moving
 any legacy plaintext into the encrypted/hashed columns happens at bootstrap,
-where the live ROUTSTR_SECRET_KEY is available.
+where the live ROUTSTR_SECRET_KEY is available. ``nsec_managed`` records that the
+vault has taken ownership of the nsec, so a cleared identity is never resurrected
+from a stale legacy ``NSEC`` env var / settings blob on the next boot.
 """
 
 import sqlalchemy as sa
@@ -32,6 +34,12 @@ def upgrade() -> None:
             "encrypted_nsec",
             sqlmodel.sql.sqltypes.AutoString(),
             nullable=True,
+        ),
+        sa.Column(
+            "nsec_managed",
+            sa.Boolean(),
+            nullable=False,
+            server_default=sa.false(),
         ),
         sa.Column("updated_at", sa.Integer(), nullable=True),
         sa.PrimaryKeyConstraint("id"),
--- a/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
+++ b/migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
@@ -6,9 +6,10 @@ Create Date: 2026-07-07 00:00:00.000000
 
 Creates the node-level singleton secret store (issue #553). Schema only; moving
 any legacy plaintext into the encrypted/hashed columns happens at bootstrap,
-where the live ROUTSTR_SECRET_KEY is available. ``nsec_managed`` records that the
-vault has taken ownership of the nsec, so a cleared identity is never resurrected
-from a stale legacy ``NSEC`` env var / settings blob on the next boot.
+where the live ROUTSTR_SECRET_KEY is available. ``nsec_state`` records the vault's
+ownership of the nsec (legacy | encrypted | cleared), so a cleared identity is
+never resurrected from a stale legacy ``NSEC`` env var / settings blob on the next
+boot.
 """
 
 import sqlalchemy as sa
@@ -36,10 +37,10 @@ def upgrade() -> None:
             nullable=True,
         ),
         sa.Column(
-            "nsec_managed",
-            sa.Boolean(),
+            "nsec_state",
+            sqlmodel.sql.sqltypes.AutoString(),
             nullable=False,
-            server_default=sa.false(),
+            server_default="legacy",
         ),
         sa.Column("updated_at", sa.Integer(), nullable=True),
         sa.PrimaryKeyConstraint("id"),
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -310,29 +310,6 @@ async def update_nsec(request: Request, payload: NsecUpdate) -> dict[str, object
     return {"ok": True, "npub": npub}
 
 
-class SetupRequest(BaseModel):
-    password: str
-
-
-@admin_router.post("/api/setup")
-async def initial_setup(request: Request, payload: SetupRequest) -> dict[str, object]:
-    async with create_session() as session:
-        secret = await get_secret(session)
-        if secret.admin_password_hash:
-            raise HTTPException(status_code=409, detail="Admin password already set")
-        pw = (payload.password or "").strip()
-        if len(pw) < vault.MIN_PASSWORD_LENGTH:
-            raise HTTPException(
-                status_code=400,
-                detail=(
-                    "Password must be at least "
-                    f"{vault.MIN_PASSWORD_LENGTH} characters"
-                ),
-            )
-        await set_admin_password(session, pw)
-    return {"ok": True}
-
-
 class AdminLoginRequest(BaseModel):
     password: str
 
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -343,11 +343,14 @@ async def admin_login(
 ) -> dict[str, object]:
     async with create_session() as session:
         secret = await get_secret(session)
+        # Read the hash while the session is open; the ORM object is detached
+        # once the context exits and its attributes can no longer be loaded.
+        password_hash = secret.admin_password_hash
 
-    if not secret.admin_password_hash:
+    if not password_hash:
         raise HTTPException(status_code=500, detail="Admin password not configured")
 
-    if not vault.verify_password(payload.password, secret.admin_password_hash):
+    if not vault.verify_password(payload.password, password_hash):
         raise HTTPException(status_code=401, detail="Invalid password")
 
     token = secrets.token_urlsafe(32)
@@ -526,7 +529,6 @@ class ModelCreate(BaseModel):
 async def upsert_provider_model(
     provider_id: str, payload: ModelCreate
 ) -> dict[str, object]:
-    print(payload)
     logger.info(
         f"UPSERT_PROVIDER_MODEL called: provider_id={provider_id}, model_id={payload.id}"
     )
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -20,6 +20,7 @@ from ..wallet import (
     send_token,
     slow_filter_spend_proofs,
 )
+from . import vault
 from .db import (
     ApiKey,
     CashuTransaction,
@@ -28,6 +29,9 @@ from .db import (
     ModelRow,
     UpstreamProviderRow,
     create_session,
+    get_secret,
+    set_admin_password,
+    set_nsec,
 )
 from .db import (
     store_cashu_transaction_with_retry as store_cashu_transaction,
@@ -35,7 +39,7 @@ from .db import (
 from .log_manager import log_manager
 from .logging import get_logger
 from .provider_slugs import allocate_unique_provider_slug
-from .settings import SettingsService, settings
+from .settings import SettingsService, derive_npub_from_nsec, settings
 
 logger = get_logger(__name__)
 
@@ -206,8 +210,6 @@ async def get_settings(request: Request) -> dict:
     data = settings.dict()
     if "upstream_api_key" in data:
         data["upstream_api_key"] = "[REDACTED]" if data["upstream_api_key"] else ""
-    if "admin_password" in data:
-        data["admin_password"] = "[REDACTED]" if data["admin_password"] else ""
     if "nsec" in data:
         data["nsec"] = "[REDACTED]" if data["nsec"] else ""
     return data
@@ -224,9 +226,10 @@ class PasswordUpdate(BaseModel):
 
 @admin_router.patch("/api/settings", dependencies=[Depends(require_admin_api)])
 async def update_settings(request: Request, update: SettingsUpdate) -> dict:
-    # Remove sensitive fields from general settings update
+    # Secrets are not editable through the general settings endpoint; they have
+    # dedicated rotation paths and never reach the settings blob.
     settings_data = update.root.copy()
-    sensitive_fields = ["admin_password", "upstream_api_key", "nsec"]
+    sensitive_fields = ["upstream_api_key", "nsec"]
     for field in sensitive_fields:
         if field in settings_data:
             del settings_data[field]
@@ -241,8 +244,6 @@ async def update_settings(request: Request, update: SettingsUpdate) -> dict:
     data = new_settings.dict()
     if "upstream_api_key" in data:
         data["upstream_api_key"] = "[REDACTED]" if data["upstream_api_key"] else ""
-    if "admin_password" in data:
-        data["admin_password"] = "[REDACTED]" if data["admin_password"] else ""
     if "nsec" in data:
         data["nsec"] = "[REDACTED]" if data["nsec"] else ""
     return data
@@ -250,26 +251,63 @@ async def update_settings(request: Request, update: SettingsUpdate) -> dict:
 
 @admin_router.patch("/api/password", dependencies=[Depends(require_admin_api)])
 async def update_password(request: Request, password_update: PasswordUpdate) -> dict:
-    current_password = settings.admin_password
-
-    if not current_password:
-        raise HTTPException(status_code=500, detail="Admin password not configured")
-
-    if password_update.current_password != current_password:
-        raise HTTPException(status_code=401, detail="Current password is incorrect")
-
-    # Validate new password
-    new_password = password_update.new_password.strip()
-    if len(new_password) < 6:
-        raise HTTPException(
-            status_code=400, detail="New password must be at least 6 characters"
-        )
-
-    # Update password
     async with create_session() as session:
-        await SettingsService.update({"admin_password": new_password}, session)
+        secret = await get_secret(session)
+
+        if not secret.admin_password_hash:
+            raise HTTPException(
+                status_code=500, detail="Admin password not configured"
+            )
+
+        if not vault.verify_password(
+            password_update.current_password, secret.admin_password_hash
+        ):
+            raise HTTPException(
+                status_code=401, detail="Current password is incorrect"
+            )
+
+        # Validate new password
+        new_password = password_update.new_password.strip()
+        if len(new_password) < vault.MIN_PASSWORD_LENGTH:
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "New password must be at least "
+                    f"{vault.MIN_PASSWORD_LENGTH} characters"
+                ),
+            )
+
+        await set_admin_password(session, new_password)
 
     return {"ok": True, "message": "Password updated successfully"}
+
+
+class NsecUpdate(BaseModel):
+    nsec: str
+
+
+@admin_router.patch("/api/nsec", dependencies=[Depends(require_admin_api)])
+async def update_nsec(request: Request, payload: NsecUpdate) -> dict[str, object]:
+    # The node's Nostr identity is a secret: it is stored encrypted in the
+    # Secret store, never in the settings blob, so it gets its own endpoint
+    # rather than riding the general settings PATCH (which strips it). An empty
+    # nsec clears the identity.
+    nsec = payload.nsec.strip()
+    npub = ""
+    if nsec:
+        derived = derive_npub_from_nsec(nsec)
+        if not derived:
+            raise HTTPException(status_code=400, detail="Invalid nsec")
+        npub = derived
+
+    async with create_session() as session:
+        await set_nsec(session, nsec)
+
+    # Reflect the change in the live runtime so Nostr signing/announcements pick
+    # it up without a restart (mirrors what bootstrap_secrets sets at boot).
+    settings.nsec = nsec
+    settings.npub = npub
+    return {"ok": True, "npub": npub}
 
 
 class SetupRequest(BaseModel):
@@ -278,15 +316,20 @@ class SetupRequest(BaseModel):
 
 @admin_router.post("/api/setup")
 async def initial_setup(request: Request, payload: SetupRequest) -> dict[str, object]:
-    if settings.admin_password:
-        raise HTTPException(status_code=409, detail="Admin password already set")
-    pw = (payload.password or "").strip()
-    if len(pw) < 8:
-        raise HTTPException(
-            status_code=400, detail="Password must be at least 8 characters"
-        )
     async with create_session() as session:
-        await SettingsService.update({"admin_password": pw}, session)
+        secret = await get_secret(session)
+        if secret.admin_password_hash:
+            raise HTTPException(status_code=409, detail="Admin password already set")
+        pw = (payload.password or "").strip()
+        if len(pw) < vault.MIN_PASSWORD_LENGTH:
+            raise HTTPException(
+                status_code=400,
+                detail=(
+                    "Password must be at least "
+                    f"{vault.MIN_PASSWORD_LENGTH} characters"
+                ),
+            )
+        await set_admin_password(session, pw)
     return {"ok": True}
 
 
@@ -298,12 +341,13 @@ class AdminLoginRequest(BaseModel):
 async def admin_login(
     request: Request, payload: AdminLoginRequest
 ) -> dict[str, object]:
-    admin_pw = settings.admin_password
+    async with create_session() as session:
+        secret = await get_secret(session)
 
-    if not admin_pw:
+    if not secret.admin_password_hash:
         raise HTTPException(status_code=500, detail="Admin password not configured")
 
-    if payload.password != admin_pw:
+    if not vault.verify_password(payload.password, secret.admin_password_hash):
         raise HTTPException(status_code=401, detail="Invalid password")
 
     token = secrets.token_urlsafe(32)
--- a/routstr/core/db.py
+++ b/routstr/core/db.py
@@ -6,6 +6,7 @@ import sqlite3
 import time
 import uuid
 from contextlib import asynccontextmanager
+from enum import Enum
 from typing import AsyncGenerator
 
 from alembic import command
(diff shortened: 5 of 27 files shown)


9qeklajc:
Spent a good while going through this one. Short version: I really like where it's going, but I wouldn't merge it as-is yet — there are a couple of correctness bugs that can quietly lose secrets, plus some smaller cleanup.

Getting secrets out of the settings JSON blob is the right call, and the `vault` module is a clean place to put the crypto. Nice touch using scrypt + `hmac.compare_digest` instead of the old plaintext compare, and the `fernet:v1:` prefix so you can tell encrypted values from legacy ones. The test coverage is honestly better than I usually see for a migration this risky. Credit for the self-review in the description too — most of the trade-offs are already called out.

Here's what I'd want fixed before it goes in.

## Blocking

**1. `upstream_api_key` can get silently dropped.** It's now in `SECRET_FIELDS` (`settings.py`), so it gets stripped from the persisted blob — but unlike `nsec`, nothing migrates it into encrypted storage. So for a node that only has it in the DB blob: it loads into memory once, the blob gets rewritten without it, and on the next restart it's gone and upstream auth breaks. Either don't strip it in this PR, or give it a real encrypted home like `nsec` got. This is the one that worries me most.

**2. Stored `nsec` can lose its live `npub`.** Bootstrap decrypts `nsec` and derives `npub`, but `SettingsService.initialize()` can later overwrite the live `npub` back to empty if the blob/env doesn't carry one. You end up with a private key in memory and no matching public key. I hit this locally — `npub` was right after bootstrap, empty after initialize. Worth a test for exactly this case.

**3. Generated admin password goes through `logger.warning()`.** The message says "shown only now," but logging usually means it also lands in app log files / collectors / backups. Anyone with log access gets admin until it's rotated. I'd print it straight to stdout and keep it out of the normal logger — or at least fix the wording so it's honest about the risk.

**4. Leftover `print(payload)` in `admin.py:505`** (`upsert_provider_model`) — dumps every upsert to stdout. Just remove it.

**5. `admin_login` touches the ORM object after the session closes.** Works today because the field's already loaded, but it's fragile — a future SQLAlchemy change turns it into a detached-instance error. Copy the hash into a local `str` inside the `with` block.

## Security — fix now or ticket clearly

**6. Login has no rate limit but now runs scrypt.** Good for storage, but `/admin/api/login` is unauthenticated and every attempt is now CPU/memory work — so it doubles as a brute-force and a DoS amplifier. You already flag this as pre-existing scope, which is fair, I just don't want it forgotten. A small per-IP limit here would be ideal; an issue is the minimum.

**7. scrypt params read from the stored hash aren't capped** (`vault.py`). Flexible format, but a tampered DB value could ask for huge N and OOM the box on the next verify. Reject anything above the module constants — small fix, worth doing now.

## Smaller stuff (not all blockers)

- README / `.env.example` say the node won't start without `ROUTSTR_SECRET_KEY`, but the code only needs it when touching `nsec`. Enforce it at startup or soften the docs.
- `reset_admin_password.py --password ...` leaks the password into shell history / process args — add `--password-stdin` or prompt with `getpass`.
- `update_password` strips the new password but login doesn't strip the entered one. Pick one rule. I'd lean toward not silently stripping passwords at all.
- Failed admin logins aren't logged — a warning with client IP would help.
- `/api/setup` has a small race if two first-setup requests land together.
- `bootstrap_secrets()` is big — splitting into `_bootstrap_admin_password()` / `_bootstrap_nsec()` would make it easier to follow.
- `get_fernet()` could be private if it's only used inside `vault.py`.

## Not blocking, on purpose

Key rotation, upstream keys still being plaintext in `upstream_providers`, and full login rate limiting are all fine as follow-ups as long as they get tracked — none of them need to hold this PR.

## On splitting it

I wouldn't force a split. It's big, but it's one coherent feature — vault, table, bootstrap, API all need each other, and splitting would just mean merging dead code and tripling the review cycles. The one cheap win, if you want it, is pulling `vault.py` + its tests into a separate PR first since it's self-contained crypto with no behavior change — review that carefully on its own, then merge the rest after the fixes. But merging it as one piece is totally fine by me.

I did run the focused tests + ruff/mypy/lint locally and everything passed, so this is about the design-level issues above, not the build.

Happy to pair on any of these if it's easier.

9qeklajc:
Re-reviewed after the latest push — looks like just a formatting tweak on `updateNsec`, so everything above still applies. One thing I want to expand on since it's the part that'll actually bite operators on upgrade: **backward compatibility.**

Right now an existing node doesn't upgrade cleanly. Three things break, and all three are fixable without giving up the security win:

**1. Nodes with an existing `nsec` but no `ROUTSTR_SECRET_KEY` crash on boot.** Bootstrap takes the legacy-nsec path → `vault.encrypt()` → "key not set" `RuntimeError`. A node that ran fine yesterday won't start. I'd make encryption *opt-in* instead of mandatory: if the key isn't set, keep the nsec live in memory (old behavior) and just log a "set the key to enable encryption at rest" warning. Only keep the hard fail for the *already-encrypted* branch, where a wrong/missing key really does mean unreadable data.

```python
legacy_nsec = _legacy_plaintext(raw_blob, "NSEC", "nsec")
if legacy_nsec:
    if not os.environ.get("ROUTSTR_SECRET_KEY"):
        settings.nsec = legacy_nsec  # legacy mode: in-memory only, like before
        logger.warning("nsec present but ROUTSTR_SECRET_KEY not set; running "
                       "without encryption at rest. Set the key to enable it.")
    else:
        secret.encrypted_nsec = vault.encrypt(legacy_nsec)
        settings.nsec = legacy_nsec
        changed = True
```

**2. `upstream_api_key` gets silently dropped** (the data-loss bug from my first pass). Simplest backward-compatible fix: just **don't add it to `SECRET_FIELDS` in this PR**. Leave it in the blob exactly as today, and do the encrypted-storage migration for it in a follow-up. That removes the regression entirely with one line.

**3. `npub` can go empty after bootstrap.** Same skip-guard you already use for secret fields — if the incoming `npub` is empty but a live one exists, keep the live one (or just re-derive from `nsec` after `initialize()`). Plus a regression test.

`ADMIN_PASSWORD` seeding already handles the password case well — existing passwords survive the first boot, so no lockout there.

Net: with #1 and #2, an existing node upgrades with **zero config changes and zero data loss**, and `ROUTSTR_SECRET_KEY` + upstream-key encryption become opt-in improvements rather than upgrade blockers. That's the version I'd be comfortable shipping.

One more nicety: nothing currently enforces "set the key *before* upgrading." Even with the graceful-degrade above, a clear one-line startup hint (what to set, how to generate it) goes a long way for anyone who upgrades without reading the release notes.

jeroenubbink:
Thanks for the thorough pass — this was genuinely useful. Pushed fixes for the blocking + security-hardening items; status against your list:

**Blocking**
- **1. `upstream_api_key` silently dropped** — fixed in `2f77f35`. Removed it from `SECRET_FIELDS` so it stays in the blob exactly as today; the encrypted-storage migration for it is a follow-up. (Agree it's the scariest one — it's also conceptually a per-provider field, not a singleton setting, so it shouldn't live in `SECRET_FIELDS` at all yet.)
- **2. Stored `nsec` loses live `npub`** — fixed in `af2ee94`. `initialize()` now re-derives `npub` from the live `nsec` when the merged settings carry none, so a store-only nsec keeps its public key. Added a regression test for exactly the bootstrap→initialize sequence you hit.
- **3. Generated admin password via `logger.warning`** — fixed in `c5815d5`. Now printed straight to stdout with `flush=True`, kept out of the logger so it never lands in the on-disk log files / collectors.
- **4. Leftover `print(payload)`** — removed in `c5815d5`.
- **5. `admin_login` touches detached ORM object** — fixed in `c5815d5`. The hash is copied into a local `str` inside the `with` block before the session closes.

**Security**
- **6. Login rate limit / scrypt DoS amplifier** — agreed, tracking as a follow-up rather than holding this PR; a per-IP limit on `/admin/api/login` is the right home for it.
- **7. scrypt params not capped** — fixed in `c5815d5`. `verify_password` now rejects any `N`/`r`/`p` above the module's own parameters, so a tampered hash can't force an unbounded work factor (memory ~ N·r) and OOM the box on verify.

**Smaller stuff / not-blocking** — the doc-wording, `--password-stdin`/getpass, strip-consistency, failed-login logging, `/api/setup` race, `bootstrap_secrets` split, and `get_fernet` visibility are all noted; I'll fold the trivial ones in before merge and ticket the rest with the key-rotation / upstream-key-encryption follow-ups so none get lost.

**On splitting** — appreciate you not forcing it. I'd like to keep it as the single PR. Pulling `vault.py` + tests out first would only merge a dead-code foundation: with no caller it carries no context, can't be exercised end-to-end, and is exactly the kind of thing that ends up needing changes (then reverts) once the bootstrap/table/API that actually use it land on top. The commit history here is already split into small logical steps for review, which gets the reviewability win without the merge-dead-code cost. The opt-in / backward-compat point I've answered separately below.

jeroenubbink:
On the backward-compat follow-up specifically — fully agree on **item 2 (`upstream_api_key`, fixed in `2f77f35`)** and **item 3 (`npub`, fixed in `af2ee94` + regression test)**, and on the "set the key before upgrading" nicety: bootstrap now emits a clear, nsec-contextual startup error with the generate command inline ([`be9b157`](https://github.com/Routstr/routstr-core/pull/570/commits/be9b1574a048ae342259b8c40a7ffdce0026c040)).

Where I want to push back is **item 1 — making encryption opt-in.** Let me steel-man it first, because the upgrade-friction worry is real: an existing node running with an `NSEC` and no `ROUTSTR_SECRET_KEY` will refuse to boot until the operator sets the key. Failing a security feature closed is exactly what pages someone at 2am, and opt-in (warn, keep plaintext, boot) avoids that.

I still think mandatory is the right call:

1. **It's not a data-loss trade — it's a pure, reversible availability event.** On the missing-key path we don't strip, drop, or mutate anything: the nsec stays exactly where it was (env / blob), untouched, until a key is set. The node just declines to start; recovery is one env var + restart. The usual reason to prefer warn-and-continue — don't break the running thing, don't lose state — doesn't apply, because nothing is at stake but a restart.

2. **Opt-in is a one-way ratchet toward the insecure state.** The whole point of this work is keeping the nsec off disk in plaintext. If the default is "boot plaintext and warn," the default population runs unencrypted — and those operators are the least likely to flip a flag later. Once a node has booted plaintext the nsec is already on disk (plus WAL, plus backups), so "encrypt it later" actually means re-encrypt *and* scrub the plaintext copies. In practice that never happens. Secure-by-default only holds if secure is the *only* default.

3. **The fail-fast is intentional and actionable, not an incidental throw.** `vault.encrypt` would already raise, but [`be9b157`](https://github.com/Routstr/routstr-core/pull/570/commits/be9b1574a048ae342259b8c40a7ffdce0026c040) adds an explicit error on the legacy-nsec path *before* it, so the operator gets what's wrong, why, and the fix in one shot:
   ```
   An nsec is configured but ROUTSTR_SECRET_KEY is not set. The key is
   required to encrypt the Nostr identity at rest. Generate one and set it
   in the environment:
       python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
   ```

4. **It only triggers when there's actually a secret to protect.** A node with no nsec boots normally (it just generates the admin password — `ADMIN_PASSWORD` seeding is unaffected, as you noted). We hard-stop *only* when a real Nostr identity would otherwise land on disk in plaintext.

5. **Opt-in is also more surface, not less.** Encryption is already mandatory everywhere `vault.encrypt` is used; an opt-in nsec path means building and maintaining a *new* plaintext-nsec branch forever, with its own tests. Mandatory is the smaller, more auditable design.

So the cost is a one-time `+1 env var` on upgrade, surfaced loudly with the fix inline and called out as a breaking change in the PR body; the benefit is that no default deployment ever silently keeps its nsec in plaintext. For a feature whose entire purpose is getting the key off disk, I think that's the right trade. If upgrade discoverability is the weak point, I'm happy to make the README / `.env.example` breaking-change note louder — that part I'll gladly take.

9qeklajc:
nice, will run some more tests with old and new nodes and see if everything works as expected

jeroenubbink:
rebased because of conflicts.

migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
```diff
@@ -0,0 +1,42 @@
...
+any legacy plaintext into the encrypted/hashed columns happens at bootstrap,
+where the live ROUTSTR_SECRET_KEY is available.
+"""
+
+import sqlalchemy as sa
+import sqlmodel
+from alembic import op
+
+revision = "c6f8d2e4a1b3"
+down_revision = "b5e7c9d1f3a2"
```
9qeklajc: This looks like it branched from an older migration. `c6d7e8f9a0b1` already uses `b5e7c9d1f3a2` as its parent, so this PR leaves Alembic with two heads and startup `upgrade head` will fail. I think this should point at the current head instead: `down_revision = "c6d7e8f9a0b1"`.

routstr/core/settings.py
```diff
@@ -132,10 +133,70 @@ def _normalize_settings_data(data: dict[str, Any]) -> dict[str, Any]:
...
+    return {k: v for k, v in data.items() if k not in SECRET_FIELDS}
+
+
+def _apply_to_live_settings(data: dict[str, Any]) -> None:
+    """Apply ``data`` onto the live ``settings`` for all in-process importers.
+
+    Secrets are owned by ``bootstrap_secrets`` (which decrypts the nsec into
+    memory before this runs) — they are never persisted to the blob, so ``data``
+    re-derived from the secret-free blob carries empty secret values. Skip those
+    empty overwrites so a live secret is never clobbered; a non-empty value
```
9qeklajc: I think this still lets a stale `NSEC` env var win after the vault has taken ownership. Boot decrypts the DB value into `settings.nsec`, but later `initialize()` starts from env and this helper applies any non-empty secret back onto live settings. So if an operator rotates nsec in the UI but leaves the old `NSEC` in `.env`, restart silently goes back to the old key. Once `encrypted_nsec` exists, env/blob `nsec` probably shouldn't be applied to live settings anymore.

routstr/core/db.py
```diff
@@ -392,6 +407,52 @@ async def get_routstr_fee(session: AsyncSession) -> RoutstrFee:
...
+async def set_nsec(session: AsyncSession, nsec: str) -> None:
+    """Store the node's nsec, Fernet-encrypted, on the Secret singleton.
+
+    An empty string clears it (the node then holds no Nostr identity and signs
+    no events).
+    """
+    from .vault import encrypt
+
+    secret = await get_secret(session)
+    secret.encrypted_nsec = encrypt(nsec) if nsec else None
```
9qeklajc, line 563: Clearing the nsec stores `None`, but that makes the next boot look the same as “never migrated”. If legacy `NSEC` is still in env/blob, `bootstrap_secrets()` will import it again and the cleared identity comes back. Could we store an explicit “cleared/migrated” state, or otherwise mark that legacy import has already been consumed?

9qeklajc (commented):
Nice work moving the sensitive bits out of the settings blob — the overall shape makes sense and the tests around hashing/encryption are helpful.

I found two things I think need fixing before merge, plus one smaller edge case around clearing the nsec. Kept the comments short inline.

migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
jeroenubbink: Fixed in [`65f6cbb`](https://github.com/Routstr/routstr-core/pull/570/commits/65f6cbbf808640900e1093cbf508318df20a671d) — rebased onto current main and repointed `down_revision` to `c6d7e8f9a0b1`. `alembic heads` now resolves to the single head `c6f8d2e4a1b3` and `upgrade head` runs clean on a fresh DB.

routstr/core/settings.py
jeroenubbink: Fixed in [`feafcaf`](https://github.com/Routstr/routstr-core/pull/570/commits/feafcaf) — `_apply_to_live_settings` no longer re-applies *any* secret field. `bootstrap_secrets` runs first and already holds the authoritative decrypted nsec, so it's now the sole writer of live secret values; a stale `NSEC` left in `.env` can no longer clobber a vault-owned nsec on restart. Guarded by `test_stale_env_nsec_does_not_override_vault_nsec`.

routstr/core/db.py
jeroenubbink: Fixed in [`feafcaf`](https://github.com/Routstr/routstr-core/pull/570/commits/feafcaf) — added a `secrets.nsec_managed` column that records the vault has taken ownership of the nsec (set on legacy import and on every `set_nsec` write, clears included). Once set, bootstrap skips the legacy import, so an nsec cleared via the admin API stays cleared instead of being resurrected from a stale env/blob copy — an empty `encrypted_nsec` alone couldn't tell "cleared" from "never migrated". Guarded by `test_cleared_nsec_stays_cleared_across_reboot`.

README.md
```diff
@@ -55,19 +55,34 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
...
-   ADMIN_PASSWORD=mysecretpassword
+   # Required: encrypts secrets at rest. The node won't start without it.
+   ROUTSTR_SECRET_KEY=<generated-key>
    NAME="My AI Node"
    DESCRIPTION="Fast access to models"
    NSEC=yournsec
    RECEIVE_LN_ADDRESS=yourname@wallet.com
    ```
 
+   Generate `ROUTSTR_SECRET_KEY` once and keep it stable — changing it makes
```
9qeklajc: sorry but really still think that breaking running nodes because of the Secret has to be set is a problem ... 
can we do it differently ? can't think of another way of doing it but maybe u have one (split?)

routstr/core/settings.py
```diff
@@ -3,6 +3,8 @@
 import asyncio
```
9qeklajc, line 3: If a node has an encrypted nsec in the vault but still has an older NSEC value in the environment, startup can keep the vault
 nsec while deriving/applying npub from the stale env value.

 That leaves the node with:

 - private key from vault
 - public key from old env

 Anything exposing settings.npub can then publish the wrong identity.

 Please make initialization derive npub from the live effective nsec whenever the vault owns the nsec, and add a regression
 test that covers stale env NSEC plus vault-owned nsec.

README.md
```diff
@@ -55,19 +55,34 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
...
+   previously encrypted secrets unreadable:
```
9qeklajc: The new setup/reset docs use plain python, but that only works if the host has the right interpreter and dependencies
 installed globally. On a clean machine, python may not exist, or cryptography may not be installed.



Please document commands that run inside the project/container environment instead, for example:

 ```bash
   uv run python -c "from cryptography.fernet import Fernet; print(Fernet.generate_key().decode())"
   uv run python scripts/reset_admin_password.py --regenerate
 ```

 For Docker users, reset should use the container Python:

 ```bash
   docker compose exec routstr /.venv/bin/python scripts/reset_admin_password.py --regenerate
 ```

 Also make sure any compose file that sets NSEC also sets ROUTSTR_SECRET_KEY, otherwise startup fails by design.

README.md
```diff
@@ -55,19 +55,34 @@ If you are a node runner, start a Routstr Core instance using Docker Compose:
```
9qeklajc, line 55: we can update the node setup also in the doc/ too

9qeklajc (changes requested):

README.md
jeroenubbink: Ok so to accommodate this is now changed to generate the string in `routstr_secret.key` with mode 0600 if no `ROUTSTR_SECRET_KEY` is set and *URGES* the end user to make a backup of the file. By default the key file is written next to the database, so it rides whatever volume already persists your data (a container recreate will not lose it); set `ROUTSTR_SECRET_KEY` explicitly to keep it off the data volume. Provided he has either stored all other secrets elsewhere and makes a backup of the key as plain text or makes a complete file backup this should be fail-safe.

routstr/core/settings.py
jeroenubbink: Found and fixed!

README.md
jeroenubbink: Updated this in the docs

README.md
jeroenubbink: updated a lot in the docs about secret key usage, where it is stored, that it needs backups and where it lives in the application by default and also in docker.

routstr/core/settings.py
```diff
@@ -355,3 +418,120 @@ async def reload_from_db(cls, db_session: AsyncSession) -> Settings:
...
+    if secret.encrypted_nsec is not None:
+        try:
+            settings.nsec = vault.decrypt(secret.encrypted_nsec)
+        except InvalidToken as exc:
+            raise RuntimeError(
+                "Stored nsec cannot be decrypted with the current "
+                "ROUTSTR_SECRET_KEY. The key changed, or this database came from "
+                "another node. Restore the original ROUTSTR_SECRET_KEY to recover."
+            ) from exc
+    elif not secret.nsec_managed:
```
9qeklajc: <s>
🔴 **Blocker — a cleared nsec is resurrected after restart.** When this row is managed but has no ciphertext, this branch is skipped, but the process-global settings may already contain the stale `NSEC`/derived `npub` loaded from the environment. The supposedly cleared identity then becomes active again. Could you kindly add an explicit managed-and-empty branch that clears both live fields?

Please cover: (1) stale live `nsec` + `npub`, managed row, no ciphertext → both become empty; (2) a fresh-process/lifespan-equivalent restart rather than pre-clearing the singleton; (3) managed ciphertext still decrypts/derives npub; and (4) unmanaged legacy nsec imports exactly once.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

routstr/core/vault.py
```diff
@@ -0,0 +1,264 @@
...
+    # ``docker compose logs``), but it must never be persisted into the on-disk
+    # log files the logger also writes. Mirrors the generated-admin-password
+    # notice so an upgrade cannot silently create an unbacked key.
+    print(
+        "No ROUTSTR_SECRET_KEY was set; generated one to encrypt node secrets at "
+        f"rest and saved it to {path}.\n"
+        "!! BACK UP THIS FILE. If it is lost, the encrypted secrets cannot be "
+        "recovered and will have to be re-entered.\n"
+        "To manage the key yourself (e.g. from a secrets manager) set it in the "
+        f"environment instead:\n    ROUTSTR_SECRET_KEY={key}",
```
9qeklajc: <s>
🟡 **Please don’t print the master key into stdout/container logs.** Log access is commonly broader than DB/key-volume access, so this gives log readers the ability to decrypt a later DB backup. The durable key file already provides recovery. Could the notice contain only the resolved path, required mode, and backup warning?

Please test that captured stdout excludes the generated key value, includes the path/backup warning, and that a new vault/process instance can still decrypt using the persisted file.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

routstr/core/vault.py
```diff
@@ -0,0 +1,264 @@
...
+    override = os.environ.get(_KEY_FILE_ENV)
+    if override:
+        return Path(override)
+    directory = _database_dir()
+    return (directory or Path()) / _DEFAULT_KEY_FILE
+
+
+def _read_key_file(path: Path) -> str | None:
+    try:
+        stored = path.read_text().strip()
```
9qeklajc, line 100: <s>
🟡 **Existing key files need a permissions contract.** A manually restored or accidentally chmodded `0644` file is currently accepted, exposing the nsec decryption key to other local principals. On POSIX, could this reject group/other permission bits or safely correct to `0600` with an explicit warning?

Please cover valid `0600`, permissive `0644`, malformed-but-private key content, and env-key precedence (which should not depend on file mode).
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

routstr/core/vault.py
```diff
@@ -0,0 +1,264 @@
...
+        flush=True,
+    )
+
+
+def _generate_and_persist_key(path: Path) -> str:
+    """Generate a Fernet key, persist it owner-only, and warn once."""
+    key = Fernet.generate_key().decode()
+    path.parent.mkdir(parents=True, exist_ok=True)
+    try:
+        fd = os.open(path, os.O_CREAT | os.O_EXCL | os.O_WRONLY, 0o600)
```
9qeklajc: <s>
🟡 **A crash can strand an empty/truncated final key file.** The final path is created before the key is durably written; after interruption, the next attempt sees `FileExistsError`, rereads no usable value, and cannot recover. Could this fully write/fsync a same-directory `0600` temp file, atomically install it without overwriting a concurrent winner, and fsync the directory?

Please test empty-final-file recovery/error handling, simulated write/fsync/install failure with no unusable final key, two concurrent creators converging on one key, and preservation of a valid pre-existing winner.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

routstr/core/settings.py
```diff
@@ -355,3 +418,120 @@ async def reload_from_db(cls, db_session: AsyncSession) -> Settings:
...
+
+    # Admin password — one-way scrypt hash.
+    if secret.admin_password_hash is None:
+        legacy_password = _legacy_plaintext(
+            raw_blob, "ADMIN_PASSWORD", "admin_password"
+        )
+        if legacy_password:
+            secret.admin_password_hash = vault.hash_password(legacy_password)
+        else:
+            generated = secrets.token_urlsafe(24)
```
9qeklajc, line 488: <s>
🟡 **Concurrent first boot can print a password that never becomes valid.** Two workers can both observe a missing hash, generate/print different passwords, and race commits; only one resulting hash works. Could initialization be claimed atomically in the DB, with output emitted only by the transaction that wins and commits?

Please run two concurrent bootstraps and assert exactly one notice, verify that sole password against the stored hash, ensure the loser does not overwrite the winner, and confirm legacy configured-password migration prints no generated password.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

docs/provider/deployment.md
```diff
@@ -156,10 +156,21 @@ Example `.env`:
...
+# your database (on the same volume) and printed once. Set it explicitly to
+# manage the key yourself.
+ROUTSTR_SECRET_KEY=
 NAME=My Provider Node
 RECEIVE_LN_ADDRESS=me@walletofsatoshi.com
 ```
 
+!!! note "Secret key persistence"
+    If you leave `ROUTSTR_SECRET_KEY` unset, the node generates one and stores it
+    as `routstr_secret.key` **next to your database**, so it persists on the same
```
9qeklajc, line 174: <s>
🟡 **This persistence guarantee doesn’t follow from the shown compose example.** It mounts `/app/data`, but does not set `DATABASE_URL` beneath that directory, so the generated key is not necessarily on the advertised volume. Could the example make the DB URL and mount agree?

Please validate that the DB and generated key resolve under the mounted data directory, that a container recreation with the same volume can still decrypt existing ciphertext, and document/test the explicit key-file override separately.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

routstr/core/admin.py
```diff
@@ -239,52 +242,92 @@ async def update_settings(request: Request, update: SettingsUpdate) -> dict:
...
-        raise HTTPException(status_code=409, detail="Admin password already set")
-    pw = (payload.password or "").strip()
-    if len(pw) < 8:
-        raise HTTPException(
-            status_code=400, detail="Password must be at least 8 characters"
-        )
     async with create_session() as session:
-        await SettingsService.update({"admin_password": pw}, session)
+        secret = await get_secret(session)
+        if secret.admin_password_hash:
```
9qeklajc: <s>
🟠 **Could we make the first-run API contract explicit?** Normal lifespan startup now creates an admin hash before serving requests, so this setup endpoint appears to return 409 on every reachable fresh-node request; the success tests bypass lifespan bootstrap. Please decide whether generated-password bootstrap replaces setup (remove/deprecate this path) or setup remains supported (preserve a setup window).

Whichever contract is chosen, please test it through the real application lifespan, verify restart cannot replace an existing password, confirm login through the selected first-run flow, and cover concurrent setup/bootstrap attempts.
</s>

**Superseded by the staged compatibility plan in the follow-up review.** Please keep these as regression scenarios, but do not implement them as one all-at-once legacy replacement. The immediate scope is Release 1: explicit ownership state, a key-provider boundary, hardened local key handling, no key material in logs, and no admin-flow breakage.

9qeklajc (dismissed):
<s>
The vault ownership marker, migration chain, npub derivation, redaction, and upgrade path are moving in a good direction. The focused and full validation suites are green. I did find one restart blocker plus several security/durability issues in the new provisioning path. Could you kindly resolve the inline concerns and add the listed regression scenarios before merge? The setup endpoint also needs an explicit API-contract decision.
</s>

9qeklajc (commented):
I digged  a bit more on how litellm works and if we could inspire our selfs from there.. 


The vault ownership marker, versioned ciphertext, migration chain, npub derivation, redaction, and upgrade work are solid foundations. After comparing this approach with LiteLLM’s secret-management design, I think the cleanest path is not to replace the legacy behavior in this PR all at once. Instead, please combine Routstr’s operator-friendly local-key UX with LiteLLM’s stable key-provider and migration patterns through a staged, backward-compatible rollout.

## Superseded point-by-point direction

The following findings remain useful as risks and test cases, but the earlier request to solve all of them as one immediate replacement is superseded by the staged plan below.

<s>

1. Treat the stale-`NSEC` restart behavior as an isolated blocker without first defining explicit secret ownership states.
2. Replace key logging immediately without introducing a recovery/download path for node runners.
3. Add local-file permission checks directly to the vault without separating key retrieval from encryption.
4. Harden local key creation in place without first introducing a key-provider boundary.
5. Change generated-admin-password behavior in the same release as secret-storage migration.
6. Repair persistence documentation before defining the supported local and externally managed key modes.
7. Decide the `/admin/api/setup` replacement contract as part of the vault migration.

</s>

## Recommended combined design

### Keep the good parts of each approach

From LiteLLM, adopt:

- a stable externally supplied encryption-key contract;
- a small pluggable key-provider abstraction;
- versioned ciphertext and format-detecting reads;
- resumable, idempotent migration tooling with `check` and `dry-run` modes;
- redaction before secrets reach API/UI responses.

From Routstr, keep:

- automatic local-key provisioning as the simple default for self-hosted node runners;
- encrypted nsec storage with an explicit ownership marker;
- a one-time recovery-key file download in the admin UX;
- a persistent backup warning until the runner confirms responsibility.

Improve on LiteLLM by making persisted vault state authoritative. Once a secret is encrypted or explicitly cleared, stale environment/blob values must never silently take ownership again.

## Target architecture

### 1. Separate encryption from key custody

The vault should encrypt/decrypt with supplied key material. Key lookup, generation, download, and external-manager integration belong behind a provider boundary:

```python
class KeyProvider(Protocol):
    async def get_key(self) -> KeyMaterial: ...
```

Initial providers:

- `EnvironmentKeyProvider`
- `LocalFileKeyProvider`

Future providers can include AWS Secrets Manager, HashiCorp Vault, Kubernetes Secrets, or another operator-managed backend without changing ciphertext or secret ownership logic.

Resolution order:

1. explicitly configured external/environment provider;
2. existing local key file;
3. atomically create a local key only when encryption is first required.

The vault must never print, download, or persist key material itself.

### 2. Use explicit secret ownership states

Do not infer lifecycle solely from nullable ciphertext. Use an explicit state such as:

```text
legacy | encrypted | cleared
```

Startup behavior:

| State | Required behavior |
|---|---|
| `legacy` | Read the old env/blob value once, encrypt and verify it transactionally, then mark `encrypted`. |
| `encrypted` | Decrypt from the vault. Never fall back to env/blob. |
| `cleared` | Explicitly clear live `nsec` and `npub`. Never consult env/blob. |

This fixes the stale-`NSEC` restart bug while preserving a safe migration source for existing nodes.

### 3. Support two operator modes

**Routstr-managed mode** — default for simple/self-hosted nodes:

- create a local key crash-safely with owner-only permissions;
- never print the key;
- offer one authenticated recovery-file download;
- retain the server’s working copy after download;
- show a persistent warning until backup confirmation.

**Operator-managed mode** — for production deployments:

- inject the key through environment/orchestration initially;
- later support external secret-manager providers;
- do not offer a recovery download;
- display only source type and key fingerprint in the UI.

Store key metadata separately from key material:

```text
key_source: local_file | environment | external_manager
backup_status: pending_download | downloaded_unconfirmed | confirmed | externally_managed
```

## Staged release plan

### Release 1 — compatibility and safety

Keep existing nodes running and make only additive changes:

1. Add explicit secret ownership state while preserving current schema/data.
2. Fix `cleared` startup semantics so stale env/blob values cannot resurrect nsec or npub.
3. Introduce `KeyProvider` behind the current behavior without changing ciphertext.
4. Harden `LocalFileKeyProvider`: `0600`, temporary same-directory write, file and directory `fsync`, atomic winner-preserving install, and clear handling of empty/corrupt files.
5. Stop printing key material. Print only the path, fingerprint, and backup instruction.
6. Keep the existing admin setup/login contract unchanged in this release.
7. Add diagnostics, but do not block upgraded nodes on backup confirmation.

### Release 2 — recovery UX

Add the node-runner workflow without changing existing-node availability:

1. Add authenticated one-time recovery-file download for locally managed keys.
2. Use `Cache-Control: no-store`, re-authentication, a short-lived purpose-bound token, and audit metadata without the key.
3. Require download plus fingerprint confirmation for **new-node setup**.
4. For upgraded nodes, continue serving traffic and show a persistent high-priority backup banner until confirmation.
5. Never place the key in JSON, URLs, logs, analytics, browser storage, or email.

### Release 3 — migration and observability

1. Add `routstr secrets check`.
2. Add `routstr secrets migrate --dry-run` and an idempotent migration command.
3. Report ownership state, key source, fingerprint, permissions, backup status, decryptability, and residual legacy values.
4. Stop writing legacy secret fields, but continue reading them only for unmigrated `legacy` rows.
5. Add migration telemetry/diagnostics that contain no secret values.

### Release 4 — contract legacy behavior

Only after a documented deprecation window and migration evidence:

1. Remove legacy secret writes permanently.
2. Remove legacy reads when residual checks show nodes have migrated.
3. Add external secret-manager providers as needed.
4. Consider a new ciphertext version in a separate migration; keep `fernet:v1:` readable until its own deprecation completes.
5. Revisit `/admin/api/setup` independently rather than coupling that API change to vault rollout.

## Required compatibility and regression scenarios

### Ownership and migration

- Upgrade a real pre-vault database with env-only, blob-only, and absent nsec values.
- Verify migration encrypts, decrypts, compares, commits, and retries safely after interruption.
- Verify `encrypted` always outranks stale env/blob values.
- Verify `cleared` empties live nsec/npub after a fresh restart and never reimports legacy values.
- Verify rollback before migration commit leaves the legacy value usable.

### Local key provider

- Existing valid `0600` key is reused.
- Permissive key permissions are rejected or safely corrected according to the documented contract.
- Interrupted write/install never publishes an unusable final key.
- Empty/corrupt files produce actionable recovery behavior.
- Concurrent workers converge on one key and never overwrite a valid winner.
- Generated key material never appears in captured logs or API responses.

### Recovery UX

- New-node setup requires successful download and fingerprint confirmation.
- Existing upgraded nodes continue serving while backup status is pending.
- A failed download has a narrowly bounded retry path and does not leak the key.
- Confirmation stores only fingerprint, actor, and timestamps.
- Externally managed keys never expose a download action.
- Container recreation with the same volume can decrypt existing ciphertext.

### Transition and contracts

- Existing admin credentials and login behavior remain unchanged through Releases 1–3.
- Ciphertext written by every supported version remains readable during the transition.
- Migration `check` and `dry-run` are read-only; repeated migration runs are idempotent.
- Legacy reads occur only for explicit `legacy` state and disappear only in the final contract release.

## Bottom line

Could you kindly use this staged plan as the implementation direction? It preserves the working parts of the current PR, adopts LiteLLM’s clean provider/versioning/migration model, gives node runners a straightforward recovery download, and avoids forcing every running node through a breaking secret-storage and admin-setup replacement in one release.

The immediate goal should be **Release 1: compatibility and safety**, not the complete end-state. Once that lands with upgrade tests, the recovery UX and eventual legacy removal can follow in separate, reviewable releases.

routstr/core/settings.py
jeroenubbink: Confirmed and fixed — good catch. The `nsec_managed` bool couldn't separate "never migrated" from "cleared" (both leave `encrypted_nsec` empty), so the cleared case fell through and a stale env `NSEC` could re-derive the identity on a fresh process; the old test pre-cleared the live singleton, so it never exercised that path. Replaced with an explicit `nsec_state` (`legacy | encrypted | cleared`) that is the single source of truth: bootstrap branches on it — `encrypted` decrypts (a missing ciphertext is a fail-fast inconsistency, never a silent fall-through to legacy), `cleared` actively empties the live nsec *and* npub, `legacy` imports once. New regression simulates a fresh process with a stale env value still present and covers the scenarios you listed. Fixed in https://github.com/Routstr/routstr-core/pull/570/commits/04f3147ab66e6ec75df301503133150dc22b1b2f.

routstr/core/vault.py
jeroenubbink: Agreed — the durable 0600 file is the recovery path, and logs are a broader surface than the key volume. The notice now prints the path and the backup imperative only, never the key value. https://github.com/Routstr/routstr-core/pull/570/commits/a4b65c86334789d84cbb6db374da953897b03f16

routstr/core/vault.py
jeroenubbink: Added a permission check on read — on POSIX we repair group/other bits back to 0600 (repair rather than hard-fail, so an upgrading node with a loose-umask key keeps booting) with the key still usable; env-key precedence stays independent of file mode. Covered by a new test that a 0644 key file is tightened to 0600 on read. https://github.com/Routstr/routstr-core/pull/570/commits/a4b65c86334789d84cbb6db374da953897b03f16

routstr/core/vault.py
jeroenubbink: Right — a crash could strand an empty final file and brick the next boot. Now writing a same-dir 0600 temp file, fsync, then `os.link` into place, then fsync the directory; the temp is cleaned up on any failure. I went with `os.link` over `os.replace` because it is atomic *and* refuses to clobber: a racing worker that published its key first keeps ownership (secrets may already be encrypted under it) and the loser adopts that key rather than overwriting it — `os.replace` would silently clobber. Both the crash path and the adopt-the-winner race have regression tests now. One tradeoff to flag: `os.link` needs a filesystem that supports hardlinks — that holds for the standard Docker data volume, but I'm happy to add an `os.replace` fallback for exotic volumes if you'd prefer belt-and-suspenders. https://github.com/Routstr/routstr-core/pull/570/commits/a4b65c86334789d84cbb6db374da953897b03f16

routstr/core/settings.py
jeroenubbink: Fixed in-PR. The generated password is now claimed atomically via a conditional `UPDATE ... WHERE admin_password_hash IS NULL`, and only the worker that wins the update (rowcount 1) prints it. A racing worker on a shared DB adopts the winner's hash and stays silent, so the operator never sees a second password that was never stored. A deterministic regression test forces the race window. https://github.com/Routstr/routstr-core/pull/570/commits/04f3147ab66e6ec75df301503133150dc22b1b2f

docs/provider/deployment.md
jeroenubbink: Fixed the examples so `DATABASE_URL` and the mount agree — the database and the key file generated beside it both resolve under the mounted data volume, and `routstr_secret.key` is now listed in the persistence table. https://github.com/Routstr/routstr-core/pull/570/commits/2d8f940e0518b8fd8880e01d88ef32750a7c611a

routstr/core/admin.py
jeroenubbink: Removed it. Generated-password bootstrap now sets a password on every fresh node, so `/api/setup`'s success path is unreachable — it 409s on any booted node — and grepping the admin UI, routstr-cli, routstr-sdk, routstrd, and routstr-chat turned up zero callers, so I dropped the orphaned endpoint rather than leave dead surface. First-run is now: generated password printed once at boot -> log in at `/admin` -> change it from the dashboard. https://github.com/Routstr/routstr-core/pull/570/commits/d6c513c41a7a60697b623212344dabdd70254626

jeroenubbink:
Thanks for digging into LiteLLM — the staged, backward-compatible instinct is right and we're adopting it. One framing that I think dissolves most of the disagreement: we treat the **key** and the **secrets** as separate layers. The secrets (nsec, hashed admin password, provider keys) live encrypted at rest in the node's own DB — that's fixed. The only thing that varies is where the **master key** that decrypts them comes from, and that has exactly one variation point that matters: an env var, or a local file beside the DB.

That's why I'd rather not add a `KeyProvider` abstraction or external secret-manager providers here. `ROUTSTR_SECRET_KEY` in the environment already *is* the external-manager seam — Vault, AWS/GCP/Azure, and K8s all render a secret to an env var at boot, so we integrate with every one of them at zero code and no runtime dependency. A pluggable provider for two implementations that are already two lines (env → file) mostly adds surface. And pushing the *secrets themselves* to an external store would make the node unable to boot/sign/proxy without that store reachable — a real availability regression for a single-operator self-hosted node (a Nostr-discovered node run via `docker compose`, not a central gateway). LiteLLM's provider model fits *its* context — a central enterprise gateway run by a platform team that already operates Vault — a different place in the ecosystem than a routstr node.

Similarly I'd keep the `routstr secrets` CLI and the two operator modes out: issue 553 deliberately picked "exactly one way to configure each thing" and "no CLI, it's a Docker project," so those pull against the founding constraint.

What we ARE taking, because it's portable and right: explicit ownership state, a hardened local key file, not leaking the key into logs, no admin-flow break, and staging the work — this PR already *is* stage 1 of a planned series: **this PR** = node singleton secrets (nsec + admin password) at rest; **next** = encrypt the per-provider upstream API keys; **then** = env-vs-DB field ownership (the original "changing .env is silently ignored" bug); **then** = the admin UI that rotates secrets and rewires first-run, where a recovery-file export is a natural fit. The inline fixes from your review are all pushed now; proposing we land this once you've had a look.

9qeklajc:
nice work .. these I think worth doing now and we are good to go

﻿ Documentation is inconsistent with the implemented onboarding:

 - README.md:65 still tells new operators to set NSEC in .env, while .env.example:13-18 says new installations should
   configure it through the admin UI.
 - docs/provider/deployment.md:41 documents an unused ADMIN_KEY environment variable.
 - Other provider documentation still presents ADMIN_PASSWORD/NSEC environment bootstrap as the normal fresh-install path.

9qeklajc:
> nice work .. these I think worth doing now and we are good to go
> 
> ﻿ Documentation is inconsistent with the implemented onboarding:
> 
>     * README.md:65 still tells new operators to set NSEC in .env, while .env.example:13-18 says new installations should
>       configure it through the admin UI.
> 
>     * docs/provider/deployment.md:41 documents an unused ADMIN_KEY environment variable.
> 
>     * Other provider documentation still presents ADMIN_PASSWORD/NSEC environment bootstrap as the normal fresh-install path.

we could merge, pls rebase and update the docs as stated above

jeroenubbink:
everything rebased, updated docs, fixed one additional leaking test.

migrations/versions/c6f8d2e4a1b3_add_secrets_table.py
```diff
@@ -0,0 +1,51 @@
+"""add secrets table
+
+Revision ID: c6f8d2e4a1b3
+Revises: d7e8f9a0b1c2
+Create Date: 2026-07-07 00:00:00.000000
```
9qeklajc: this should be recreated, same for https://github.com/Routstr/routstr-core/pull/588 and https://github.com/Routstr/routstr-core/pull/623

jeroenubbink: Done — recreated. The migration had been amended in place across the review rounds (notably the `nsec_state` column), so its revision id no longer mapped to a single schema step. Deleted the old `c6f8d2e4a1b3` and re-minted it via `alembic revision` as a fresh `fc4fa29630d2` chained onto the current head `d7e8f9a0b1c2`, with an identical `create_table("secrets")` body. `alembic heads` now resolves to a single head and `upgrade head` runs clean on a fresh DB (all five columns incl. `nsec_state`); nothing in the code or tests referenced the old id. Fixed in https://github.com/Routstr/routstr-core/pull/570/commits/88fe975.

routstr/core/settings.py
```diff
@@ -291,20 +337,37 @@ async def initialize(cls, db_session: AsyncSession) -> Settings:
...
-            if db_json_raw != merged_dict:
+            # Keep npub consistent with the live nsec. bootstrap_secrets has
+            # already run and holds the single authoritative nsec (decrypted from
+            # the encrypted store, or freshly imported). merged_dict starts from
+            # the env/blob, which may carry a STALE nsec — and therefore a stale
+            # derived npub — after the vault took ownership. Derive from the live
+            # value and OVERRIDE, not just fill: otherwise the node keeps the
+            # vault's private key but announces the old env key's npub (npub is a
+            # pure derivation of nsec, never configured independently of it).
+            if settings.nsec:
```
9qeklajc, line 348: This leaves a stale `npub` behind when the vault’s authoritative state is `cleared`. I reproduced it with a full `bootstrap_secrets()` → `SettingsService.initialize()` test: bootstrap clears both fields, then initialize rebuilds `merged_dict` from the stale env `NSEC` and reapplies its derived `npub` because this branch only handles a non-empty live nsec. Could we explicitly set `merged_dict["npub"] = ""` when the authoritative live nsec is empty/cleared? A regression test extending `test_cleared_nsec_stays_cleared_across_reboot` through `initialize()` would catch it.

routstr/core/admin.py
```diff
@@ -241,53 +244,70 @@ async def update_settings(request: Request, update: SettingsUpdate) -> dict:
...
+        if not derived:
+            raise HTTPException(status_code=400, detail="Invalid nsec")
+        npub = derived
+
     async with create_session() as session:
-        await SettingsService.update({"admin_password": pw}, session)
-    return {"ok": True}
+        await set_nsec(session, nsec)
+
+    # Reflect the change in the live runtime so Nostr signing/announcements pick
```
9qeklajc, line 306: Updating the singleton alone doesn’t make announcements pick up the new key without a restart. If startup had no nsec, `main.py` never creates `nip91_task`; if it did, `announce_provider()` captured the old nsec/keypair before sleeping in its loop. Could the endpoint reconcile the announcement task — start it when adding an identity and cancel/restart it when rotating or clearing one — or otherwise make the task read the current identity each cycle?

9qeklajc (commented):
Thanks for putting the vault migration together — the full backend, UI, docs, typing, and migration checks all passed locally. I did find two runtime edge cases tied directly to the new nsec ownership/update flow. What do you think: should we close these in this PR, or would you prefer a small follow-up PR? Either approach is fine with me; I mainly want to make sure they’re tracked explicitly.

9qeklajc:
follow up PRs for open points

Issue #209, opened 2025-10-27
Nostr: Announcement management page (relays, edit/delete, publish)


sh1ftred:
Goal: Build a dedicated page and settings for managing the Nostr announcement.

Scope:
- Management UI for announcement content with edit/delete
- Relay management: add/remove, enable/disable, and save
- Publish if not yet published; re-publish/update when changed
- Basic permission checks and audit logging
- Wire UI to routstr-core endpoints; ensure chat/frontend can surface status

Acceptance Criteria:
- Page renders with current announcement content and relay list
- User can add/remove relays and save
- User can edit announcement content and save
- User can delete announcement
- User can publish if not yet published and see published status
- All actions validate and return clear errors

Notes:
- Reuse existing Nostr integration utilities
- Consider idempotent publish/update behavior
- Add docs in README or /docs for operators

Issue #208, opened 2025-10-27
Onboarding: Node setup flow (initial configuration)


sh1ftred:
Goal: Implement an onboarding flow to set up a Routstr node.

Scope:
- Provide a step-by-step onboarding wizard for initial node configuration (e.g., network, keys/identity, storage paths, minimal Nostr bootstrap where applicable)
- Validate inputs with clear error states; include permission checks and audit logging for critical actions
- Persist configuration with idempotent operations where possible
- Wire UI to routstr-core endpoints; ensure chat/frontend can surface onboarding status and completion

Acceptance Criteria:
- User can complete the onboarding flow end-to-end and see a clear success state
- Invalid inputs show actionable errors; partial progress can be resumed
- Configuration persists and can be re-run safely (idempotent where possible)
- Docs added/updated in README or /docs for operators

Notes:
- Reuse existing utilities and patterns from routstr-core
- Coordinate with chat/frontend for status surfacing

Issue #206, opened 2025-10-27
Onboarding: Nostr announcement management page (relays, edit/delete, publish)


sh1ftred:
Goal: Add onboarding and a dedicated section/page for managing the Nostr announcement so it can be manually controlled.

Scope:
- Create an onboarding flow and a settings/management page for the Nostr announcement
- Allow manual management of:
  - Relays (add/remove, enable/disable)
  - Edit/delete existing announcement content
  - Publish if not yet published; re-publish/update when changed
- Include basic permission checks and audit logging for changes
- Wire UI to backend endpoints in routstr-core, and ensure chat/frontend can surface status

Acceptance Criteria:
- Page renders with current announcement content and relay list
- User can add/remove relays and save
- User can edit announcement content and save
- User can delete announcement
- User can publish if not yet published and see published status
- All actions validate and return clear errors

Notes:
- Reuse any existing Nostr integration utilities
- Consider idempotent publish/update behavior
- Add docs in README or /docs for operators

sh1ftred:
Split note: This issue has been split into two focused issues for better tracking and execution:

1. Onboarding: Node setup flow (initial configuration) — https://github.com/Routstr/routstr-core/issues/208
2. Nostr: Announcement management page (relays, edit/delete, publish) — https://github.com/Routstr/routstr-core/issues/209

Both have been added to Project #1 and set to Status: Todo.

PR #205, opened 2025-10-24
V0.2.0 final

9qeklajc:

PR #571, opened 2026-06-27
fix(deps): patch reachable pip security advisories (aiohttp, pyjwt, python-multipart, pydantic-settings)

9qeklajc:
## Overview
Bumps the **pip dependencies in `uv.lock` that can be patched today** to clear their Dependabot security advisories. Only `uv.lock` changes — no source/runtime behavior change.

| Package | From | To | Severity | Alerts |
|---|---|---|---|---|
| aiohttp | 3.12.15 | 3.14.1 | medium/low ×9 | #192–#200 |
| pyjwt | 2.10.1 | 2.13.0 | high ×1, medium/low | #183–#187 |
| python-multipart | 0.0.20 | 0.0.32 | high ×1, low ×3 | #188–#191 |
| pydantic-settings | 2.14.0 | 2.14.2 | medium | #210 |

All transitive; resolved via `uv lock --upgrade-package`.

## ⚠️ Blocked advisories (NOT in this PR — need upstream `cashu` / `fastapi`)
These cannot be bumped without breaking the dependency graph; `cashu==0.20.0` (and 0.20.1) is the bottleneck:

- **litellm → 1.84.0 (CRITICAL, #209, auth bypass via Host header)** — `cashu>=0.20` is incompatible with `litellm>=1.84.0`.
- **cryptography → 48.0.1 (high, #204)** — `cashu` pins `cryptography>=43.0.1,<44.0.0`.
- **starlette → 1.3.1 (high, #206–#208 + #205)** — `cashu` is incompatible with `fastapi>0.115.14` (starlette 1.x needs newer fastapi).

Unblocking requires a `cashu` release that relaxes its `cryptography`/`fastapi` pins (or dropping `cashu`, which is core). Tracked separately — see PR discussion. The litellm CRITICAL should be prioritized for a mitigation/upstream bump.

## Verification
- `uv sync --frozen` ✅
- `uv run ruff check .` ✅ All checks passed
- `uv run mypy .` ✅ no issues in 174 files
- `uv run pytest tests/unit` ✅ 413 passed


jeroenubbink:
LGTM — clean lock-only bump, scope and the cashu-blocked list check out.

One note on the deferred litellm CRITICAL (CVE-2026-49468, Host-header auth bypass): that flaw is in litellm's **proxy/gateway** auth layer. routstr-core only depends on `litellm` as a **library** (no `[proxy]` extra, no `litellm.proxy` import anywhere) — cost maps, token counting, and the messages-dispatch adapters. So the vulnerable path isn't reachable here; still worth tracking the cashu unblock, but it's low-priority for this codebase rather than urgent.

PR #575, opened 2026-06-29
EHBP proxy support, Tinfoil direct integration, rate limiting, wallet fixes and more

sh1ftred:
## Summary

EHBP (Encrypted HTTP Body Protocol) proxy support for privacy-preserving AI inference, Tinfoil direct blind-upstream integration for TEE attestation, rate-limit-aware upstream failover, plus wallet fixes, dead-key pruning, and code quality improvements.

---

### EHBP — Encrypted HTTP Body Protocol

New module `routstr/upstream/ehbp.py` (911 lines) implementing a protocol where the proxy forwards an HPKE-sealed binary body without being able to read it. The client SDK sends an `Ehbp-Encapsulated-Key` header and the model ID via `X-Routstr-Model`.

- **ehbp.py** — full forwarding logic, HPKE key management, streaming with HTTP trailers for usage metrics
- **Proxy routing** — EHBP requests bypass body parsing and are routed directly to the upstream's private endpoint
- **Auth flow** — works with both x-cashu tokens and bearer key auth
- **Per-request cost headers** returned on EHBP/Tinfoil responses
- **Streaming** — buffered EHBP body streaming with trailer capture for accurate token usage
- **Provider opt-in** — explicit `supports_ehbp` flag on providers, EHBP forwarding target validation

Key files: `routstr/upstream/ehbp.py`, `routstr/proxy.py`, `routstr/auth.py`, `routstr/upstream/base.py`
Docs: `docs/ehbp-proxy-support.md`

### Tinfoil Direct Blind-Upstream Integration

Direct Tinfoil attestation support without requiring a separate Tinfoil sidecar.

- `routstr/upstream/tinfoil.py` (220 lines) — Tinfoil upstream provider with attestation-bundle proxying
- `routstr/upstream/tinfoil_trailer.py` (140 lines) — HTTP trailer parsing for TEE attestation headers
- Attestation GET requests (`tee/`, `attestation`) routed directly to Tinfoil providers only (avoids unrelated upstream 404s short-circuiting)
- EHBP enclave URL override validation
- Registered in `routstr/upstream/__init__.py` alongside other providers
- `TINFOIL_API_KEY` env var added to `.env.example` and `helpers.py` provider auto-seeding
- Docs: `docs/tinfoil-direct-integration.md`

### Rate Limiting & Upstream Failover

- `routstr/upstream/rate_limit.py` (136 lines) — proper rate-limit detection, classification, and forwarding (429s now propagate correctly instead of being swallowed)
- Upstream failover retries on 502, 429, 400, 401, 403, 404
- `on_upstream_error_redirect` callback on `BaseUpstreamProvider` for provider-specific cleanup on failover
- `UpstreamError` now carries structured `code` (e.g. `UPSTREAM_RATE_LIMIT`) and `details` for machine-readable error dispatch

---

### Wallet & Payments

- **Swap fee retry** — retry foreign-mint swaps with observed fees instead of trusting `fee_reserve` (fixes swap failures when reserve estimate is too low)
- **Stale reservation fix** — always charge balance on reservation overrun, even after a stale sweep
- **Melt error handling** — require shortfall text for generic 11000 melt errors

### DeepSeek V4 Pricing Backfill

- `routstr/upstream/deepseek_v4_pricing_shim.py` (66 lines) — temporary backfill for DeepSeek V4 models missing from litellm's cost map (BerriAI/litellm#30430). Registered at startup in `main.py`. Corrects cache pricing calculations.

### OpenRouter Tool-Use Routing

- OpenRouter provider now sets `provider.require_parameters` on requests with `tools`, preventing routing to endpoints that don't support function calling (which would 404 with "No endpoints found that support tool use")

### Dead API Key Pruning

- Background janitor task (`periodic_dead_key_prune`) that prunes zero-balance, never-used API keys
- Configurable via `DEAD_KEY_PRUNE_INTERVAL_SECONDS` (default 3600) and `DEAD_KEY_MIN_AGE_SECONDS` (default 1 week)
- Integration tests in `tests/integration/test_prune_dead_api_keys.py`

### Centralized Redaction

- `routstr/core/redaction.py` (51 lines) — centralized org-ID redaction used across logging (`routstr/core/logging.py`) and upstream error messages (`routstr/upstream/messages_dispatch.py`)

### Code Quality

- **Mypy fixes** — resolved all mypy errors (Field defaults, missing return types, ASGITransport arg-type)
- **Ruff fixes** — removed unused imports, added TYPE_CHECKING import
- **pyproject.toml** — removed stale `secp256k1` git source pin

### UI Documentation

- `ui/README.md` — documented dev/build/serve workflow, made missing-bundle warning actionable

### Misc

- `.gitignore` — added `.worktrees`

---

## Files Changed (38 files, +5370 / -304)

| File | Description |
|------|-------------|
| `routstr/upstream/ehbp.py` | EHBP forwarding, encryption, streaming (new) |
| `routstr/upstream/tinfoil.py` | Tinfoil direct upstream provider (new) |
| `routstr/upstream/tinfoil_trailer.py` | TEE attestation trailer parsing (new) |
| `routstr/upstream/rate_limit.py` | Rate limit detection & forwarding (new) |
| `routstr/core/redaction.py` | Centralized org-ID redaction (new) |
| `routstr/upstream/deepseek_v4_pricing_shim.py` | DeepSeek V4 cost-map backfill (new) |
| `routstr/proxy.py` | EHBP routing, Tinfoil attestation routing, upstream failover |
| `routstr/upstream/base.py` | EHBP hooks, `on_upstream_error_redirect`, `supports_ehbp` |
| `routstr/auth.py` | EHBP auth, dead-key prune janitor |
| `routstr/wallet.py` | Swap fee retry, melt error handling |
| `routstr/core/exceptions.py` | UpstreamError gains `code` + `details` |
| `routstr/upstream/openrouter.py` | Tool-use routing fix |
| `routstr/upstream/messages_dispatch.py` | Redaction + rate limit propagation |
| `routstr/payment/helpers.py` | `create_upstream_error_response` helper |
| `routstr/payment/cost_calculation.py` | Discounted max-cost support |
| `routstr/core/db.py` | Prune dead keys query |
| `routstr/core/settings.py` | Dead-key prune interval settings |
| `routstr/core/logging.py` | Redaction integration in log records |
| `routstr/core/main.py` | Startup hooks (prune janitor, DeepSeek shim) |
| `routstr/upstream/__init__.py` | Tinfoil provider registration |
| `routstr/upstream/helpers.py` | TINFOIL_API_KEY env mapping |
| `.env.example` | Tinfoil API key template |
| `.gitignore` | Add `.worktrees` |
| `docs/` (2 new files) | EHBP + Tinfoil documentation |
| `tests/` (7 new files) | Integration + unit tests |

---

## Testing

- **Unit tests**: EHBP streaming (`test_streaming_sse_providers.py`), Tinfoil attestation routing (`test_proxy_tinfoil_attestation_routing.py`), Tinfoil integration (`test_tinfoil_integration.py`), rate limiting (`test_upstream_rate_limit.py`), require_parameters (`test_require_parameters.py`), cache pricing (`test_cache_pricing.py`)
- **Integration tests**: dead API key pruning (`test_prune_dead_api_keys.py`), swap fee retry (`test_swap_fee_retry.py`), stale reservation charges (`test_free_response_stale_reservation.py`)
- Manual testing with EHBP SDK client and Tinfoil attestation


9qeklajc:
some thought from me and my agent 

1. Blind EHBP model/billing spoofing risk
     - PR trusts client-supplied X-Routstr-Model because the body is encrypted.
     - Tinfoil path does not appear to bind usage metrics to a provider-attested model identity.
     - A client may declare a cheap model while sending an encrypted request for a more expensive one unless
       Tinfoil/provider enforces or signs model identity.

 2. Architecture is too Tinfoil/EHBP-specific for broader TEE future
     - ehbp.py hard-codes Tinfoil headers, usage parsing, enclave URL policy, and host allowlist.
     - This is fine for a Tinfoil-specific PR, but not a scalable abstraction for Redpill, Phala, NEAR AI, Nillion,
       Privatemode, Azure Confidential Inferencing, etc.

 Fixes worth doing now

 - /.well-known/* docs claim unauthenticated forwarding, but implementation only bypasses auth/model lookup for tee/ and
   attestation; /.well-known/* currently falls through to invalid_model.
 - Bearer EHBP billing should share the clamped _compute_ehbp_actual_cost() path used by X-Cashu.
 - forward_with_trailer() should add timeouts and response-size limits.
 - Add direct h11 dependency if importing it directly.
 - PPQ is marked supports_ehbp, but shared usage extraction only understands Tinfoil usage headers; likely max-cost billing
   fallback.

 Alternative I recommend

 Introduce a provider-neutral ConfidentialInferenceProfile:

 - protocol: EHBP / provider-side TEE / sidecar / none
 - attestation routes + verifier
 - target resolver + override policy
 - trusted model-binding strategy
 - provider-specific usage extractors
 - missing-usage billing policy
 - strict header forwarding policy

sh1ftred:
Okay, #1 is a valid concern, I didn't think about that.

sh1ftred:
I've created an issue in the Tinfoil proxy repository. For now, a temporary fix is to only support 3 models, GLM-5.2, Deepseek V4 Pro and Kimi k2.6 as they all have the same pricing. We don't support any other models for now.

sh1ftred:
ok, damn. that doesn't work either. We need this information from Tinfoil and we don't have any other choice :(

sh1ftred:
Okay, for now, we'll just take the risk and hope nobody tries to scam my node here.

jeroenubbink (commented):
Seen @9qeklajc's review — +1 rather than repeating, especially the `forward_with_trailer` timeouts/size-limits, the `.well-known/*` fall-through, and the "too Tinfoil-specific to generalize across TEE providers" point (the `ConfidentialInferenceProfile` direction sounds right). A few things I didn't see covered (one inline on `tinfoil_trailer.py`):

**PR description vs. diff** — The description covers a lot that's already in `main` (rate limiting, wallet fixes, dead-key pruning, the DeepSeek shim, redaction…), pulled in via the `main` merges; the actual net change is the EHBP + Tinfoil addition (~16 files). Could you regenerate it from `git diff main...HEAD`? This repo merges without squashing so the body is permanent history, and the current "Testing" section lists tests already in `main` rather than the ones this PR adds.

**Test coverage** — The pure helpers are well tested (usage parsing, SSRF rejection, cost computation), but the functions that move money or do the networking aren't: `forward_ehbp_request`, `forward_ehbp_x_cashu_request` (redeem/refund math + error cascade), `finalize_ehbp_max_cost_payment` (the balance UPDATE), and `forward_with_trailer` itself — including the trailer-capture path, which is the whole reason that module exists. Combined with the max-cost fallback, a silent regression in trailer capture would over-charge streaming requests to max-cost behind a warning log with nothing failing in CI. I'd prioritize the x-cashu refund path and the trailer client.

**Billing / cache (heads-up, not a current bug)** — The usage header only carries `prompt`/`completion` and `TinfoilModelPricing` has no cache tier, so billing is correct today. But if Tinfoil ever reports cached tokens (enclave stacks commonly do prefix caching), they stay folded into `prompt` and bill at the full input rate — the over-charge-on-cache-reads pattern that was fixed for other upstreams. Worth a parser field + pricing tier (or at least an assert/log) so it can't regress silently.

**Minor** — `finalize_ehbp_max_cost_payment`: the child-key UPDATE's `rowcount` isn't checked, and the parent check runs after `commit()`, so it can only log.

pyproject.toml
```diff
@@ -86,3 +86,6 @@ disallow_untyped_decorators = true
 
 [tool.uv.sources]
 routstr = { workspace = true }
+
+[tool.uv]
+exclude-newer = "2 weeks"
```
jeroenubbink: This `[tool.uv]` block looks like it slipped in from a local environment rather than being intentional — `exclude-newer` expects an RFC 3339 date/timestamp (e.g. `2026-06-15`), not a relative string. I tested it on uv 0.11.16: it doesn't error (so CI stays green) but it's silently ignored — it pins nothing and does nothing. Worth dropping these two lines before merge so a stricter/newer uv doesn't later start rejecting or misinterpreting it.

routstr/upstream/tinfoil_trailer.py
```diff
@@ -0,0 +1,140 @@
...
+        return TrailerResponse(
+            status_code=status_code,
+            headers=resp_headers,
+            body=b"".join(body_chunks),
+            trailers=trailers,
+        )
+    finally:
+        try:
+            writer.close()
+            await writer.wait_closed()
```
jeroenubbink: Beyond adding a timeout (per @9qeklajc): this `await writer.wait_closed()` runs in the `finally` before the already-built `TrailerResponse` is returned, so even a fully successful request waits on the TLS shutdown — a slow shutdown adds latency to every call and a hung one turns a completed request into a client-visible hang. With `Connection: close` and the body already fully read, a fire-and-forget `writer.close()` is enough here, or bound the wait with a timeout.

pyproject.toml
sh1ftred: What's the best way to handle this? We want to block packages newer than 2 weeks.

jeroenubbink: Hey @sh1ftred — I think the `exclude-newer` line was reaching for something real, just worth separating two threat models, because a cooldown only addresses one:

- **Stability / "don't run buggy code"** — this is what your example points at ("critical bug patched 2 days later"). A cooldown doesn't help here: a time window delays the fix as much as the bug, so you're never better off. You're right that it's a weird benchmark for this.
- **Supply-chain / "don't install a compromised release"** — the xz / PyPI-account-takeover pattern, where a legit package name suddenly ships a malicious version. Those get caught and yanked within hours-to-days, so a cooldown means you're never the one installing it while it's live. This is the only thing `exclude-newer` is designed for.

If the goal is supply-chain hardening, the instinct is right — but a blanket cutoff is the blunt version: it delays legit security patches by the same window, with no escape hatch, and it's largely redundant with `uv.lock` (which already pins exact versions + hashes).

Two cleaner paths, curious which fits your intent:

1. **Keep chasing latest, but with a real cooldown → Dependabot.** Native uv/`uv.lock` support, and its `cooldown` option went GA 2026-07-01. Cooldown applies to *version* updates only — **security updates bypass it**, so a CVE fix still lands immediately — and every bump is a reviewable, CI-gated PR instead of resolving silently.
2. **Don't chase latest → pin to versions that have proven to work and let semver patch ranges handle patching.** Not everything here is fully test-covered, so a major bump relies on manual verification; pytest can't confirm compatibility to 100%. Holding known-good pins and moving majors deliberately is safer than any time-window heuristic.

What's the actual driver here — supply-chain worry, or staying on latest?

sh1ftred:
Thanks for the reviews and updates! 
Re: trailer timeouts, yes, valid concern and the fix. 
Re: generalizing to ConfidentialInferenceProfile, we don't know how other E2EE TEE upstreams work to generalize this already. Was not an immediate requirement, could be a refactor for later. 
Re: .well-known, i had it before and removed it later as I felt that it wasn't really needed, but shouldn't be a problem i guess. 

So, the problem was that I removed it from the pass through links but my clanker didn't update the docs and the other place where it was added. 

> The .well-known/ entry in _API_PATH_PREFIXES and the doc line claiming /.well-known/ paths are forwarded without auth appear to be a mistake — it was likely added
 speculatively (or copy-pasted alongside attestation) during development, but Tinfoil's own docs never ask for it. The tee/ prefix also isn't in Tinfoil's docs — that
 looks like an internal naming choice for the attestation route (tee/attestation).

 > The review comment from 9qeklajc/Jeroen about .well-known/* fall-through is technically correct (the code is broken for that path), but the deeper point is:
 .well-known/ shouldn't be there at all. It should just be removed from both _API_PATH_PREFIXES and the docs, since Tinfoil doesn't use it.

We should remove it altogether.

jeroenubbink (approved):
Re-reviewed the advanced branch — approving.

The money path is now covered where it counts: `test_ehbp_finalize_payment.py` exercises both `finalize_ehbp_actual_cost_payment` and `finalize_ehbp_max_cost_payment` against a real SQLite session — happy path (balance debit, reserve release, parent+child spend) **and** both zero-rowcount branches, confirming the rollback-before-commit discipline actually fires on a concurrently-deleted key. That was the main gap from the earlier pass. Timeouts + size cap on the trailer client, the rowcount guards, PPQ opt-out, and the explicit `h11` pin are all in; dropping `.well-known/` entirely is a clean resolution.

Non-blocking follow-ups (fine to defer — noting so they aren't lost, not asking for changes here):
- The trailer-based client buffers the whole response then re-emits one chunk, so streaming EHBP has full time-to-last-byte latency. In-band usage (a normal header non-streaming / final SSE `usage` event streaming) would let this whole custom client go — the better long-term shape once a second confidential upstream forces the question.
- The Tinfoil usage header carries only prompt/completion; the day the enclave reports cached tokens they'll fold into the prompt total and bill at full input rate. A parser field + a cache pricing tier (or an assert-and-log on an ignored cache count) closes that before it can mischarge.
- `X-Routstr-Cost-Usd` is emitted as `0.0` on the bearer usage path — informational header only, billing (msats) is correct, but worth populating when convenient.
- Dispatch stays as central `is_ehbp` branches + a hardcoded tinfoil check; agreed that generalizing is premature, but it's the thing to lift into the provider seam before a third confidential mode lands.

Two housekeeping notes before merge: the branch is ~16 behind main (worth a rebase so CI runs against current main), and the `exclude-newer` thread is still open — happy with either direction you land on there.

sh1ftred:
Re: follow ups: 
1. Can you clarify what you mean here? 
2. Caching isn't supported on Tinfoil rn, they're working on it, we'll work on this once that is live. 
3. Cool
4. Yep, we'll handle that later :)

Yes, will rebase. Re: exclude-newer, we can remove it for now. We should think more about supply chain attacks tho.

routstr/upstream/tinfoil_trailer.py
```diff
@@ -0,0 +1,161 @@
...
+
+    try:
+        # Build HTTP/1.1 request
+        header_lines = [f"{method} {path} HTTP/1.1"]
+        has_host = any(k.lower() == "host" for k in headers)
+        if not has_host:
+            header_lines.append(f"Host: {host}")
+        header_lines.append("Connection: close")
+
+        for key, value in headers.items():
```
9qeklajc, line 116: Could we strip hop-by-hop headers here before forwarding the request? Client values such as `Transfer-Encoding`, `TE`, `Trailer`, `Upgrade`, and headers named by `Connection` can currently be sent alongside the forwarder’s own framing. Since FastAPI has already decoded the incoming body, that can produce a malformed upstream request.

A regression test against the serialized h11 request would be helpful too.

routstr/proxy.py
```diff
@@ -185,20 +210,54 @@ async def proxy(
...
+        if is_responses_api:
+            model_id = extract_model_from_responses_request(request_body_dict)
+        else:
+            model_id = request_body_dict.get("model", "unknown")
+
+    # /tee/* and /attestation GET requests don't map to models — forward
+    # without model/cost/auth lookups. Tinfoil attestation paths are routed
+    # only to Tinfoil providers so an unrelated upstream's 404 cannot
+    # short-circuit before the attestation proxy is tried.
+    if request.method == "GET" and (
```
9qeklajc: Could we exact-match the supported attestation routes instead of using `startswith("attestation")`? The current check also admits paths like `/attestation/foo` and `/attestationjunk`, which may then be forwarded without authentication.

Please also cover a trailing slash and a non-attestation `attestation*` path in the tests.

docs/ehbp-proxy-support.md
```diff
@@ -0,0 +1,158 @@
...
+- `forward_ehbp_request()` — forwards the raw encrypted body to an EHBP-capable
+  provider, streams the encrypted response back untouched, and finalizes bearer
+  billing at max cost because usage is encrypted
+- `forward_ehbp_x_cashu_request()` — redeems the Cashu token, forwards raw,
+  refunds the full token on upstream failure, and refunds any value above
+  `max_cost_for_model` on success
+
+### `routstr/upstream/ppqai.py`
+
+- Sets `supports_ehbp = True`.
```
9qeklajc: This section looks out of date: PPQ still has EHBP disabled, while Tinfoil responses can now use usage headers or trailers for actual-cost billing. Could we update this to describe the current Tinfoil-only behavior?

docs/tinfoil-direct-integration.md
```diff
@@ -0,0 +1,493 @@
...
+
+If the actual model is not found in Routstr's model registry, billing falls
+back to the requested model's pricing.
+
+### What still needs verification
+
+- ~~End-to-end test with a real Tinfoil SDK client against a Routstr node with
+  `TINFOIL_API_KEY` set.~~ Verified: both non-streaming (header) and streaming
+  (trailer) responses include `model=<name>`.
+- Streaming requests: usage is delivered as an HTTP trailer. Currently the
```
9qeklajc: This TODO is now obsolete because streaming trailer capture and EHBP payment finalization are implemented. The reference to `adjust_payment_for_tokens()` also appears stale; the code now uses the dedicated EHBP finalizers.

9qeklajc (commented):
Nice work overall. 

- **Blocker:** strip hop-by-hop headers before raw EHBP forwarding.
- **Worth fixing:** only bypass authentication for exact attestation routes.
- **Worth fixing:** update the EHBP/Tinfoil docs to match the implementation.

The broader EHBP abstraction is reasonable for Tinfoil today, but it still contains several Tinfoil-specific assumptions. I’d either describe it as Tinfoil-first for now or move usage parsing, target validation, and attestation routing behind provider hooks in a follow-up.

9qeklajc:
we can do a followup for refactoring to adopt more TEE providers

PR #576, opened 2026-06-30
Refactor pr575 review

9qeklajc:


routstr/upstream/ehbp.py
```diff
@@ -0,0 +1,1083 @@
...
+@dataclass(frozen=True)
+class EHBPForwardingTarget:
+    """Provider-specific destination for an EHBP opaque request."""
+
+    url: str
+    headers: Mapping[str, str] = field(default_factory=dict)
+    profile: ConfidentialInferenceProfile | None = None
+
+
+async def finalize_ehbp_actual_cost_payment(
```
jeroenubbink, line 391: The new `finalize_ehbp_actual_cost_payment` isn't exercised at the DB level anywhere — the clamping in `_compute_ehbp_actual_cost` is well covered, but this balance-mutating function isn't. Since it debits `balance`, releases the reserve, and now rolls back on a zero-rowcount race, it's exactly the path worth a test: assert the `balance` / `reserved_balance` transitions on the happy path, and that a 0-row parent-or-child update rolls back with no partial charge. The same gap applies to `finalize_ehbp_max_cost_payment`. Non-blocking — the networking half of the coverage is already handled by the new trailer test.

routstr/upstream/ehbp.py
```diff
@@ -0,0 +1,1083 @@
...
+        if k.lower() == usage_header_name_lower:
+            return v
+    for k, v in trailers:
+        if k.lower() == usage_header_name_lower:
+            return v
+    return None
+
+
+@dataclass(frozen=True)
+class ConfidentialInferenceProfile:
```
jeroenubbink, line 373: `ConfidentialInferenceProfile` is only partly consumed: `usage_response_header`, `client_target_url_header`, `allow_client_target_override` and `proxy_only_headers` are all threaded through, but `protocol`, `trusted_model_binding_header` and `missing_usage_billing_policy` are set here and never read. `missing_usage_billing_policy="max_cost"` happens to match the hardcoded fallback so there's no bug, but nothing drives behaviour off it. The one I'd single out is `trusted_model_binding_header` — it reads like it addresses the client-supplied-model spoofing concern, but since it's never checked, its presence doesn't mitigate anything, and a later reader could assume model-binding is handled when it isn't. I'd either wire them (drive the missing-usage fallback off the field, enforce the binding header when set) or drop them until the code that consumes them lands. Non-blocking.

jeroenubbink (commented):
Thanks for turning this around so quickly — this addresses essentially the whole review. The `asyncio.wait_for` timeouts + 25 MB cap on the trailer client, the rowcount-checked-before-commit-with-rollback on both finalize paths, disabling `supports_ehbp` on PPQ so it can't silently fall back to max-cost, and pulling the Tinfoil specifics behind `ConfidentialInferenceProfile` all look right. `test_tinfoil_trailer.py` covering the h11 client (trailer capture + size limit) closes the gap that worried me most.

Left two small non-blocking notes inline. One more minor, related to the new profile: `_validated_confidential_target_url` only knows the Tinfoil validator, so a future profile that opts into `allow_client_target_override` with a different header returns `None` and always rejects. That fails closed (the safe direction), but a one-line comment noting the override path is still Tinfoil-only in practice would save the next person some confusion.

sh1ftred:
Thanks for the reviews and updates! 
Re: trailer timeouts, yes, valid concern and the fix. 
Re: generalizing to ConfidentialInferenceProfile, we don't know how other E2EE TEE upstreams work to generalize this already. Was not an immediate requirement, could be a refactor for later. 
Re: .well-known, i had it before and removed it later as I felt that it wasn't really needed, but shouldn't be a problem i guess. 

So, the problem was that I removed it from the pass through links but my clanker didn't update the docs and the other place where it was added. 

> The .well-known/ entry in _API_PATH_PREFIXES and the doc line claiming /.well-known/ paths are forwarded without auth appear to be a mistake — it was likely added
 speculatively (or copy-pasted alongside attestation) during development, but Tinfoil's own docs never ask for it. The tee/ prefix also isn't in Tinfoil's docs — that
 looks like an internal naming choice for the attestation route (tee/attestation).

 > The review comment from 9qeklajc/Jeroen about .well-known/* fall-through is technically correct (the code is broken for that path), but the deeper point is:
 .well-known/ shouldn't be there at all. It should just be removed from both _API_PATH_PREFIXES and the docs, since Tinfoil doesn't use it.

We should remove it altogether.

PR #577, opened 2026-07-01
refactor(upstream): resolve a provider's own row by primary key

jeroenubbink:
A live upstream provider re-finds its own database row by `(base_url, api_key)` in two places — PPQ.AI's insufficient-balance self-disable and the base `refresh_models_cache`. That uses a rotatable secret as a self-handle: if the row's `api_key` changes underneath a live object (a key rotation racing an in-flight request), the object can no longer find itself.

This resolves a provider's own row by its **primary key** instead:

- `from_db_row` becomes a template method that stamps the row's id onto the in-memory object (`db_id`) and delegates construction to a `_build_from_row` hook — the provider subclasses override the hook, not the wrapper. `db_id` was previously set only via a fragile `setattr` in one init path, so objects built through `from_db_row` (auto-topup, tests) never learned their identity.
- The two self-referential lookups now use `session.get(UpstreamProviderRow, self.db_id)`.
- `auto_topup` gains a `None` guard for the `from_db_row` return.

Pure internal refactor — no schema change, no migration, no API change.

Covered by `tests/integration/test_provider_self_lookup.py`: the object carries its persistent identity, and both the self-disable and `refresh_models_cache` survive a key rotation on a live object.

Context: prep for encrypting `upstream_providers.api_key` (part of #553) — getting the secret out of identity and lookup positions first.

### DIFF
--- a/routstr/upstream/anthropic.py
+++ b/routstr/upstream/anthropic.py
@@ -24,7 +24,7 @@ class AnthropicUpstreamProvider(BaseUpstreamProvider):
         )
 
     @classmethod
-    def from_db_row(
+    def _build_from_row(
         cls, provider_row: "UpstreamProviderRow"
     ) -> "AnthropicUpstreamProvider":
         return cls(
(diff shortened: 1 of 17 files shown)


9qeklajc:
LGTM

PR #578, opened 2026-07-01
fix: specific, sanitized errors for failed Cashu token redemption

9qeklajc:
## Summary

When a new Cashu token is used as a bearer key and redemption fails, the API previously returned a blanket `401` with `code: "invalid_api_key"` and echoed the raw exception text to the client. This PR sanitizes those failures and — following review — routes them through the **same failure taxonomy the X-Cashu redemption path already uses**, so both token-redemption endpoints agree on the classification instead of introducing a third parallel code.

## Behaviour

Bearer-path (`Authorization: Bearer cashu…`) redemption failures now reuse the shared taxonomy (the machine-readable class is carried in `type`, matching `create_error_response` on the X-Cashu path), with sanitized, stable messages:

| Failure | Status | `type` | Message |
|---|---|---|---|
| Token already spent | 400 | `token_already_spent` | `Cashu token already spent` |
| Token too small for swap fees | 422 | `mint_error` | `Token value is too small to cover swap fees` |
| Foreign mint melt failed | 422 | `mint_error` | `Failed to swap token from foreign mint` |
| Malformed / undecodable token | 400 | `invalid_token` | `Invalid Cashu token` |
| Other expected wallet error (`ValueError`) | 400 | `cashu_error` | `Failed to redeem Cashu token` |

Unexpected failures (DB/runtime errors) return `500` with `type: "api_error"`, `code: "internal_error"` and a generic message instead of a 4xx token error.

Example response body:

```json
{
  "detail": {
    "error": {
      "message": "Cashu token already spent",
      "type": "token_already_spent",
      "code": 400
    }
  }
}
```

Additional guarantees / fixes:

- Raw exception text (mint URLs, fee/amount details, nested errors) never reaches the client; full details stay in server logs.
- The broad `invalid`/`decode` buckets are anchored to co-occur with `token`, so internal faults whose text merely contains those words (e.g. `Invalid isoformat string`, SQLAlchemy `Invalid …`) fall through to `500` instead of masquerading as a token error.
- The X-Cashu path leak twin is fixed too: `mint_error`/`cashu_error` responses no longer interpolate the raw mint/exception string (both duplicated blocks in `upstream/base.py`).
- A failed redemption rolls back the session, so no orphan zero-balance `ApiKey` row is persisted.

## ⚠️ Client-observable changes

These are raised inside a dependency and are **not** documented in `openapi.json` today, but consumers branching on status/`code`/`type` will observe:

- Bearer redemption failures move from `401 invalid_api_key` to `400`/`422` with the taxonomy above.
- Malformed-token status changed `401 → 400` (`invalid_token`).
- Fee/melt failures now return `422` (`mint_error`) rather than `400`.

**SDK follow-up (`routstr-sdk`):** `RoutstrClient._handleErrorResponse` gates spent/invalid-key storage cleanup on `status === 401`. With these codes the SDK needs to trigger cleanup on the new taxonomy (`token_already_spent`/`invalid_token`) — paired update required so failed tokens still get evicted.

## Notes

- `/v1/wallet/topup` keeps its existing string-only `detail` errors; migrating it is a client-visible contract change left for a follow-up.

## Test plan

- [x] Parametrized unit tests in `tests/unit/test_auth_cashu.py` covering all failure classes: status code, shared `type`, exact sanitized message, no raw text leak, no orphan key row
- [x] Test asserting a non-wallet fault whose text contains `invalid` (but not `token`) falls through to `500` (anchored-gate guard)
- [x] Test asserting unexpected errors return `500 internal_error` without leaking details
- [x] Full suite: 679 passed, 13 skipped; ruff and mypy clean

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -81,34 +81,55 @@ async def check_and_reset_limit(key: ApiKey, session: AsyncSession) -> bool:
 def redemption_error_to_http_exception(error: Exception) -> HTTPException:
     """Map a Cashu token redemption failure to a sanitized client-facing error.
 
-    Known redemption failure patterns (from the wallet or the mint) and expected
-    wallet errors (ValueError) map to 400/401 with a stable message under a
-    single "token_redemption_failed" code. Anything else is an internal fault
-    and maps to a generic 500. Raw error text never reaches the client — it
-    stays in server logs.
+    Uses the same failure taxonomy as the X-Cashu redemption path
+    (``token_already_spent``/``invalid_token``/``mint_error``/``cashu_error``,
+    see ``create_error_response`` in ``payment/helpers.py``) so both paths that
+    redeem a token agree on the classification (carried in ``type``) and status.
+    Expected wallet errors (``ValueError``) and known mint failures map to a
+    4xx with a stable message; anything else is an internal fault and maps to a
+    generic 500. Raw error text never reaches the client — it stays in logs.
     """
     lowered = str(error).lower()
-    status_code = 400
+    # (type, status, message) — type mirrors the X-Cashu taxonomy strings.
     if "already spent" in lowered:
-        message = "Cashu token already spent"
+        error_type, status_code, message = (
+            "token_already_spent",
+            400,
+            "Cashu token already spent",
+        )
     elif (
         "insufficient" in lowered
         or "melt fee" in lowered
         or "exceed token amount" in lowered
         or "estimate fees" in lowered
     ):
-        message = "Token value is too small to cover swap fees"
+        error_type, status_code, message = (
+            "mint_error",
+            422,
+            "Token value is too small to cover swap fees",
+        )
     elif "failed to melt" in lowered:
-        message = "Failed to swap token from foreign mint"
+        error_type, status_code, message = (
+            "mint_error",
+            422,
+            "Failed to swap token from foreign mint",
+        )
     elif ("invalid" in lowered or "decode" in lowered) and "token" in lowered:
         # Anchor the broad buckets to "token" so internal faults whose text
         # merely contains "invalid"/"decode" (e.g. "invalid literal",
         # SQLAlchemy "Invalid …") fall through to the generic 500 below
-        # instead of masquerading as a 401 token error.
-        message = "Invalid Cashu token"
-        status_code = 401
+        # instead of masquerading as a token error.
+        error_type, status_code, message = (
+            "invalid_token",
+            400,
+            "Invalid Cashu token",
+        )
     elif isinstance(error, ValueError):
-        message = "Failed to redeem Cashu token"
+        error_type, status_code, message = (
+            "cashu_error",
+            400,
+            "Failed to redeem Cashu token",
+        )
     else:
         return HTTPException(
             status_code=500,
@@ -125,8 +146,8 @@ def redemption_error_to_http_exception(error: Exception) -> HTTPException:
         detail={
             "error": {
                 "message": message,
-                "type": "invalid_request_error",
-                "code": "token_redemption_failed",
+                "type": error_type,
+                "code": status_code,
             }
         },
     )
@@ -396,9 +417,12 @@ async def validate_bearer_key(
                     "Token redemption returned zero or negative amount",
                     extra={"msats": msats, "key_hash": hashed_key[:8] + "..."},
                 )
-                # Defense-in-depth: credit_balance now refuses to commit on a
-                # zero/negative redemption, but if a row was nonetheless
-                # persisted, drop it so we never leave an orphan zero-balance key.
+                # Defense-in-depth: credit_balance already raises
+                # ValueError("Redeemed token amount must be positive…") before
+                # returning (wallet.py), so this branch is only reachable if a
+                # zero/negative row was somehow persisted; drop it so we never
+                # leave an orphan zero-balance key. Reuse the shared taxonomy
+                # (cashu_error) so the envelope matches the mapper above.
                 await session.delete(new_key)
                 await session.commit()
                 raise HTTPException(
@@ -406,8 +430,8 @@ async def validate_bearer_key(
                     detail={
                         "error": {
                             "message": "Failed to redeem Cashu token: token yielded no value",
-                            "type": "invalid_request_error",
-                            "code": "token_redemption_failed",
+                            "type": "cashu_error",
+                            "code": 400,
                         }
                     },
                 )
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -78,6 +78,51 @@ async def check_and_reset_limit(key: ApiKey, session: AsyncSession) -> bool:
     return False
 
 
+def redemption_error_to_http_exception(error: Exception) -> HTTPException:
+    """Map a Cashu token redemption failure to a sanitized client-facing error.
+
+    Known redemption failure patterns (from the wallet or the mint) and expected
+    wallet errors (ValueError) map to 400/401 with a stable message under a
+    single "token_redemption_failed" code. Anything else is an internal fault
+    and maps to a generic 500. Raw error text never reaches the client — it
+    stays in server logs.
+    """
+    lowered = str(error).lower()
+    status_code = 400
+    if "already spent" in lowered:
+        message = "Cashu token already spent"
+    elif "insufficient" in lowered or "melt fee" in lowered:
+        message = "Token value is too small to cover swap fees"
+    elif "failed to melt" in lowered:
+        message = "Failed to swap token from foreign mint"
+    elif "invalid" in lowered or "decode" in lowered:
+        message = "Invalid Cashu token"
+        status_code = 401
+    elif isinstance(error, ValueError):
+        message = "Failed to redeem Cashu token"
+    else:
+        return HTTPException(
+            status_code=500,
+            detail={
+                "error": {
+                    "message": "Internal error during token redemption",
+                    "type": "api_error",
+                    "code": "internal_error",
+                }
+            },
+        )
+    return HTTPException(
+        status_code=status_code,
+        detail={
+            "error": {
+                "message": message,
+                "type": "invalid_request_error",
+                "code": "token_redemption_failed",
+            }
+        },
+    )
+
+
 async def validate_bearer_key(
     bearer_key: str,
     session: AsyncSession,
@@ -334,7 +379,8 @@ async def validate_bearer_key(
                         "error_type": type(credit_error).__name__,
                     },
                 )
-                raise credit_error
+                await session.rollback()
+                raise redemption_error_to_http_exception(credit_error) from credit_error
 
             if msats <= 0:
                 logger.error(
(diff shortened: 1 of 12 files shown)


sh1ftred:
What's the error you've added for mint not being reachable by the node? So that routstrd can send a token from a different mint?

9qeklajc:
> What's the error you've added for mint not being reachable by the node? So that routstrd can send a token from a different mint?

for instance routstrd can fallback to a second mint if the error:code is `token_redemption_failed` 

see above:
```json
{
  "detail": {
    "error": {
      "message": "Cashu token already spent",
      "type": "invalid_request_error",
      "code": "token_redemption_failed"
    }
  }
}
```

routstr/auth.py
```diff
@@ -78,6 +78,51 @@ async def check_and_reset_limit(key: ApiKey, session: AsyncSession) -> bool:
     return False
 
 
+def redemption_error_to_http_exception(error: Exception) -> HTTPException:
```
jeroenubbink, line 85: This duplicates the failure taxonomy we already have on the X-Cashu path (`upstream/base.py:4031-4064`, via `create_error_response` in `payment/helpers.py`): `token_already_spent`/`invalid_token`/`mint_error`/`cashu_error`. Same failure classes, different codes+statuses. Can we route through the shared helper and reuse those codes so both redemption paths agree? If the intent is to standardize on `token_redemption_failed`, let's update both paths in one go rather than leave them divergent.

routstr/auth.py
```diff
@@ -78,6 +78,51 @@ async def check_and_reset_limit(key: ApiKey, session: AsyncSession) -> bool:
...
+    """
+    lowered = str(error).lower()
+    status_code = 400
+    if "already spent" in lowered:
+        message = "Cashu token already spent"
+    elif "insufficient" in lowered or "melt fee" in lowered:
+        message = "Token value is too small to cover swap fees"
+    elif "failed to melt" in lowered:
+        message = "Failed to swap token from foreign mint"
+    elif "invalid" in lowered or "decode" in lowered:
```
jeroenubbink: These substring checks run on *any* exception before the `isinstance(ValueError)` gate, so the "internal faults no longer masquerade as token errors" guarantee leans entirely on internal messages never containing these words. `"invalid"` is broad — it shows up in plenty of stdlib/driver text (`invalid literal`, `invalid isoformat string`, SQLAlchemy `Invalid …`), any of which would surface as `401 Invalid Cashu token` instead of `500`. A full `isinstance` gate isn't viable (mint "already spent" isn't a `ValueError`), but anchoring the broad buckets — e.g. requiring `"invalid"`/`"decode"` to co-occur with `"token"` — would tighten it. The `test_unexpected_…` case uses a clean message, so this gap is currently untested.

jeroenubbink (commented):
Nice cleanup — killing the `{str(e)}` leak on the catch-all and the rollback-so-no-orphan-key are both clear wins, and the taxonomy mirrors the existing `balance.py` topup handling.

Two things I'd like to resolve before this lands, both about consistency with error handling we already have (inline), plus smaller notes below.

**1. There's already a Cashu-redemption error taxonomy — this adds a third naming for the same failures.** (see inline on `redemption_error_to_http_exception`) The X-Cashu header path (the default mode) already sanitizes these in `upstream/base.py` via the shared `create_error_response` helper: `token_already_spent` (400), `invalid_token` (400), `mint_error` (422), `cashu_error` (400). This PR introduces a parallel `token_redemption_failed` for the same failure classes on the `Authorization: Bearer cashu…` path, so the two endpoints that redeem a token now disagree on the code for "already spent" (`token_already_spent` vs `token_redemption_failed`). Could we reuse `create_error_response` + the existing codes, or — if a unified `token_redemption_failed` is the goal — change both paths together so they stay in lockstep?

**2. The leak fix is only half-done.** This PR stops the raw-text leak on the bearer path, but the twin on the X-Cashu path still echoes it: `base.py:4052` (`"CASHU mint error: {error_message}"`) and `base.py:4060` (`"CASHU token processing failed: {error_message}"`) both interpolate the raw mint/exception string — and there's a second identical pair at `base.py:4720/4728`. Same bug class this PR is fixing — worth fixing there too (or a follow-up), otherwise the leak just moves to the other door.

**On the status change (401 → 400):** semantically defensible (these are payment/token failures, not auth failures), and I checked — this isn't breaking a published contract: `openapi.json` doesn't document these responses at all (they're raised in a dependency), and `docs/api/errors.md` is stale (most codes it lists don't exist in code). But it is client-observable for anyone branching on status/`code`. Two asks: (a) call the change out in the PR description as client-observable, and (b) consider declaring these via `responses=` on the route so the contract gets a machine source of truth it currently lacks.

**Cross-repo heads-up.** In `routstr-sdk`, `RoutstrClient._handleErrorResponse` gates the invalid/spent-key storage cleanup on `status === 401` — failover still fires on 400/500, but the spent token no longer gets evicted from storage on these failures, and the SDK has no handling for the new `token_redemption_failed` code at all. So as soon as this lands the SDK becomes a consumer branching on an undocumented code it doesn't recognize — worth a paired SDK update to trigger cleanup on `token_redemption_failed`.

**Minor nits:**
- **Taxonomy gap (`auth.py:94`):** one "too small for fees" case slips into the generic bucket — `_calculate_swap_amount` raises `Fees (X) exceed token amount`, re-wrapped at `wallet.py:261` as `Failed to estimate fees: …` (no `insufficient`/`melt fee` keyword), so it maps to `Failed to redeem Cashu token` rather than the specific "too small to cover swap fees". Same status, less accurate message.
- **Dead-ish branch + DRY (`auth.py:399`):** the `msats <= 0` block hand-rolls the same envelope `redemption_error_to_http_exception` builds (could share a constructor), and it's mostly defense-in-depth now that `credit_balance` raises `ValueError("Redeemed token amount must be positive…")` before returning (`wallet.py:537`) — a comment noting that would help.
- **Internal race → 400 (`auth.py:383`):** `credit_balance` raises `ValueError("API key disappeared before credit could be recorded")` (`wallet.py:556`) for a prune-vs-topup race — an internal fault that lands as a client `400 token_redemption_failed`; arguably a 500. Rare, low priority.

9qeklajc:
Thanks for the thorough review — addressed in 9dc4c97 + 66bc126:

**1. Unified taxonomy.** Bearer path now routes through the same failure classes as the X-Cashu path (`token_already_spent`/`invalid_token`/`mint_error`/`cashu_error`, carried in `type`) with matching statuses, instead of a third `token_redemption_failed` code. Both paths agree on "already spent" now.

**2. Leak fix completed on X-Cashu path.** `base.py` `mint_error`/`cashu_error` no longer interpolate raw `{error_message}` — fixed in both duplicated blocks (4052/4060 and 4720/4728).

**Broad substring gate (`auth.py`).** `invalid`/`decode` buckets now require `token` to co-occur, so `Invalid isoformat string` / SQLAlchemy `Invalid …` fall through to 500. Added a regression test with `RuntimeError("Invalid isoformat string")` → 500.

**Fee taxonomy gap.** `Failed to estimate fees: …` / `exceed token amount` now map to the specific "too small to cover swap fees" message (→ `mint_error` 422). Test added.

**`msats<=0` branch.** Aligned to the shared `cashu_error` envelope + comment noting it's now practically unreachable (`credit_balance` rejects non-positive amounts).

**Status change + SDK.** Called out as client-observable in the PR description, incl. the `routstr-sdk` `_handleErrorResponse` cleanup gate that needs a paired update for the new codes.

**Deferred:** declaring these via `responses=` on the route, the internal prune-vs-topup race → 400 (rare), and `/v1/wallet/topup` migration — left as follow-ups. Happy to fold in if you'd prefer them here.

sh1ftred:
A lot of LLM slop comments on this thread lol.

sh1ftred:
again, why is there no specific error code for mint being unreachable? That's a very specific scenario compared to all other error codes that's already in the code right now. 

token_redemption_failed doesn't tell much, it could be anything. If we can isolate mint being unreachable, which is a very real scenario, the client can simply switch to a different mint for the same node. 

none of the clankers highlighted this. smh.

9qeklajc:
we could send u all the different kind of error that u would handle, used `token_redemption_failed` to just put all mint error in one pot and when routstrd sees it, try different mint .. fallback to different mint shouldn't only happen when mint not reachable imo. every other case too .. but yeah not problem the node can send all error codes

jeroenubbink:
Re-reviewed at `e2aa023`. The three points from the last pass are resolved:

1. **Unified taxonomy** — `classify_redemption_error` is now the single classifier for all three redemption paths (bearer, X-Cashu, top-up). The parallel `token_redemption_failed` code is gone.
2. **Raw-text leak** — the X-Cashu handlers no longer interpolate the exception message; both route through the classifier, and there's a test asserting the raw error text isn't in the response.
3. **Substring vs `ValueError` gate** — the invalid/decode bucket is anchored to co-occur with `"token"`, and mint-unreachable is now detected by exception type (`is_mint_connection_error`) rather than message text.

The `mint_unreachable` 503 and the 502 on post-redemption forwarding failures are good additions — they close the "retry a spent token" case.

One process note: the final commit (`e2aa023`, "explicit code for error report & update doc") is ~1200 lines and mixes three things — the taxonomy rework, the docs, and a large unrelated test file (`test_messages_litellm_dispatch.py`, which covers the `/v1/messages` litellm dispatch, not error codes). Since upstream merges without squashing, that history is permanent, so any one of those can't be reverted later without dragging the others with it. Worth splitting into the taxonomy change / docs / messages-dispatch tests before merge.

One thing left, and I'm explicitly *not* asking you to expand this PR for it: the refund path (`/v1/wallet/refund` in `balance.py`) still has its own mint/connection substring check and echoes the raw error text into the 503/500 `detail` — the same leak this PR removes on the redemption side. It's the send side so it shouldn't reuse `classify_redemption_error`, but it could reuse `is_mint_connection_error` and drop the raw text. Fold it in only if you're already touching that file; otherwise it's a separate change.

Two follow-ups that are worth their own PRs:

- **Declare these errors in the OpenAPI schema** (`responses=` on the owning routes + a shared error model). None of them are in the machine contract today, so clients branch on codes they can't discover. With the classifier as the single source, keeping the schema in sync is straightforward.
- **A paired `routstr-sdk` update** to consume the codes: `RoutstrClient` gates spent-key cleanup on 401, but invalid/spent redemptions now return 400, and it has no handling for the new `type`/`code` or the retryable 503.

jeroenubbink (approved):
Re-reviewed at `46bbba7`. Approving.

The refund-path leak I raised last pass is fixed in `46bbba7`: `/v1/wallet/refund` now classifies mint failures via the type-based `is_mint_connection_error` instead of substring matching, and no longer echoes the raw exception text into the `503`/`500` `detail` — the raw text stays in the logs. Verified this keeps genuine transport failures at `503` (httpx `ConnectError` is caught through the cause chain) and only shifts mint-answered-with-error / internal faults to `500`, which is the right direction. Both branches are covered by tests (typed `MintConnectionError` → sanitized 503, `RuntimeError` → sanitized 500).

That completes the redemption-error sanitization consistently across bearer, X-Cashu, top-up, and now the refund send side. All three points from the earlier review are resolved and CI is green.

On the mint-unreachable question raised in the thread: this PR already gives that its own class — `cashu_mint_unreachable` / `503`, the single retryable failure — so a client can switch mints on that code specifically.

The two follow-ups (declaring these codes via `responses=` in the OpenAPI schema, and a paired `routstr-sdk` update to consume the new `type`/`code` and the retryable `503`) are still worth doing as their own PRs, but nothing here blocks merge.

PR #579, opened 2026-07-02
fix(db): increase SQLite busy timeout to 10s

jeroenubbink:
## What

Python's `sqlite3` driver (and therefore `aiosqlite`) defaults to a **5s** busy timeout, so a writer waiting for SQLite's single write lock gives up with `SQLITE_BUSY` — surfaced as a 500 — after only 5s of write contention. This bumps it to **10s**.

On the financial hot path, write transactions are tiny (e.g. the reserve `UPDATE` commits in `auth.py` *before* the upstream network call, so no write transaction spans a network `await`). Locks are therefore always short-held, and a longer wait absorbs bursts of concurrent writes without masking a genuinely stuck writer. 10s doubles the existing headroom; a much larger value would only help if a writer legitimately held the lock for that long — which doesn't happen here — while worsening resource-pinning and detection latency under overload.

## How

- Extract `create_db_engine()` and a shared `SQLITE_BUSY_TIMEOUT_SECONDS` constant in `core/db.py`; the module engine is built through it.
- Apply the same timeout to the integration test engine so the concurrent-reservation tests reproduce production lock-contention behaviour rather than the driver default.

## Tests

- `tests/unit/test_db_busy_timeout.py` — asserts the engine reports `PRAGMA busy_timeout == 10000`.
- `tests/integration/test_engine_busy_timeout.py` — guards the integration engine against silently drifting back to the default.

`make lint` clean (ruff + mypy); `make test-fast` = 668 passed, 11 skipped.

### DIFF
--- a/routstr/core/db.py
+++ b/routstr/core/db.py
@@ -11,7 +11,7 @@ from alembic.config import Config
 from alembic.util.exc import CommandError
 from sqlalchemy import UniqueConstraint, delete
 from sqlalchemy.exc import OperationalError
-from sqlalchemy.ext.asyncio.engine import create_async_engine
+from sqlalchemy.ext.asyncio.engine import AsyncEngine, create_async_engine
 from sqlalchemy.orm import aliased
 from sqlmodel import Field, Relationship, SQLModel, col, func, select, update
 from sqlmodel.ext.asyncio.session import AsyncSession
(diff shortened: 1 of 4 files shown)


jeroenubbink:
Closing for now — the existing 5s busy timeout (aiosqlite default) is fine at current load. Write locks are short-held, so this is a tuning knob rather than a fix; deferring until we actually see SQLITE_BUSY 500s under write contention. Branch kept if we want to revive it.

Issue #358, opened 2026-02-10
Generic provider API pricing display error


jooray:
The user interface for generic provider multiplies the price with million, but the price is already per million tokens, resulting in displayed price in millions :)

9qeklajc:
should be fixed by now, let me know if u still have the issue after update

PR #581, opened 2026-07-03
Resolve generic provider pricing instead of fabricating it

jeroenubbink:
## Problem

When a generic (OpenAI-compatible) upstream answers `/models` without pricing — which is the common case: DeepSeek, OpenAI, Groq and most others return bare `{id, object, owned_by}` entries — the provider silently fell back to a hardcoded **$0.001 / M tokens** and a **4096** context window.

That placeholder looks plausible but is fiction. For DeepSeek, whose real rate is ~$0.28/M in, it **undercharged live traffic by roughly 280x** — a direct money leak, served without any signal that the price was invented.

## Fix

Resolve each model through trust-ordered sources instead of fabricating:

1. **Native** — the provider's own schema (Venice's `model_spec`), authoritative when present.
2. **litellm** — the bundled cost map (curated list prices that mirror provider docs).
3. **OpenRouter** — the discovery feed (broad coverage; resale prices).
4. **Fail closed** — if no source knows the model, import it **disabled** with a warning so an operator can price it before it serves traffic, rather than invent a number.

While resolving, capture the richer metadata these sources already carry — cache read/write rates, input/output modalities, max completion tokens, tokenizer — instead of only price and context.

Context has no trustworthy last-resort source, but it is **not** a billing input, so a model that resolved a price but no context window falls back to an id-based estimate (unchanged heuristic, just relocated).

## Shape

- New `upstream/pricing_resolver.py` holds the whole *source-incomplete fallback* — the litellm→OpenRouter price chain **and** the context estimate — as one unit, so it can later be hoisted into the base provider for the other upstream types unchanged. The OpenRouter feed is fetched at most once per discovery pass, and only when a model actually misses litellm.
- `payment/models.py` gains a small `litellm_cost_entry` helper (extracted from `backfill_cache_pricing`, which now reuses it) so the resolver shares the exact same lookup semantics.
- No schema change, no UI change, no migration.

## Tests

`tests/unit/test_upstream_generic.py` drives it all through the public `fetch_models` API: native `model_spec` resolution, DeepSeek resolving via litellm (with the money assertion), a litellm-miss falling through to OpenRouter, the feed being fetched only once across models, and the fail-closed disabled-import path.

## Scope note

This is the price-*fabrication* leak. It is distinct from #358 (a UI bug that double-multiplies an already-per-million price on display) — that is not addressed here.

## Follow-ups (not in this PR)

This is the first slice of a short sequence; a couple of things are deliberately left out so this PR stays focused on the fabrication leak:

- **Unify native-schema parsing in the base provider.** Today only the generic provider consults a native `/models` schema (Venice's `model_spec`). Lifting that into `BaseUpstreamProvider` lets every upstream type share this resolver, and lets us honor *any* in-band provider pricing — including upstreams that answer in OpenRouter's shape (top-level `pricing`/`architecture`) — as native, above the litellm/OpenRouter fallbacks. The resolver was written standalone so that step is a move, not a rewrite.
- **Surface price provenance.** `ResolvedPricing.source` (`native`/`litellm`/`openrouter`/`unresolved`) is already set but not yet read — it's the seam for a later step that records and shows where each price came from, so an operator can tell a resolved price from an estimate.

### DIFF
--- a/routstr/upstream/generic.py
+++ b/routstr/upstream/generic.py
@@ -147,10 +147,13 @@ class GenericUpstreamProvider(BaseUpstreamProvider):
                     else:
                         enabled = True
 
-                    modality = (
-                        "text->text"
-                        if "image" in resolved.input_modalities
-                        else "text"
+                    # Prefer the source's own modality string (OpenRouter ships
+                    # one, e.g. "text+image->text"); otherwise derive it from the
+                    # captured input/output modalities in the same "in->out" shape
+                    # rather than flattening vision models to "text->text".
+                    modality = resolved.modality or (
+                        f"{'+'.join(resolved.input_modalities)}"
+                        f"->{'+'.join(resolved.output_modalities)}"
                     )
 
                     # A source can carry a price but no context (e.g. a litellm
--- a/routstr/upstream/generic.py
+++ b/routstr/upstream/generic.py
@@ -8,6 +8,7 @@ from .base import BaseUpstreamProvider
 from .pricing_resolver import (
     FallbackPricingResolver,
     ResolvedPricing,
+    _as_float,
     estimate_context_length,
 )
 
@@ -74,15 +75,20 @@ class GenericUpstreamProvider(BaseUpstreamProvider):
     ) -> ResolvedPricing | None:
         """Read pricing/metadata from Venice's bespoke ``model_spec`` schema.
 
-        Returns ``None`` when the upstream reported no native price (the common
-        case for bare OpenAI-compatible ``/models`` responses), so the caller
-        falls through to the shared resolution chain instead of fabricating a
-        number.
+        Returns ``None`` when the upstream reported no *usable* native price —
+        absent, non-numeric, negative, or both-zero — so the caller falls
+        through to the shared resolution chain instead of fabricating a number
+        or trusting a bogus one. This mirrors the money-safety guards the
+        litellm and OpenRouter rungs already apply: a both-zero price would
+        serve the model free, a negative one would credit the caller, and a
+        non-numeric string would otherwise throw and drop the whole catalog.
         """
         pricing_info = model_spec.get("pricing", {})
-        input_usd = pricing_info.get("input", {}).get("usd")
-        output_usd = pricing_info.get("output", {}).get("usd")
+        input_usd = _as_float(pricing_info.get("input", {}).get("usd"))
+        output_usd = _as_float(pricing_info.get("output", {}).get("usd"))
         if input_usd is None or output_usd is None:
+            return None
+        if input_usd < 0 or output_usd < 0 or (input_usd == 0 and output_usd == 0):
             return None
 
         capabilities = model_spec.get("capabilities", {})
--- a/routstr/upstream/generic.py
+++ b/routstr/upstream/generic.py
@@ -5,6 +5,11 @@ from typing import TYPE_CHECKING
 import httpx
 
 from .base import BaseUpstreamProvider
+from .pricing_resolver import (
+    FallbackPricingResolver,
+    ResolvedPricing,
+    estimate_context_length,
+)
 
 if TYPE_CHECKING:
     from ..core.db import UpstreamProviderRow
@@ -64,6 +69,35 @@ class GenericUpstreamProvider(BaseUpstreamProvider):
             "platform_url": cls.platform_url,
         }
 
+    def _native_pricing(
+        self, model_id: str, model_spec: dict
+    ) -> ResolvedPricing | None:
+        """Read pricing/metadata from Venice's bespoke ``model_spec`` schema.
+
+        Returns ``None`` when the upstream reported no native price (the common
+        case for bare OpenAI-compatible ``/models`` responses), so the caller
+        falls through to the shared resolution chain instead of fabricating a
+        number.
+        """
+        pricing_info = model_spec.get("pricing", {})
+        input_usd = pricing_info.get("input", {}).get("usd")
+        output_usd = pricing_info.get("output", {}).get("usd")
+        if input_usd is None or output_usd is None:
+            return None
+
+        capabilities = model_spec.get("capabilities", {})
+        input_modalities = ["text"]
+        if capabilities.get("supportsVision", False):
+            input_modalities.append("image")
+
+        return ResolvedPricing(
+            prompt=input_usd / 1_000_000,
+            completion=output_usd / 1_000_000,
+            context_length=model_spec.get("availableContextTokens"),
+            source="native",
+            input_modalities=input_modalities,
+        )
+
     async def fetch_models(self) -> list[Model]:
         """Fetch models from upstream API using /models endpoint."""
         from ..payment.models import Architecture, Model, Pricing, TopProvider
@@ -78,6 +112,7 @@ class GenericUpstreamProvider(BaseUpstreamProvider):
                 response.raise_for_status()
                 data = response.json()
 
+                resolver = FallbackPricingResolver()
                 models_list = []
                 for model_data in data.get("data", []):
                     model_id = model_data.get("id", "")
@@ -89,41 +124,41 @@ class GenericUpstreamProvider(BaseUpstreamProvider):
                     owned_by = model_data.get("owned_by", "unknown")
                     model_spec = model_data.get("model_spec", {})
 
-                    context_length = 4096
-                    if model_spec.get("availableContextTokens"):
-                        context_length = model_spec["availableContextTokens"]
-                    elif any(
-                        pattern in model_id.lower() for pattern in ["32k", "32000"]
-                    ):
-                        context_length = 32768
-                    elif any(
-                        pattern in model_id.lower() for pattern in ["16k", "16000"]
-                    ):
-                        context_length = 16384
-                    elif any(pattern in model_id.lower() for pattern in ["8k", "8000"]):
-                        context_length = 8192
-                    elif "gpt-4" in model_id.lower():
-                        context_length = 8192
-                    elif "claude" in model_id.lower():
-                        context_length = 200000
+                    resolved = self._native_pricing(model_id, model_spec)
+                    if resolved is None:
+                        resolved = await resolver.resolve(model_id)
 
-                    pricing_info = model_spec.get("pricing", {})
-                    input_pricing = pricing_info.get("input", {})
-                    output_pricing = pricing_info.get("output", {})
+                    if resolved is None:
+                        # Fail closed: never invent a price. Import the model
+                        # disabled with a warning so the operator can price it
+                        # (the admin UI surfaces disabled remote models).
+                        logger.warning(
+                            f"No pricing source resolved for '{model_id}' from "
+                            f"{self.upstream_name}; importing it disabled",
+                            extra={"model_id": model_id, "base_url": self.base_url},
+                        )
+                        resolved = ResolvedPricing(
+                            prompt=0.0,
+                            completion=0.0,
+                            context_length=None,
+                            source="unresolved",
+                        )
+                        enabled = False
+                    else:
+                        enabled = True
 
-                    prompt_price = input_pricing.get("usd", 0.001) / 1000000
-                    completion_price = output_pricing.get("usd", 0.001) / 1000000
+                    modality = (
+                        "text->text"
+                        if "image" in resolved.input_modalities
+                        else "text"
+                    )
 
-                    capabilities = model_spec.get("capabilities", {})
-                    input_modalities = ["text"]
-                    output_modalities = ["text"]
-
-                    if capabilities.get("supportsVision", False):
-                        input_modalities.append("image")
-
-                    modality = "text"
-                    if capabilities.get("supportsVision", False):
-                        modality = "text->text"
+                    # A source can carry a price but no context (e.g. a litellm
+                    # entry missing max_input_tokens); fall back to an id-based
+                    # estimate so we never persist a zero-length window.
+                    context_length = resolved.context_length or estimate_context_length(
+                        model_id
+                    )
 
                     spec_name = model_spec.get("name", model_name)
                     description = f"{spec_name}"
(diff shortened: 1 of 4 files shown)


9qeklajc (commented):
Took a careful pass through this, and the core idea is exactly right — the PR description made the intent easy to follow. Swapping the fabricated `$0.001/M` placeholder for a trust-ordered resolver (native → litellm → OpenRouter → fail-closed) is the correct shape, and pulling the litellm lookup into `litellm_cost_entry` so `backfill_cache_pricing` shares it is a clean touch. Importing an unpriced model **disabled + warned** instead of inventing a number is the right instinct.

One thing I'd block on before merge: the litellm branch can still let a `$0` model through **enabled**, which quietly re-opens the exact undercharge hole this PR is closing — just from a different source. A few smaller things are worth folding in while you're in here, none of them blockers. Details are inline on the relevant lines.

### 🔴 Blocker
- **litellm zero/negative prices can enable an unbillable model** — `pricing_resolver.py:90`

### 🟡 Worth fixing now
- Context window can be set from litellm's *output* cap — `pricing_resolver.py:100`
- OpenRouter bare-tail match is feed-order dependent — `pricing_resolver.py:121`
- Recomputed `modality` contradicts the resolved modalities — `generic.py:150`

### 🟢 Nice-to-have
- The tests assert litellm's *shipped* DeepSeek prices as literals (`tests/unit/test_upstream_generic.py`). Since the dep is pinned loosely (`litellm>=...`), a routine bump can redden green code — deriving the expected values from `litellm_cost_entry()` or a patched fixture would make them robust. While there, a couple of assertions on the richer metadata you're now capturing (cache rates, tokenizer, `max_completion_tokens`, and the price-resolved-but-context-missing heuristic) would lock in the new behaviour.

Overall this is a real money bug getting fixed the right way — nice work. Once the `$0`-enable case is closed I'm happy.

routstr/upstream/pricing_resolver.py
```diff
@@ -0,0 +1,177 @@
...
+    # lookup semantics used by cache-rate backfill.
+    from ..payment.models import litellm_cost_entry
+
+    info = litellm_cost_entry(model_id)
+    if info is None:
+        return None
+
+    prompt = info.get("input_cost_per_token")
+    completion = info.get("output_cost_per_token")
+    if not isinstance(prompt, (int, float)) or not isinstance(completion, (int, float)):
```
9qeklajc, line 91: This guard only rejects *non-numeric* costs — `0.0` passes straight through, so a litellm entry priced `0/0` returns a valid `ResolvedPricing` and `generic.py` imports the model **enabled** at `$0`. That's the same "served at a price nobody vetted" failure this PR exists to kill, just sourced from litellm instead of the old hardcoded placeholder.

It's not hypothetical: the bundled litellm cost map currently ships **123 both-zero entries** (`sample_spec`, the rerank models, the whole Cloudflare LoRA family, `codestral/codestral-2405`, …). Any generic upstream advertising one of those ids serves it free, enabled, with no warning.

The OpenRouter path already handles this — `_has_valid_pricing` in `payment/models.py` rejects both-zero *and* negative prices. Applying the same rule here before returning keeps them consistent and lets these fall through to OpenRouter / fail-closed:

```python
if prompt < 0 or completion < 0 or (prompt == 0 and completion == 0):
    return None
```

routstr/upstream/pricing_resolver.py
```diff
@@ -0,0 +1,177 @@
...
+        return None
+
+    input_modalities = ["text"]
+    if info.get("supports_vision"):
+        input_modalities.append("image")
+
+    return ResolvedPricing(
+        prompt=float(prompt),
+        completion=float(completion),
+        context_length=_as_int(info.get("max_input_tokens") or info.get("max_tokens")),
```
9qeklajc: `max_tokens` in litellm is the max *output* cap, not the context window (e.g. gpt-4o is `max_tokens=16384` vs `max_input_tokens=128000`). For entries that only carry `max_tokens`, this sets `context_length` to the small output number, which then feeds `_calculate_usd_max_costs` (`context_length * price`) and under-caps the billing reservation.

Only a few entries actually lack `max_input_tokens` today (`gemini/gemini-gemma-2-9b-it`, `openrouter/gryphe/mythomax-l2-13b`, `together-ai-8.1b-21b`), so the blast radius is small — but since context isn't a billing input anyway, I'd drop the `or info.get("max_tokens")` and let `estimate_context_length()` fill the gap:

```python
context_length=_as_int(info.get("max_input_tokens")),
```

routstr/upstream/pricing_resolver.py
```diff
@@ -0,0 +1,177 @@
...
+    Bare-tail matching (``deepseek-chat`` ↔ ``deepseek/deepseek-chat``) is a
+    looser, lower-trust match — OpenRouter fans a model out across resellers —
+    so an exact id match always wins first.
+    """
+    bare = model_id.split("/", 1)[-1]
+    exact = next((m for m in feed if m.get("id") == model_id), None)
+    if exact is not None:
+        return exact
+    return next(
+        (m for m in feed if m.get("id", "").split("/", 1)[-1] == bare), None
```
9qeklajc: OpenRouter fans one model out across resellers, so the bare-tail fallback can match several feed entries at different prices — and `next(...)` takes whichever appears first in feed order. That makes the resolved price depend on OpenRouter's ordering rather than on the model itself.

If you want to keep this rung, picking the cheapest match deterministically is safer than first-wins:

```python
matches = [m for m in feed if m.get("id", "").split("/", 1)[-1] == bare]
return min(
    matches,
    key=lambda m: _as_float(m.get("pricing", {}).get("prompt")) or float("inf"),
) if matches else None
```

routstr/upstream/generic.py
```diff
@@ -89,41 +124,41 @@ async def fetch_models(self) -> list[Model]:
...
+                            prompt=0.0,
+                            completion=0.0,
+                            context_length=None,
+                            source="unresolved",
+                        )
+                        enabled = False
+                    else:
+                        enabled = True
+
+                    modality = (
```
9qeklajc: This derives `modality` purely from `input_modalities`, so an OpenRouter-resolved model comes out as `text->text` even when the feed handed you an authoritative `architecture.modality` (e.g. `text+image->text`). The persisted Architecture then has a `modality` that contradicts its own `input_modalities`.

`_from_openrouter` already has the real value — carrying it through on `ResolvedPricing` and only falling back to this derivation when the source didn't provide one would keep the metadata self-consistent.

routstr/upstream/pricing_resolver.py
jeroenubbink: Good catch — fixed in 9f04525. `_from_litellm` now rejects a both-zero (or negative) entry the same way `_has_valid_pricing` does for the OpenRouter feed, so a 0/0 litellm hit (free moderation/rerank tiers) falls through instead of importing the model enabled and served for free. Added a regression test driving `omni-moderation-latest` (a real 0/0 litellm id) through the public `fetch_models` → imports disabled + warned.

routstr/upstream/pricing_resolver.py
jeroenubbink: Fixed in ccee76b. Context now comes from `max_input_tokens` only; the `or max_tokens` fallback is gone. Confirmed against the bundled map that `max_tokens` equals `max_output_tokens` for 2305/2448 models (~94%), i.e. it really is the completion cap — so a model that reports no `max_input_tokens` now falls through to the id-based estimate rather than mislabelling the output cap as the window. Test uses `gemini/gemini-gemma-2-9b-it` (max_input_tokens=None, max_tokens=8192): context resolves via the estimate, and 8192 still feeds `max_completion_tokens`.

routstr/upstream/pricing_resolver.py
jeroenubbink: Addressed in ae9748d. Before settling the direction I pulled the live OpenRouter feed and measured it: **zero bare-tail collisions** today (every bare tail is globally unique across resellers, and `:free` variants have distinct tails), so this is a latent case, not an active one. Rather than drop the rung, I made it deterministic and money-safe: among bare-tail matches it now picks the **highest-priced** candidate instead of the first in feed order. Highest (not lowest) because undercharging is the hazard here — an order-dependent or cheapest-wins match could silently bill below true cost. Exact-id match still wins ahead of any tail match. Regression test lists the pricier reseller second so a first-wins match would fail.

routstr/upstream/generic.py
jeroenubbink: Fixed in ec0fd11. Added a `modality` field to `ResolvedPricing`, populated from the OpenRouter entry's `architecture.modality`, and the generic provider now carries that through verbatim. When a source doesn't supply one (native/litellm), it derives `"{'+'.join(input)}->{'+'.join(output)}"` — so a vision model becomes `text+image->text` rather than being flattened to `text->text`. Covered by asserts on both the native (vision) and OpenRouter paths.

jeroenubbink:
Thanks for the review @9qeklajc — all four findings are addressed in four small commits (`9f04525`..`ec0fd11`), one per comment, each with a regression test:

- **Zero-price litellm guard** (`9f04525`) — reject both-zero/negative litellm hits so they fail through instead of serving for free.
- **Context source** (`ccee76b`) — take context from `max_input_tokens` only; `max_tokens` is the completion cap (==`max_output_tokens` for ~94% of the bundled map).
- **Bare-tail tie-break** (`ae9748d`) — deterministic highest-price match. I measured the live feed first: **0 bare-tail collisions** today, so this only hardens the latent case, money-safe direction.
- **Modality** (`ec0fd11`) — carry the source's modality through, derive `in->out` otherwise instead of flattening vision to `text->text`.

`make lint` + fast suite green (675 passed). Ready for another look when you have a moment.

routstr/upstream/generic.py
```diff
@@ -64,6 +69,35 @@ def get_provider_metadata(cls) -> dict[str, object]:
...
+        output_usd = pricing_info.get("output", {}).get("usd")
+        if input_usd is None or output_usd is None:
+            return None
+
+        capabilities = model_spec.get("capabilities", {})
+        input_modalities = ["text"]
+        if capabilities.get("supportsVision", False):
+            input_modalities.append("image")
+
+        return ResolvedPricing(
```
9qeklajc, line 99: This native path probably needs the same price validation as the fallback resolver. Right now any Venice-style `model_spec.pricing` with both prices at `0`, negative values, or non-numeric strings gets treated as authoritative. The first two can enable a model at a bogus/free price; the string case can throw during `/ 1_000_000` and make the whole `fetch_models()` return `[]` via the outer catch.

Could we coerce these through the same `_as_float`-style path and reject negative / both-zero native prices before returning `ResolvedPricing`? That keeps the new "never invent or accept fake money inputs" rule consistent across native, litellm, and OpenRouter sources.

routstr/upstream/generic.py
```diff
@@ -89,41 +124,44 @@ async def fetch_models(self) -> list[Model]:
...
-                    input_modalities = ["text"]
-                    output_modalities = ["text"]
-
-                    if capabilities.get("supportsVision", False):
-                        input_modalities.append("image")
-
-                    modality = "text"
-                    if capabilities.get("supportsVision", False):
-                        modality = "text->text"
+                    resolved = self._native_pricing(model_id, model_spec)
```
9qeklajc, line 133: One provider shape this still misses: a generic upstream can return OpenRouter-shaped model entries with top-level `pricing`, `architecture`, and `top_provider`. Since we only pass `model_spec` into the native parser, that provider-supplied price is ignored and we fall through to litellm / a separate OpenRouter fetch.

That means an upstream can already tell us its actual price, but we might override it with list prices, resale prices, or even disable the model if the fallbacks miss. I’d treat validated top-level `pricing` as native/provider data before litellm — probably by passing the full `model_data` into the native parse and reusing the OpenRouter-shaped parsing logic.

routstr/upstream/pricing_resolver.py
```diff
@@ -0,0 +1,197 @@
...
+    bare = model_id.split("/", 1)[-1]
+    exact = next((m for m in feed if m.get("id") == model_id), None)
+    if exact is not None:
+        return exact
+    matches = [m for m in feed if m.get("id", "").split("/", 1)[-1] == bare]
+    if not matches:
+        return None
+    return max(
+        matches,
+        key=lambda m: _as_float(m.get("pricing", {}).get("prompt")) or 0.0,
```
9qeklajc: The comment promises the bare-tail fallback picks the "highest-priced" match for money safety, but the key only looks at prompt price. If two OpenRouter entries share the same bare tail and one is cheaper on prompt but much more expensive on completion, this can still select the entry that undercharges output-heavy traffic.

I’d either fail closed on ambiguous bare-tail matches, or rank by a conservative combined/max cost that includes completion too, then add a test where prompt and completion disagree.

9qeklajc (commented):
Nice fix overall — I like that this moves the generic provider away from invented prices and makes the fallback order explicit. I left three comments that are all in the same spirit: make the fail-closed/money-safety guarantee hold for the weird provider shapes too, not only the happy-path missing-pricing case.

routstr/upstream/generic.py
jeroenubbink: Fixed in 64d9460. `_native_pricing` now coerces both prices through the resolver's `_as_float` and rejects absent / non-numeric / negative / both-zero values, falling through to the shared chain — the same guard the litellm and OpenRouter rungs already apply, now extended to the native source.

The string case was the sharpest: an unguarded `"free" / 1_000_000` threw mid-parse and the outer catch dropped the provider's *entire* catalog, not just that model. New test drives all three shapes: both-zero and negative native prices fall through to litellm's real rate, and a non-numeric price fails closed for its own model while every other model in the same response still resolves.

routstr/upstream/generic.py
jeroenubbink: Agreed in principle: an upstream's own in-band price is the most authoritative source we have and should win over the litellm/OpenRouter fallbacks, whatever shape it arrives in.

I'd rather land it with the native-schema unification than as a second parser here. `generic` is currently the only provider that reads a native `/models` schema at all (Venice's `model_spec`); the plan is to lift native parsing into `BaseUpstreamProvider` so every upstream shares this resolver, and honoring OpenRouter-shaped in-band pricing as native slots in there — reusing the entry parser `_from_openrouter` already has, rather than a new bespoke path in `generic.fetch_models`. Flagged under follow-ups in the description.

Worth noting the blast radius is small today: a real OpenRouter upstream routes through the dedicated `openrouter` provider, so this only affects a non-OpenRouter gateway that mimics the shape under `generic`. Better to do it in the refactor than bolt it on fast here.

routstr/upstream/pricing_resolver.py
jeroenubbink: Fixed in 2baea41. The tie-break now ranks by combined prompt + completion per-token cost, so it can't pick an entry that's cheap on input but dear on output — the "highest-priced wins" promise now holds whichever way traffic leans. Added a test where prompt and completion disagree (the dearer-overall entry has the lower prompt) to lock it in.

As with the first bare-tail pass, this is latent-only: still zero bare-tail collisions in the live feed, so no resolved price changes today — it just aligns the tie-break with its stated invariant.

jeroenubbink:
Thanks @9qeklajc — second round addressed, `ec0fd11..2baea41`:

- **Native price validation** (`64d9460`) — the native `model_spec` path now rejects non-numeric / negative / both-zero prices and falls through, same guard as the other rungs. Also stops one malformed entry (a string price) from throwing and dropping the whole catalog.
- **Bare-tail tie-break** (`2baea41`) — now ranks on combined prompt + completion cost, so it can't undercharge output-heavy traffic. Still zero collisions in the live feed, so latent-only.
- **In-band OpenRouter-shaped provider pricing** — agreed it should be honored as native, but I'd land it with the native-schema unification rather than a second parser in `generic` (details in the thread). Added a follow-ups note to the description covering that and the provenance seam.

`make lint` clean, fast suite green (679 passed). Ready for another look.

9qeklajc:
LGTM

PR #588, opened 2026-07-06
add model path endpoint

9qeklajc:


jeroenubbink (approved):
LGTM. Please next time add a problem description to the PR text so i also understand why we are doing something.

sh1ftred:
## Why this PR exists

Capturing the rationale here so reviewers don't have to guess (asked @9qeklajc — these are his answers):

This endpoint is being built to be consumed by **routstrd**, the downstream node project. It gives routstrd **more granular control over which upstream provider to route a request through**.

The motivation:

- Some upstream providers are clearly **worse** than others — latency, reliability, quality all vary even for the same model.
- routstrd wants to be able to **choose the better provider vs. the worse ones** when multiple are available for a single model, which directly helps end users.
- This is especially relevant through **OpenRouter**, where a single model is often reachable through several sub-providers (e.g. `openrouter:Anthropic`, `openrouter:Amazon Bedrock`, `openrouter:Fireworks`). Red's own node has 4–5 upstreams that each expose `glm-5.2`.
- The main goal is: **let routstrd specify which upstream to use on any node.**

So `/v1/models/paths` is discovery data — it tells the client (routstrd) which provider paths exist for each model so it can make an informed routing choice, rather than relying on the proxy's internal cheapest-first selection.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+    return "openrouter.ai" in (base_url or "")
+
+
+def exposed_model_id(model: object) -> str:
+    """Client-visible ``/v1/models`` id for a cached model."""
+    forwarded = getattr(model, "forwarded_model_id", None)
+    return forwarded or getattr(model, "id")
+
+
+def public_model_id(model_id: str) -> str:
```
jeroenubbink, line 164: **Edit 2026-07-17: withdrawn — see reply below. Keeping the original struck through for context.**

~~Follow-up on my approval, flagging this early so the table shape does not need a migration later: keying the paths table by this unqualified tail merges two genuinely different models that share a trailing segment (e.g. an OpenRouter `deepseek/deepseek-v4-pro` next to a generic provider serving a bare `deepseek-v4-pro`) — a per-model query then returns paths the model is not actually reachable through, and their prices can differ.~~

~~We are moving toward making both spellings first-class: a model's identity is its exposed id (`forwarded_model_id or id`, exactly what `exposed_model_id` above already computes), the bare tail stays a convenience alias, and when two exposed ids collide on a tail both get listed/pickable in `/v1/models` rather than collapsed. Same direction as the provider-scoping in #589 and the per-model provenance in #617.~~

~~Suggestion: store `ModelPathRow.model_id` as the exposed id instead of the tail, and let the `/v1/models/paths/model` lookup match tail as a secondary convenience (exact exposed-id match first). That keeps today's query UX while paths stay attributable to the right model. Not blocking this PR — happy to do the re-key as a follow-up on our side if you prefer, but if it is cheap to switch the stored key now it saves a migration.~~

jeroenubbink (commented):
see inline comments

routstr/upstream/model_paths.py
jeroenubbink: Withdrawing the suggestion above. After discussing this within the team we're aligning on tail-as-identity: `deepseek/deepseek-v4-pro` and a bare `deepseek-v4-pro` are the same model, provider prefixes are not part of a model's identity, and choosing between upstream paths is a client-side concern. Under that model, keying the paths table by the unqualified id is the right shape — please disregard the re-key proposal.

jeroenubbink:
Building on the rationale captured above (routstrd choosing the upstream path): two questions about the next step, since this PR is discovery-only and routing still picks the provider per request.

1. **What's the request-side mechanism to pin a path?** Some spellings already narrow routing implicitly (OpenRouter's qualified id, a per-provider `forwarded_model_id`), but they aren't advertised anywhere, nothing guarantees a unique match, and the `openrouter:<sub>` paths this endpoint enumerates can't be selected via model id at all. Is the plan a request field, a header, or path-qualified ids — and does it belong in the new models upstream endpoints work?
2. **Will the paths payload grow per-path metadata** — at minimum the exact string to request and that path's pricing? `/v1/models` exposes a single price per model, so a client comparing paths has nothing to compare on price. It also matters node-side: max_cost reserve and settlement need the serving path's pricing (we're preparing a fix for the settlement half separately).

jeroenubbink:
One more suggestion before this merges, building on the questions above: since the follow-up work will need the paths payload to carry more (at minimum the exact model string to request per path, likely pricing), could we mark the endpoint shape as experimental for now — e.g. a `version` field in the payload or a line in the endpoint docs? That way routstrd/sdk can consume it behind an adapter treating path values as opaque, and the shape stays free to evolve without a cross-repo breaking change. Cheap now, expensive after clients pin it.

9qeklajc:
> Building on the rationale captured above (routstrd choosing the upstream path): two questions about the next step, since this PR is discovery-only and routing still picks the provider per request.
> 
>     1. **What's the request-side mechanism to pin a path?** Some spellings already narrow routing implicitly (OpenRouter's qualified id, a per-provider `forwarded_model_id`), but they aren't advertised anywhere, nothing guarantees a unique match, and the `openrouter:<sub>` paths this endpoint enumerates can't be selected via model id at all. Is the plan a request field, a header, or path-qualified ids — and does it belong in the new models upstream endpoints work?
> 
>     2. **Will the paths payload grow per-path metadata** — at minimum the exact string to request and that path's pricing? `/v1/models` exposes a single price per model, so a client comparing paths has nothing to compare on price. It also matters node-side: max_cost reserve and settlement need the serving path's pricing (we're preparing a fix for the settlement half separately).

1. this doesn't change the way the inference work, and should be seen as something completely different .. in the future clients like routstrd fetches the existing model paths for a given model and could set the prefered path in a header (for example X-ROUTSTR-MODEL-PATH) which then routes to the correct upstream .. 
2. it's not for comparison it's a client preference for now, we could build on top of this later and have the prices set per upstream too

9qeklajc:
> One more suggestion before this merges, building on the questions above: since the follow-up work will need the paths payload to carry more (at minimum the exact model string to request per path, likely pricing), could we mark the endpoint shape as experimental for now — e.g. a `version` field in the payload or a line in the endpoint docs? That way routstrd/sdk can consume it behind an adapter treating path values as opaque, and the shape stays free to evolve without a cross-repo breaking change. Cheap now, expensive after clients pin it.

I'd rather add a v2 in the path instead e.g. `/v1/models/paths` -> `/v2/models/paths`

sh1ftred:
reviewing this rn.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+    for provider in provider_rows:
+        if not provider.enabled:
+            continue
+        if provider.id is not None:
+            enabled_provider_ids.add(provider.id)
+        for model in provider.models:
+            if model.enabled:
+                overrides_by_id[model.id] = (model, provider.provider_fee)
+            else:
+                disabled_model_ids.add(model.id)
```
jeroenubbink: **Item 1 — cross-provider model-disable isolation, reintroduced.**

`disabled_model_ids` is a global `set[str]` keyed by bare model id, so it drops the provider scoping routing uses. Consumed at 215: `if not getattr(model, "enabled", True) or model_id in disabled_model_ids: continue`.

Compare `proxy.py:174`, which does it correctly — `model_key = (model.id.lower(), model.upstream_provider_id)` into `disabled_model_keys: set[tuple[str, int]]`. `ModelRow` has a **composite primary key** `(id, upstream_provider_id)` (db.py:289-292), so the same id legitimately exists on several providers.

Reproduced: provider 1 (anthropic) caches `shared-model`; provider 2 has `ModelRow(id='shared-model', upstream_provider_id=2, enabled=False)`. `refresh_model_paths([p1, p2])` then `get_all_model_paths()` returns `[]`. Expected `[{'id': 'shared-model', 'paths': [{'path': 'anthropic'}]}]`.

Routing still serves it via provider 1 and `/v1/models` still lists it, so discovery tells clients a live model is reachable through nothing. This is the same isolation bug https://github.com/Routstr/routstr-core/pull/589 fixed on the routing side.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+    enabled_provider_ids: set[int] = set()
+
+    for provider in provider_rows:
+        if not provider.enabled:
+            continue
+        if provider.id is not None:
+            enabled_provider_ids.add(provider.id)
+        for model in provider.models:
+            if model.enabled:
+                overrides_by_id[model.id] = (model, provider.provider_fee)
```
jeroenubbink: **Item 2 — same missing provider dimension, on the override map.**

`overrides_by_id[model.id] = (model, provider.provider_fee)` is last-writer-wins across providers, and it is applied at 218-223 with no check that the row belongs to *this* provider. So one provider's `forwarded_model_id` alias **and its `provider_fee`** land on a different provider's cached model.

`algorithm.py:200-202` guards exactly this, keyed on `(model.id.lower(), upstream_db_id)`, with the comment `# Apply overrides only for this provider's model row.`

Reproduced: provider 1 (anthropic) caches `shared-model` with no DB row; provider 2 has `ModelRow(id='shared-model', upstream_provider_id=2, forwarded_model_id='openai-alias')`. Result: `[{'id': 'openai-alias', 'paths': [{'path': 'anthropic'}, {'path': 'openai'}]}]` — provider 2's private alias stamped onto provider 1's path, and the real id gone entirely.

The second loop's own guard at 236 shows the scoping was understood; it also can never work, because only one row per id survives the collapsed dict, so a provider's DB-only override is silently dropped whenever another provider has a row with the same id.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+) -> None:
+    """Replace all rows for ``upstream_provider_id`` with ``pairs``.
+
+    Replacement (not upsert) so stale paths disappear when provider config or
+    upstream availability changes.
+    """
+    unique_pairs = list(dict.fromkeys(pairs))
+    async with create_session() as session:
+        await session.exec(  # type: ignore[call-overload]
+            delete(ModelPathRow).where(
```
jeroenubbink: **Item 3 — a transient upstream failure deletes good persisted data, and the graceful path is more destructive than the crash path.**

`delete(ModelPathRow).where(...)` executes even when `unique_pairs` is empty, and every degradation path returns `[]` rather than signalling "unknown": request exception (112), 429 (119), non-200 (125), bad payload (134).

So one rate-limited cycle blanks an OpenRouter provider's entire path set for ~660 s while `/v1/models` keeps listing every model — `refresh_models_cache` (base.py:5202) deliberately *retains* `_models_cache` on failure, so the two surfaces diverge.

Meanwhile `refresh_model_paths`'s `except Exception` (378) **skips** the persist and preserves the old rows. Graceful failure is punished; hard failure is not.

Compounding it: `_prune_inactive_provider_paths` builds `stmt = delete(ModelPathRow)` and only adds the WHERE `if active_provider_ids:` (340-344). An empty upstream list issues an **unfiltered `DELETE FROM model_paths`** — reachable milliseconds into boot, since `main.py:120-122` swallows a failing `initialize_upstreams()` via `return_exceptions=True`, leaving `_upstreams == []`.

The test suite currently enshrines this: `await mp.refresh_model_paths([]); assert await mp.get_all_model_paths() == []`.

Suggest merge-on-success / skip-on-unknown, and never issuing the prune with an empty active set.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+                col(ModelPathRow.upstream_provider_id) == upstream_provider_id
+            )
+        )
+        for model_id, path in unique_pairs:
```
jeroenubbink: **Item 4 — ~1 s exclusive SQLite write lock per provider per refresh, on the same file as billing.**

Rows go in one-by-one via `session.add()` inside a single transaction. Measured on this head at realistic OpenRouter scale (400 models x 5 endpoints = 2000 pairs): **2000 separate INSERT statements, 1011 ms of wall time in one transaction**. The ORM reads back `lastrowid` per row on aiosqlite, so `insertmanyvalues` never engages.

WAL still serializes writers, and the node writes balance reservations and settlements on the hot path against this same file with a 5 s busy timeout. An independent measurement of a competing writer showed it blocked **1136 ms**. That is a >1 s write stall injected into billing every 600 s, per provider.

A chunked Core bulk insert (`insert(ModelPathRow).values([...])`) measured **1 statement / 315 ms**.

Related write amplification: `ix_model_paths_model_id` is fully subsumed by the leading column of the `uq_model_paths_model_path_provider` autoindex — `EXPLAIN` always picks `sqlite_autoindex_model_paths_1` — so it costs 2000 extra B-tree insertions per refresh for zero read benefit.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+        endpoints = resp.json().get("data", {}).get("endpoints", [])
+    except Exception as e:  # noqa: BLE001
+        logger.warning(
+            "OpenRouter endpoint discovery bad payload",
+            extra={"author_slug": author_slug, "error": str(e)},
+        )
+        return []
+
+    paths: list[str] = []
+    for endpoint in endpoints:
```
jeroenubbink: **Item 5 — the loop sits outside the try that is meant to swallow bad payloads.**

Line 128 is inside the try/except at 127-134, but this loop is not. And `.get("endpoints", [])` never applies its default when the key is present with value `None`.

Reproduced against `_fetch_openrouter_endpoint_paths`:
- `{"data": {"endpoints": null}}` -> `TypeError: 'NoneType' object is not iterable`
- `{"data": {"endpoints": "none"}}` -> `AttributeError: 'str' object has no attribute 'get'` (it iterates characters)

Both escape a helper whose docstring promises "Failures (network, rate limit, bad payload) are logged and swallowed so one model never breaks the whole refresh". They get caught two frames up by `gather(return_exceptions=True)` and surface as the generic `"OpenRouter endpoint discovery task errored"` (301) — the bucket an operator reads as a bug rather than an upstream payload shape. The model then silently contributes nothing, which feeds the row-wipe in item 3.

Fix: move the loop inside the try, or `endpoints if isinstance(endpoints, list) else []`.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+) -> list[object]:
+    """Return provider models after DB disabled/override state is applied."""
+    overrides_by_id = overrides_by_id or {}
+    disabled_model_ids = disabled_model_ids or set()
+    visible_models: list[object] = []
+    seen_model_ids: set[str] = set()
+
+    for model in upstream.get_cached_models():
+        model_id = getattr(model, "id", "")
+        if not getattr(model, "enabled", True) or model_id in disabled_model_ids:
```
jeroenubbink: **Item 6 — case-sensitive matching here, case-insensitive in routing, and inconsistent within this function.**

Lines 215 and 218 compare raw, while routing lowercases both sides (proxy.py:174 building, algorithm.py:191 consuming). `litellm_cost_entry` (payment/models.py:88-96) exists precisely because mixed-case ids like `deepseek-ai/DeepSeek-V4-Flash` are real.

Reproduced: cached id `deepseek-ai/DeepSeek-V4-Flash`, DB row `deepseek-ai/deepseek-v4-flash` with `forwarded_model_id='public-alias'`. Routing applies the override and `/v1/models` advertises `public-alias`; `get_all_model_paths()` returns `[{'id': 'DeepSeek-V4-Flash', ...}]`.

The mirror case is worse: line 228 records `model_id.lower()`, so the second loop's `if model_id.lower() in seen_model_ids` (240) then skips the override row as well — dropped end to end. With a *disabled* mixed-case row, discovery advertises a model the node refuses to serve.

Fixed by the same `(id.lower(), upstream_provider_id)` keying as items 1 and 2.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+def openrouter_author_slug(model: object) -> str | None:
+    """Return a canonical ``author/slug`` for the OpenRouter endpoints API.
+
+    OpenRouter requires the canonical id, never ``forwarded_model_id``. Prefer
+    ``canonical_slug``, then a slash-containing ``id``; otherwise there is no
+    usable form and endpoint discovery is skipped for this model.
+    """
+    canonical = getattr(model, "canonical_slug", None)
```
jeroenubbink, line 183: **Item 7 — `openrouter_author_slug` refuses `forwarded_model_id`, which is exactly the value the proxy forwards upstream.**

Lines 78-84 try `canonical_slug`, then a slash-containing `id`, else `None`; the docstring asserts "OpenRouter requires the canonical id, never `forwarded_model_id`".

But base.py:2719 and base.py:3090 both forward `(model_obj.forwarded_model_id or model_obj.id)`, and admin.py:542 always sets `forwarded_model_id = payload.forwarded_model_id or payload.id`.

Reproduced: `id='my-alias'`, `forwarded_model_id='anthropic/claude-opus-4.6'`, `canonical_slug=None` gives `openrouter_author_slug -> None` while `exposed_model_id -> 'anthropic/claude-opus-4.6'`. The author/slug OpenRouter needs is in hand and discarded, `_for_model` returns `[]`, and every admin-created OpenRouter alias row ends up invisible in `/v1/models/paths` though it routes fine and its responses stamp `openrouter:Anthropic`.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+    Emits the direct ``<provider_type>`` path for normal upstreams. For
+    OpenRouter-compatible providers, emits one path per OpenRouter sub-provider
+    endpoint, prefixed the same way response stamping prefixes it.
+    """
+    provider_type = (upstream.provider_type or "").strip()
+    models = _apply_model_visibility(upstream, overrides_by_id, disabled_model_ids)
+    is_openrouter = is_openrouter_base_url(upstream.base_url)
+
+    pairs: list[tuple[str, str]] = []
+    if not is_openrouter:
```
jeroenubbink: **Item 8 — the emitted path strings do not match what `_apply_provider_field` actually stamps, for three separate provider shapes.**

That equivalence is the contract both the module docstring and `docs/api/endpoints.md` assert.

(a) The `if not is_openrouter:` early return (268-272) makes the bare `<provider_type>` path and the `<provider_type>:<sub>` paths mutually exclusive, but base.py:257-266 chooses between them **per response**, based on whether the upstream reported its own `provider`. A generic-at-openrouter upstream whose `/endpoints` call fails gets zero rows while its responses stamp a bare `generic` — a value discovery now claims does not exist.

(b) `OpenRouterUpstreamProvider._apply_provider_field` (openrouter.py:40-45) strips repeated prefixes and maps a missing or self-echoing sub-provider to the literal `"unknown"`. Line 140's `f"{path_prefix}:{provider_name}"` can produce neither, so when the endpoints API returns `provider_name: "OpenRouter"`, discovery advertises `openrouter:OpenRouter` while every real response is stamped `unknown`.

(c) `is_openrouter_base_url` only matches `openrouter.ai`, so any other aggregator reporting its own provider field is mis-mirrored. `RoutstrUpstreamProvider` (routstr.py:18, no override) emits `routstr:anthropic` / `routstr:unknown` on the wire but only a bare `routstr` in discovery — a 0% hit rate for path-matching clients.

Right depth for the fix: a `discovery_paths_for(model)` hook on `BaseUpstreamProvider`, sitting beside `_apply_provider_field` and overridden in `OpenRouterUpstreamProvider`, so the two cannot drift.

(Separately: `generic:Anthropic` is correct — the docs' `generic:my-upstream` is the wrong artifact.)

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+        for model in models:
+            if provider_type:
+                pairs.append((exposed_model_id(model), provider_type))
```
jeroenubbink: **Item 9 — the path is a provider *type*, not a provider identity, so providers are indistinguishable — and on a stock deployment every path is the literal `base`.**

`UpstreamProviderRow.provider_type` is a non-unique column documented as "Provider type: custom, openai, anthropic..." (db.py:549-551). The unique stable identity is `slug` (db.py:543-548, `unique=True`), which is never read here. Two `generic` upstreams at different vendors collapse to one `generic` path, as do two Anthropic keys.

Worse for the commonest shape: `_seed_providers_from_settings` writes `provider_type="custom"` (helpers.py:376) for env-configured nodes; `_instantiate_provider` maps `custom` to `BaseUpstreamProvider.from_db_row` (helpers.py:422-423); and `from_db_row` (base.py:160-163) never copies `provider_type` onto the instance. So the class default `provider_type = "base"` (base.py:96) wins, and **every model on a stock single-upstream node reports `{"path": "base"}`** — a string absent from the docs that identifies nothing.

Separately, `create_model_mappings` keeps only the lowest-fee provider per `base_url` (algorithm.py:142-145), so a duplicate provider row on the same URL is unroutable yet still advertised here.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+                upstream.base_url,
+                upstream.api_key,
+                author_slug,
+                provider_type,
+                semaphore,
+            )
+            model_id = exposed_model_id(model)
+            return [(model_id, path) for path in paths]
+
+        results = await asyncio.gather(
```
jeroenubbink, line 463: **Item 10 — default-ON fan-out with no kill switch, no caching and no 429 backoff, whose worst-case pass exceeds its own interval.**

~400 OpenRouter models at `_OPENROUTER_CONCURRENCY = 5` and `_OPENROUTER_TIMEOUT_SECONDS = 10.0` is 80 rounds x 10 s = **800 s worst case, longer than the 600 s default** — so `await asyncio.sleep(...)` (421) is never a rest period and 5 sockets stay held around the clock. Steady state is ~57,600 requests/day per provider, re-fetching endpoint lists that change on the order of days.

It is also default-ON: `model_paths_refresh_interval_seconds: int = Field(default=600, ...)` (settings.py:98), unlike its sibling `providers_refresh_interval_seconds: int = Field(default=0, ...)` (settings.py:89), and there is no `enable_model_paths_refresh` boolean beside `enable_pricing_refresh` / `enable_models_refresh` (settings.py:101-102). Every node that upgrades with an OpenRouter provider silently starts spending the operator's API key whether or not anything consumes the endpoint.

The 429 branch (114-119) only logs and returns `[]` — no backoff, no cooldown, no reduction of the next burst — and each empty result feeds the row-wipe in item 3.

Because `is_openrouter_base_url` matches any base_url containing `openrouter.ai` and providers are awaited sequentially (368), a node running native `openrouter` plus one generic upstream there fetches every slug **twice** for byte-identical bodies, differing only in the prefix applied at 140.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+
+
+def public_model_id(model_id: str) -> str:
+    """Model id exposed by model-path API responses.
+
+    Provider-prefixed ids such as ``z-ai/glm-5v-turbo`` are returned as
+    ``glm-5v-turbo`` so clients can search and display the same unqualified id
+    they pass to ``/v1/models/paths/model``.
+    """
+    return model_id.rsplit("/", 1)[-1]
```
jeroenubbink: **Item 11 — `public_model_id` splits on the last slash while every other id-normalizing site splits on the first, so the two endpoints expose ids that cannot be joined.**

`rsplit("/", 1)[-1]` here, versus `algorithm.py:160` `get_base_model_id` = `model_id.split("/", 1)[1] if "/" in model_id else model_id`, and helpers.py:40 inside `resolve_model_alias`. They agree only for ids with exactly one slash.

Three-segment ids are real: `accounts/fireworks/models/glm-5` appears in tests/unit/test_messages_litellm_dispatch.py:1103,1319, and gemini.py:82 documents `gemini/google/gemini-3.1-flash-lite-preview`.

Verified: for `accounts/fireworks/models/glm-5`, `/v1/models` reports `fireworks/models/glm-5` while `/v1/models/paths` reports `glm-5`. A client reading the paths id and sending `{"model": "glm-5"}` hits `get_candidates('glm-5')`, which has no such alias registered, and gets a 404.

To be clear about scope: the unqualified-base-id collapse is a deliberate ecosystem decision and I am not questioning it. The defect is that discovery uses a *different* rule from the rest of the system.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+async def get_paths_for_model(model_id: str) -> list[dict]:
+    """Paths for a single model, shaped for ``GET /v1/models/paths/model``.
+
+    Match by the public, unqualified model id, mirroring the model cache alias
+    behavior. Both ``deepseek-v4-pro`` and ``deepseek/deepseek-v4-pro`` resolve
+    every row whose stored id has the same base model id.
+    """
+    requested_id = public_model_id(model_id)
+    async with create_session() as session:
+        rows = (
```
jeroenubbink: **Item 12 — both unauthenticated read routes materialize the entire table per request, and the two disagree on ordering.**

Lines 454-461 have no predicate; line 466 then discards nearly everything via `if public_model_id(row.model_id) != requested_id: continue`. `EXPLAIN QUERY PLAN` confirms `SCAN model_paths` plus `USE TEMP B-TREE FOR ORDER BY` — the migration's index can never be used.

Measured on 1,976 rows:
- `GET /v1/models/paths/model?model_id=...` — **77.7 ms to return 78 bytes** (~1000:1)
- `GET /v1/models/paths` — **85.5 ms for a 77.5 KiB body**, so ~12 req/s on one core

`models_router` carries no auth dependency (main.py:426) and shares the event loop with the billing proxy, so a trivial unauthenticated loop starves paying inference. ORM hydration runs on the loop thread, so aiosqlite does not offload it. Pushing the filter into SQL measured **14.4 ms** (11x).

Ordering is also incoherent: `get_all_model_paths` sorts by the raw stored `model_id` (430) but keys the response by the stripped `public_model_id` (436), so rows `anthropic/zebra` and `openai/aardvark` emit as ids `["zebra", "aardvark"]` — the ORDER BY buys nothing a consumer can rely on — while `get_paths_for_model` orders by `path, upstream_provider_id, model_id` (456-459). Within a model, `paths` order follows rowid, and since every cycle deletes and reinserts, the arrays reorder across refreshes with no data change, defeating client diffing and ETags.

Minor, same area: `order_by(ModelPathRow.model_id)` / `order_by(ModelPathRow.path)` (430, 456-459) type-check only by accident — mypy sees `str` and SQLAlchemy accepts a `str` as a raw SQL *label*. That is why only the int column had to be forced into `col()`.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+                extra={
+                    "provider": upstream.provider_type or upstream.base_url,
+                    "db_id": upstream.db_id,
+                    "error": str(e),
+                    "error_type": type(e).__name__,
+                },
+            )
+
+
+async def refresh_model_paths_periodically(
```
jeroenubbink, line 695: **Item 13 — no invalidation hook at all, an inert CASCADE, and rowid reuse that re-attributes orphans to unrelated providers.**

`refresh_model_maps()` is called at admin.py:581/635/654/745/945/970/1005/1019 and proxy.py:68. Nothing calls `refresh_model_paths` — it has **no caller besides the timer**, so nothing runs it at boot, on an admin mutation, or on demand. After an operator disables a provider, `/v1/models` updates immediately while `/v1/models/paths` keeps advertising it for ~660 s.

Deletion is worse. `ModelPathRow` declares `ondelete="CASCADE"` (db.py:341), but `PRAGMA foreign_keys` reads **0** (db.py:769 sets only `journal_mode`), and unlike `ModelRow` there is no `Relationship` on `UpstreamProviderRow` to trigger `cascade="all, delete-orphan"` (db.py:564-567). Verified: `session.delete(provider)` (admin.py:1016) left `[('m1', 'anthropic', 1)]` behind; inserting `upstream_provider_id=99999` succeeded. And SQLite **reuses rowids** — delete provider id 1, create another, and you get id 1 again, so those orphans are silently re-attributed to a completely different provider.

Disabling the feature does not work either: main.py:134 skips task creation when the interval is <= 0, so the documented way to turn it off (`docs/provider/configuration.md`, "set `0` to disable") instead leaves the routes registered serving a **permanently frozen** snapshot — no expiry, no prune, no 503. `interval` is also read once before the loop (398), so the DB-persisted setting cannot take effect until restart, and 0 -> N never starts the task.

Finally, `ModelPathRow` has no `updated_at` and the envelope is a bare `{"data": [...]}` (payment/models.py:601), so a consumer cannot distinguish 10-minute-old data from 10-month-old data.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,472 @@
...
+def _row_to_visible_model(
+    model_id: str,
+    row: ModelRow,
+    provider_fee: float,
+) -> Model | None:
+    """Convert an enabled DB override row into a routed model object."""
+    from ..payment.models import _row_to_model
+
+    try:
+        return _row_to_model(row, apply_provider_fee=True, provider_fee=provider_fee)
```
jeroenubbink: **Item 14 — every enabled override row is rebuilt into a fully priced `Model` on the event loop, to read two fields that are already plain columns.**

`_row_to_visible_model` calls `_row_to_model(row, apply_provider_fee=True, provider_fee=provider_fee)` from `_apply_model_visibility` at 220 and 242.

`_row_to_model` (payment/models.py:224-290) runs 4x `json.loads`, 3x `Pricing.parse_obj`, `backfill_cache_pricing` (a litellm cost-map lookup), `_calculate_usd_max_costs`, `sats_usd_price()` and `_update_model_sats_pricing` — **measured at 0.698 ms/call** on this checkout.

`_apply_model_visibility` is a plain `def` with no `await` in either loop, so a node with 1,000 enabled override rows burns ~0.7 s of **uninterruptible CPU** every cycle, during which no billing settle, streaming chunk relay, or reservation release can run.

All of it is discarded except `getattr(model, 'enabled', True)` (225, 245) and `forwarded_model_id or id` (57-58) — both readable straight off `override_row`.

Secondary: `_row_to_model`'s `except Exception: logger.warning(f"Could not calculate sats pricing: {e}")` (payment/models.py:287-288) fires once per row, so a node with a failing BTC price fetch emits N warning lines per refresh. I saw this in my own repro runs.

tests/unit/test_model_paths.py
```diff
@@ -0,0 +1,712 @@
...
+
+
+@pytest.mark.asyncio
+async def test_direct_provider_single_path_uses_provider_type() -> None:
+    provider = _FakeProvider(
+        provider_type="anthropic",
+        base_url="https://api.anthropic.com/v1",
+        models=[_model("claude-opus-4.6")],
+    )
+    pairs = await mp._collect_provider_paths(provider)  # type: ignore[arg-type]
```
jeroenubbink: **Item 15 — the suite is green with all fourteen items above present, via four independent seams.**

**(1) Private-helper seams.** 27 references drive internals: `mp._collect_provider_paths(...)  # type: ignore[arg-type]` at L200/211/225/259/290/314/338/372 — the `type: ignore` on every single one is itself the signal that the seam is wrong — plus 18 `_persist_provider_paths` seeding calls, and monkeypatched `mp._OPENROUTER_CONCURRENCY` (L346) and `mp._collect_provider_paths` (L644).

Crucially, the only two-provider test, `test_same_model_two_providers_two_paths` (L505), seeds via `_persist_provider_paths` and therefore **bypasses `_load_model_visibility` entirely** — which is exactly why items 1, 2 and 6 ship green.

**(2) `httpx` monkeypatching.** `monkeypatch.setattr("httpx.AsyncClient.get", _fake_get)` couples each fake's signature to the call site, and a mismatch raises `TypeError` into the SUT's own `except Exception` (model_paths.py:107), returning `[]` — precisely what the tests assert. Verified by adding `follow_redirects=True` to the SUT: only 2 of 5 OpenRouter tests failed. L295, L319 and L343 (`assert state["max"] <= 3`, trivially true when the fake is never invoked) stayed green while covering nothing. Deleting the entire 429 branch also leaves every test green.

**(3) The fixture violates the invariant it documents.** `patched_session` comments `# Seed the FK target so ModelPathRow inserts satisfy the constraint.` then seeds only `for pid in (1, 2):` (L117), while L538/539/563/589/591 insert for provider ids 4, 5 and 7. Verified to raise `IntegrityError: FOREIGN KEY constraint failed` under `PRAGMA foreign_keys=ON`. Those tests pass only because of the very missing PRAGMA that makes item 13's CASCADE dead — and deleting the `foreign_key=` / `ondelete=` declaration outright would not turn a single test red.

**(4) The production entry point is untested.** `grep -rn refresh_model_paths_periodically tests/` returns 0 hits, though `tests/unit/test_models_refresh_loop.py` already establishes the pattern for the sibling loop (L37 picks-up-new-providers, L100 disabled-when-non-positive). The first of those is exactly the test that would have caught the interval snapshot in item 13.

Worth adding: a two-provider test that goes through the public refresh path rather than `_persist_provider_paths`, and a real transport-level fake (or `respx`) so a signature drift fails loudly instead of silently returning `[]`.

jeroenubbink (changes requested):
**This is a fresh review, and it supersedes my approval of 2026-07-16.** That earlier pass was a read-through of the endpoint shape and it approved on that basis. Since then the branch picked up new commits, and I re-reviewed this properly — building the module out at realistic scale, running the failure paths, and measuring the hot paths. What came back is materially different from what I signed off on, so I am withdrawing the approval and requesting changes rather than leaving a stale green on the PR. Apologies for the churn — the earlier approval was too shallow for a module that adds a public unauthenticated endpoint plus a background loop on the billing node.

Everything below is reproduced or measured on head `4c29258`, not read off. 15 inline comments; the ones I would block on are the first three.

### The one that matters most

`_load_model_visibility` / `_apply_model_visibility` key their maps on the **bare model id**, dropping the provider dimension that routing carries everywhere else. `ModelRow` has a composite primary key `(id, upstream_provider_id)`, so ids legitimately repeat across providers. The consequence is that disabling a model on one provider erases it from `/v1/models/paths` for **every** provider, and one provider's `forwarded_model_id` alias and `provider_fee` get applied to a different provider's model.

That first half is the cross-provider model-disable isolation bug that https://github.com/Routstr/routstr-core/pull/589 fixed on the routing side back in July — reintroduced here in a new module. `proxy.py:174` builds `disabled_model_keys: set[tuple[str, int]]` and `algorithm.py:200-202` applies overrides only for the provider's own row, with a comment saying exactly that. This module is the only copy that does neither.

**Root cause, and the cheapest fix:** these two functions are a *third* open-coded copy of "cached provider models + enabled DB overrides − disabled rows" — alongside `refresh_model_maps` (proxy.py:153), `create_model_mappings` (algorithm.py:181), and the unused `get_all_models_with_overrides` (helpers.py:72). Keying both maps on `(model.id.lower(), upstream_provider_id)` fixes items 1, 2 and 6 in one change. The deeper fix is to not re-derive at all: only the OpenRouter sub-provider list is genuinely new data here; the rest is a projection of `_provider_map`, which is already refreshed every 60 s **and** synchronously on every admin mutation.

### Data-loss and hot-path items

The refresh is replace-not-merge, and every degradation path returns `[]` rather than signalling "unknown" — so a single rate-limited cycle **deletes** a provider's entire persisted path set while `/v1/models` keeps listing every model. The exception path is safer than the graceful path, which is inverted. Related: `_prune_inactive_provider_paths` issues an **unfiltered `DELETE FROM model_paths`** when the active-provider list is empty, which is reachable at boot.

The persist holds SQLite's exclusive write lock for ~1 s per provider per cycle (measured, 2000 pairs), and the two read routes are unauthenticated full-table scans sharing the event loop with the billing proxy (measured ~12 req/s on one core). Both land directly on money-path latency.

### Coverage

The suite is green with every one of these present. I have put the detail on the test file, but the short version: the only two-provider test seeds via `_persist_provider_paths` and therefore never exercises `_load_model_visibility`, which is precisely why the top three items ship green. Deleting the entire 429 branch also leaves every test green, and `refresh_model_paths_periodically` — the module's only production entry point — has zero test references.

### Three smaller things, no inline anchor

- The migration retains the `# ### commands auto generated by Alembic - please adjust! ###` markers (lines 20 and 51). Migrations here are hand-written because autogenerate is broken by a legacy table. The emitted DDL is faithful and the chain is fine, so this is a process point, not a broken migration.
- `ruff format --check` fails on this branch's own code (`routstr/core/db.py:334-336`, `tests/unit/test_model_paths.py:347,577`). CI only runs `ruff check`, so the next `make format` will fold these reformats into an unrelated commit.
- The migration chains onto `7f2843d3f4e4`, but main's head moved to `9c4d8e2f1a6b` when the fee-migration reapply landed. This will double-head on merge and needs a re-chain regardless of the above.
- Minor: the four commits on the branch carry no conventional-commit prefixes, and upstream merges without squashing, so the granularity is permanent.

### Verified clean, so nobody re-spends the time

Single alembic head with an intact chain; the migration-built schema is byte-equivalent to `create_all` (`run_migrations()` precedes `init_db()`, so no collision); downgrade is correct; no route shadowing — all four new routes are static and registered ahead of `/{path:path}`, and a missing `model_id` returns a clean 422; `except Exception` cannot swallow `CancelledError` on 3.11 and cancellation propagates correctly out of the fan-out (I traced a full lifespan shutdown); no busy loop; the semaphore is released only after the body is fully read; the unique constraint cannot raise `IntegrityError` because `dict.fromkeys` dedupes first; `sqlmodel 0.0.24` does support `Field(ondelete=...)` — it emits into DDL and is simply inert on SQLite; `make lint` and all 32 tests pass. Also, `generic:Anthropic` is the *correct* stamped value — the docs' `generic:my-upstream` is the wrong artifact, and I have folded that into the provider-path comment rather than raising it separately.

sh1ftred (dismissed):
The latest update substantially improves this implementation and resolves the concrete correctness and operational issues raised earlier. In particular, provider-scoped visibility, failure-safe persistence, bulk writes, normalized model IDs, query performance, deterministic output, freshness reporting, and the test seams are now much stronger.

I am still requesting changes because the public path identity and response payload do not yet support the stated goal: allowing routstrd to discover and later select a precise configured upstream and, for OpenRouter, a precise backend.

## 1. Make `path` an unambiguous future routing selector

The current values (`anthropic`, `generic`, `openrouter:Anthropic`) identify a provider type or display name, not a uniquely selectable route. In particular, multiple generic upstreams collapse to `generic`.

Please include the configured upstream URL in every path identity and encode its named components using standard percent-encoding. The resulting value can later be returned unchanged in the single planned header:

```http
X-Routstr-Model-Path: <encoded-path>
```

Examples:

```text
url=https%3A%2F%2Fapi.anthropic.com%2Fv1
url=https%3A%2F%2Fcustom-provider.example%2Fv1
url=https%3A%2F%2Fopenrouter.ai%2Fapi%2Fv1&provider=anthropic
url=https%3A%2F%2Fopenrouter.ai%2Fapi%2Fv1&provider=google-vertex%2Fus-east5
```

Conceptually this is `urlencode({"url": upstream.base_url, "provider": endpoint_tag})`, with `provider` omitted for direct paths. Clients should treat the value as opaque: fetch it and send it back unchanged. Named percent-encoded components avoid ambiguous custom delimiters when URLs or endpoint tags contain reserved characters.

Different URLs are different client-visible paths. Multiple internal accounts using the same OpenRouter URL may intentionally collapse, since account selection is a node implementation detail rather than a meaningful client preference.

## 2. Use OpenRouter's routable endpoint `tag`, not `provider_name`

The per-model endpoints response includes both:

```json
{
  "provider_name": "OpenAI",
  "tag": "openai"
}
```

`provider_name` is display text. `tag` is the machine-readable provider/endpoint slug used by OpenRouter routing and can preserve variants such as `google-vertex/us-east5` or `deepinfra/turbo`.

Please retain the exact `tag` in the encoded selector and expose `provider_name` separately for display. Do not reconstruct a routing identifier from the display name.

For the later request-side implementation, selecting a precise OpenRouter path should constrain routing using that tag (for example through `provider.only`). The fallback behavior should also be explicit; silently serving through another backend would normally violate an exact client path selection.

## 3. Return the endpoint metadata already fetched

The OpenRouter per-model endpoint call already returns the information routstrd needs to choose among providers, including:

- pricing;
- status and quantization;
- context and token limits;
- supported parameters and implicit-caching support;
- latency and throughput percentiles;
- uptime metrics;
- `tag` and `provider_name`.

The current implementation downloads this payload and discards everything except `provider_name`. Please preserve and return the useful endpoint-specific data when a client asks for the providers for a model.

The existing path-object shape can evolve without replacing the envelope, for example:

```json
{
  "path": "url=https%3A%2F%2Fopenrouter.ai%2Fapi%2Fv1&provider=openai",
  "upstream_url": "https://openrouter.ai/api/v1",
  "provider": {
    "name": "OpenAI",
    "slug": "openai"
  },
  "status": 0,
  "quantization": "unknown",
  "context_length": 128000,
  "max_prompt_tokens": null,
  "max_completion_tokens": 16384,
  "supported_parameters": ["temperature", "tools", "tool_choice"],
  "supports_implicit_caching": false,
  "performance": {
    "latency_last_30m": null,
    "throughput_last_30m": null,
    "uptime_last_5m": 100,
    "uptime_last_30m": 99.985,
    "uptime_last_1d": 99.98
  },
  "pricing": {}
}
```

The separate OpenRouter providers API can optionally enrich this with headquarters, datacenters, policy URLs, and status-page data by matching the endpoint tag's base slug. The exact per-model endpoint tag and metrics are the primary requirement.

## 4. Return final Routstr-facing pricing per path

Raw OpenRouter pricing is not necessarily the amount the Routstr client will pay. The endpoint should calculate and return the final client-facing path price after applying the node's applicable pricing policy, including provider fee/markup, fixed-pricing behavior, exchange-rate conversion, minimum charges, and the supported pricing dimensions.

Please reuse the same shared pricing logic used by reservation and settlement rather than implementing an approximation in `model_paths.py`. Align the response fields with `/v1/models` where possible so clients do not need incompatible pricing parsers.

This is also important for the later selection flow: once a path is selected, reservation, routing, and settlement must use that selected path and its pricing rather than the default cheapest candidate.

## 5. Persist selectable identity and metadata

The persistence model now needs to represent endpoint identity and its discovery snapshot, not only a response-stamping string. A reasonable shape is typed identity fields plus extensible JSON metadata, e.g.:

```text
model_id
path                    # encoded selector
upstream_url
provider_tag            # nullable
provider_name           # nullable
upstream_provider_id
metadata_json
updated_at
```

Uniqueness should follow the selectable route, such as `(model_id, upstream_url, provider_tag)`, not the human-readable provider name. Distinct routes must not collapse merely because their display labels match.

## 6. Define missing-model behavior

A missing model currently returns `200` with an empty collection. Please choose and document this intentionally. I prefer `404` because it distinguishes an unknown model from a known model that currently has no paths, but an explicitly documented collection-style response is also defensible.

## Scope

It is fine for this PR to remain discovery-only and add request-side routing in a follow-up. However, this PR defines the public identifiers that follow-up will consume, so those identifiers need to be stable, unambiguous, URL-specific, and capable of identifying exact OpenRouter endpoint variants now.

## Resolved in the latest update

For clarity, I consider the earlier implementation findings around provider-scoped overrides/disables, case-insensitive matching, destructive transient failures, empty-upstream deletion, malformed payload handling, forwarded OpenRouter IDs, bulk persistence, redundant indexing, multi-segment IDs, full-table per-model reads, deterministic ordering, freshness timestamps, dynamic refresh settings, migration chaining/formatting, and test coverage to be addressed by the latest commit. I am not requesting further changes on those points here.

9qeklajc:
1. Make path an unambiguous future routing selector

I think we shouldn't be using url here .. it's enough to use a unique path to the current provider ... what if someone is running custom models in one of his servers, we shouldn't be leaking that 

provider id instead of url should be enough

9qeklajc:
3. Return the endpoint metadata already fetched

this is planned for next step in another PR

routstr/core/admin.py
```diff
@@ -51,6 +51,23 @@
 MAX_USAGE_ANALYTICS_HOURS = 365 * 24
 
 
+async def _refresh_provider_model_paths(upstream_provider_id: int) -> None:
```
jeroenubbink, line 54: **Blocking — the OpenRouter fan-out now runs inside admin requests, and ignores its own kill switch.**

This is `await`ed inline in seven handlers: `upsert_provider_model` (596), `delete_provider_model` (651), `delete_all_provider_models` (671), `batch_override_provider_models` (767), and the three provider create/update paths (968, 994, 1030).

Measured on this head, mocked upstream at 50 ms/call, 100 models:

- mutation #1: **1.13 s, 100 OpenRouter calls**
- mutation #2, same provider, nothing changed: **1.10 s, another 100 calls**

`_RefreshCycleState` is constructed inside `refresh_model_paths`, so `endpoint_cache` dedupes *within* a cycle and never *across* mutations. Saving ten models one at a time in the admin UI is ten full fan-outs. Worst case is `400 / 5 × 10 s = 800 s` inside one HTTP request; the `rate_limited` latch bounds a 429 storm but not the timeout case, which is what produces that number. Concurrent mutations multiply the socket usage with no global bound.

The half I would fix regardless of whether you buy the latency argument: **`refresh_model_paths_for_provider` never consults `_refresh_interval_seconds()`**, so the kill switch added for item 10 does not cover this path. Verified with `enable_model_paths_refresh=False` and the interval at `0` — the background loop correctly idles, and the admin path still made 20 OpenRouter calls and wrote 20 models. An operator who has explicitly disabled the feature still spends their API key on every admin mutation.

Making this fire-and-forget (or a dirty flag the existing loop picks up) and gating it on the same enable flag closes both halves.

Scope so nobody over-reads it: non-OpenRouter providers return before any HTTP — verified, 50 models, **0 calls, 36 ms**. This only bites nodes with an OpenRouter upstream.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,700 @@
...
+        {(path.model_id, path.path): path for path in snapshot.paths}.values()
+    )
+    now = int(time.time())
+    async with create_session() as session:
+        delete_stmt = delete(ModelPathRow).where(
+            col(ModelPathRow.upstream_provider_id) == upstream_provider_id
+        )
+        if snapshot.preserve_model_ids:
+            delete_stmt = delete_stmt.where(
+                col(ModelPathRow.model_id).not_in(sorted(snapshot.preserve_model_ids))
```
jeroenubbink, line 502: **Advisory (the one I would most like fixed here) — a partial degradation can freeze a provider's whole path set.**

`preserve_model_ids` is the right shape for item 3, but exposed ids deliberately collapse across vendors, so two models on one provider that collapse to the same public id — `vendora/shared` and `vendorb/shared` → `shared` — plus one of them degrading means the survivor's rows are exempted from this delete *and* re-inserted by the insert below.

Reproduced on this head: `IntegrityError: UNIQUE constraint failed: model_paths.model_id, model_paths.path, model_paths.upstream_provider_id`. The provider's transaction rolls back, the per-provider `except` in `refresh_model_paths` swallows it as `Failed to refresh model paths for provider`, and that provider stops updating entirely until both models succeed in the same cycle. It does self-heal, but only once the degradation clears.

Exempting by `(model_id, path)` rather than by `model_id`, or upserting instead of delete-then-insert, avoids it.

Calling this out as the advisory most worth taking now specifically because it is only reachable through the degradation path — the suite will stay green over it indefinitely.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,700 @@
...
+        "endpoint": endpoint,
+    }
+
+
+async def get_all_model_paths() -> dict:
+    """All models with their exact selectable routes."""
+    async with create_session() as session:
+        rows = (
+            await session.exec(
+                select(ModelPathRow).order_by(
```
jeroenubbink, line 759: **Optional — the listing route still materializes the entire table per request.**

The per-model route is properly fixed (SQL predicate, **77.7 ms → 1.17 ms** measured). This one has no predicate: **72.7 ms for 400 models**, unauthenticated, uncached, on the same event loop as the billing proxy — roughly 14 req/s on one core, from an unauthenticated caller.

Not blocking, since the table is bounded and the ordering/dedupe work here is otherwise good. But it is the unresolved half of item 12, and a cheap cache keyed on `max(updated_at)` would close it now that rows carry a timestamp.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,700 @@
...
+            model_id=exposed_model_id(model),
+            path=encode_model_path(provider_identity.id),
+            provider=provider_identity,
+        )
+
+    if not is_openrouter_base_url(upstream.base_url):
+        return ProviderPathSnapshot(paths=tuple(_base_path(model) for model in models))
+
+    if not (upstream.provider_type or "").strip():
+        return ProviderPathSnapshot(paths=())
```
jeroenubbink, line 423: **Optional — an empty model cache is still treated as authoritative-empty.**

This returns an empty snapshot with an empty `preserve_model_ids`, so `_persist_provider_paths` deletes every row for the provider. Measured: 1 row → 0.

The OpenRouter path above now correctly distinguishes "unknown" from "empty" and preserves prior rows; this path does not, which is the same inversion `preserve_model_ids` was introduced to fix. Low severity because `/v1/models` is also empty in that state, so the two surfaces at least agree — but a transient empty cache should preserve, not wipe.

routstr/upstream/base.py
```diff
@@ -237,6 +237,28 @@ def _fold_cache_into_input_tokens(usage: object) -> None:
             except (TypeError, ValueError):
                 pass
 
+    def discovery_path_for_subprovider(self, sub_provider: str | None) -> str | None:
```
jeroenubbink: **Optional — dead code from the pre-pivot design.**

`discovery_path_for_subprovider` (here) and `discovery_base_paths` (257), plus their overrides in `openrouter.py:21` and `openrouter.py:33`, have no production caller after the move to opaque selectors. Their only reference in the tree is `tests/unit/test_model_paths.py::test_discovery_paths_mirror_response_stamping`.

That makes the test's entire subject dead code, which is coverage moving backwards — it will keep passing no matter what discovery does. Worth deleting all four methods and the test.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,700 @@
...
+    if not upstreams:
+        logger.warning("Skipping model paths refresh: no live upstreams")
+        return
+
+    (
+        overrides_by_key,
+        disabled_model_keys,
+        provider_identities,
+    ) = await _load_model_visibility()
+    await prune_model_paths_for_inactive_providers()
```
jeroenubbink, line 583: **Optional — double prune per mutation, and a periodic write on an idle table.**

An admin mutation runs `refresh_model_maps()` (which now calls `prune_model_paths_for_inactive_providers`) and then `refresh_model_paths`, which prunes again here — two identical `DELETE` statements per mutation.

Separately, `refresh_model_maps_periodically` runs every 60 s, so this now takes a write lock on `model_paths` every minute on the billing DB even when no provider has changed. Cheap on an empty delete, but it is a new periodic writer on the hot file; a guard on whether the enabled-provider set actually changed would remove it.

routstr/upstream/model_paths.py
```diff
@@ -0,0 +1,700 @@
...
+
+@dataclass(frozen=True)
+class ProviderPathSnapshot:
+    """Refresh result plus model IDs whose prior rows must survive degradation."""
+
+    paths: tuple[DiscoveredPath, ...]
+    preserve_model_ids: frozenset[str] = frozenset()
+
+
+def encode_model_path(provider_id: int, endpoint_tag: str | None = None) -> str:
```
jeroenubbink: **Optional, on the selector — I think `slug` settles the open thread, rather than a new objection.**

@sh1ftred asked for an unambiguous route selector and proposed the upstream URL; the objection to leaking a private server's URL is right, and dropping it was the correct call.

But `provider=<id>` is `upstream_providers.id` — an `INTEGER PRIMARY KEY`, i.e. a rowid alias with no `AUTOINCREMENT`. SQLite reuses those after a delete, so an operator who removes a provider and adds another can hand a client's cached selector to a completely different upstream, silently. That is the same rowid-reuse property I raised as item 13 for orphaned rows, surfacing now in the public contract.

`slug` is `unique=True`, admin-settable, already the public provider identity elsewhere in the repo, and leaks nothing about the URL. `provider=<slug>&endpoint=<tag>` gives the stability that was asked for and keeps the property that was asked for.

Everything else about the encoding — percent-encoded named components, opaque to clients, the machine-readable `tag` rather than `provider_name` — reads right as built.

routstr/core/admin.py
```diff
@@ -705,6 +725,9 @@ async def batch_override_provider_models(
                     json.dumps(model_data.alias_ids) if model_data.alias_ids else None
                 )
                 existing_row.enabled = model_data.enabled
+                existing_row.forwarded_model_id = (
```
jeroenubbink: **Withdrawn as a finding — but a scope note, and a main-side issue.**

I had this down as a regression, then checked before posting: `main` already does `existing_row.forwarded_model_id = payload.forwarded_model_id or payload.id` in `upsert_provider_model`, and `main`'s batch handler had no `forwarded_model_id` handling at all. So this makes the two endpoints consistent rather than introducing anything, and the resulting row state is identical either way. Withdrawn.

Two things still worth saying:

**Scope.** It is unrelated to path discovery and lands under a commit titled "make model path identity selectable". More practically, the batch dialog is raw operator-pasted JSON whose own sample template omits `forwarded_model_id`, so one paste now writes it for every model in a provider at once, where previously that took one deliberate save per model.

**The underlying behaviour deserves a fix, just not in this PR.** Defaulting `forwarded_model_id` to the model's own id changes the `unique_key` used in `create_model_mappings`, so a model served by two providers can appear twice in `/v1/models` under two different ids — measured `['deepseek-chat']` becoming `['deepseek-chat', 'deepseek/deepseek-chat']`. Neither endpoint can set the field back to null. That is a `main` bug rather than yours; I have filed it separately so it does not ride on this review.

tests/unit/test_fee_payout_migration.py
```diff
@@ -37,7 +37,9 @@ def test_fresh_node_migrates_fee_payout_schema_to_head(tmp_path: Path) -> None:
             "payout_in_progress_msats, payout_started_at FROM routstr_fees"
         ).fetchone()
 
-    assert version == ("9c4d8e2f1a6b",)
+    # Head of the 7f2843d3f4e4 lineage: model-paths chains onto the fee-payout
+    # repair migration.
+    assert version == ("4e0c3d195a49",)
```
jeroenubbink: **Optional — this pins a branch-specific migration id into an unrelated test.**

Asserting against `ScriptDirectory.get_current_head()` instead means the next migration does not have to edit this file, and the test keeps checking what it actually cares about. Same note you gave me on the pricing-provenance PR, so worth applying in both directions.

jeroenubbink (changes requested):
Re-reviewed on `16fc548b`. The rewrite is substantial and most of it holds up under measurement, so this round is much shorter: **one item I would block on, one advisory, and the rest explicitly optional.** Detail is inline; this is the map.

**Verified fixed** from my previous 15, each re-run rather than read off: provider-scoped disable and override keying (items 1, 2 — the isolation bug that https://github.com/Routstr/routstr-core/pull/589 fixed on the routing side is gone), case-insensitive matching (6), malformed payload handling (5), the `forwarded_model_id` fallback in `openrouter_author_slug` (7), first-slash `public_model_id` (11), chunked persistence (4 — **1011 ms → 26 ms** for 2,000 rows), the redundant `model_id` index, the SQL-filtered per-model route (**77.7 ms → 1.17 ms**), invalidation via `prune_model_paths_for_inactive_providers` (13), no priced `Model` rebuild in the visibility pass (14), and the test seams (15 — the private-helper drives are mostly gone and the periodic loop is now covered).

Three landed better than I asked for: ordering is genuinely deterministic now that the stored `model_id` is the public id, `updated_at` gives consumers real freshness, and the loop re-reads its interval and enable flag every iteration and *idles* rather than exiting when disabled, so `0 → N` works without a restart. Migration is re-chained onto `9c4d8e2f1a6b`, single head, and the repo-wide format commit is reverted.

Items 8 and 9 are dissolved by the opaque-selector pivot rather than patched, which is the better outcome.

**Blocking:** one item, on `admin.py` — the discovery fan-out now runs inside admin requests and ignores its own kill switch. Comment inline.

**Advisory:** a partial degradation can freeze a provider's whole path set. Reproduced, inline. Not blocking, but it is the one I would most like to see taken here rather than later.

**Explicitly optional — take them, defer them or decline them, none of it affects my sign-off:** the unauthenticated full-table read on `GET /v1/models/paths`, the empty-model-cache wipe, two dead methods left over from the pivot, the double prune per mutation, the pinned migration head literal, and a suggestion on the selector value. All inline and marked as optional.

**On the selector** (optional), I think `slug` settles the open thread between @sh1ftred's stability requirement and the objection to leaking upstream URLs — inline on `encode_model_path`.

**One withdrawal.** I had the `forwarded_model_id` line in `batch_override_provider_models` down as a regression, then checked before posting: `main` already does exactly that in `upsert_provider_model`, and `main`'s batch handler had no handling at all. It makes the two endpoints consistent rather than introducing anything, so I withdraw it as a finding. What remains is scope, and a main-side issue I have opened separately — inline.

@sh1ftred — your "Resolved in the latest update" list is accurate for the head it was written against, I agree with the direction and with most of it, and I have not re-opened any of it. Flagging only because it bears on this PR: that review landed at 16:01 and https://github.com/Routstr/routstr-core/pull/588/commits/16fc548b48d982211f4f4a8afde63a6a36f6da9c landed at 18:12, rewriting the persistence shape, the public identity and the response payload, and adding all of the `admin.py` wiring. The one item I am blocking on sits entirely inside that window — nothing wrong with the assessment, the commit just moved underneath it.

Requesting changes on the admin-path item only. Happy to re-review on that alone.

routstr/core/admin.py
jeroenubbink: Filed as https://github.com/Routstr/routstr-core/issues/639 — nothing for this PR to do about it.

sh1ftred:
> 1. Make path an unambiguous future routing selector
> 
> 
> I think we shouldn't be using url here .. it's enough to use a unique path to the current provider ... what if someone is running custom models in one of his servers, we shouldn't be leaking that
> 
> provider id instead of url should be enough

No, URL is necessary. If you're running a custom model yourself, you simply say that. It's crucial to know which upstream is serving the models. Pls expand on your reasoning here if you're really against it.

9qeklajc:
> > 1. Make path an unambiguous future routing selector
> > 
> > I think we shouldn't be using url here .. it's enough to use a unique path to the current provider ... what if someone is running custom models in one of his servers, we shouldn't be leaking that
> > provider id instead of url should be enough
> 
> No, URL is necessary. If you're running a custom model yourself, you simply say that. It's crucial to know which upstream is serving the models. Pls expand on your reasoning here if you're really against it.

u will never know which upstream I'm running if It's a VPS IP Address for example where I'm hosting ollama, u don't want that information to be public

sh1ftred:
Following up on my earlier review after the `16fc548` rewrite.

## The upstream URL should be in the path

I initially conceded that omitting the URL was the right call. I was wrong, and I want to correct that.

Users do not trust Routstr nodes yet. A client sending paid inference requests has no way to verify where their data is actually going. Including the configured upstream URL in the path selector is a transparency feature, not a privacy leak — it lets the client see and choose which upstream URL their request will route through. That is a big plus for trust, and it outweighs the operator's preference to keep the URL private.

The current `provider=<id>` design is opaque in the wrong direction: the client sees an integer that identifies nothing, cannot distinguish between upstreams by URL, and — as @jeroenubbink noted — is unstable because SQLite reuses integer primary keys after a delete.

I'd ask for the URL back in the selector, combined with the endpoint tag for OpenRouter:

```text
url=https%3A%2F%2Fapi.anthropic.com%2Fv1
url=https%3A%2F%2Fopenrouter.ai%2Fapi%2Fv1&endpoint=openai
url=https%3A%2F%2Fopenrouter.ai%2Fapi%2Fv1&endpoint=google-vertex%2Fus-east5
```

If the `id`-vs-`slug` stability concern is also in play, the URL itself already solves it — a URL is stable, human-readable, and survives delete/recreate cycles. The `provider` field can remain as display metadata (`slug`, `type`) but should not be the primary identity.

## Endpoint metadata and pricing — still open

Two items from my earlier review remain unaddressed in this commit:

1. **Full endpoint metadata**: The OpenRouter per-model endpoint API returns pricing, status, quantization, context length, token limits, supported parameters, caching support, latency, throughput, and uptime. The current code fetches all of this and keeps only `tag` and `provider_name`. Please preserve and return the useful fields — that is the entire point of asking "which providers serve this model?"

2. **Final Routstr-facing pricing per path**: Raw OpenRouter pricing is not what the client pays. Please calculate and return the final client-facing price per path using the same shared pricing logic as reservation and settlement, after applying provider fee, node markup, exchange-rate conversion, and minimum charges.

Both of these require either typed columns or a `metadata_json` column on `ModelPathRow` to persist the endpoint snapshot.

9qeklajc:
> Endpoint metadata and pricing — still open

is going to be in part two of this issue

9qeklajc:
> The upstream URL should be in the path

> The current provider=<id> design is opaque in the wrong direction: the client sees an integer that identifies nothing, 
> cannot distinguish between upstreams by URL, and — as @jeroenubbink noted — is unstable because SQLite reuses 
> integer primary keys after a delete.

we can provide a url but not sure how is that unique when multiple provider with the same url are configured ?

I think there still some confusion here:

1. url is not enough, provider can configure multiple upstream with the same url (we have that already)
2. id is unique for the current state, and true when delete and recreate the id would be reused but that's an easy fix, a migration should fix that
3. as mentioned above when we use something like this `X-ROUTSTR-MODEL-PATH` to address a specific upstream and model url isn't correct, using a provider id and a model id would be the proper approach long term

sh1ftred:
> > The upstream URL should be in the path
> 
> > The current provider= design is opaque in the wrong direction: the client sees an integer that identifies nothing,
> > cannot distinguish between upstreams by URL, and — as @jeroenubbink noted — is unstable because SQLite reuses
> > integer primary keys after a delete.
> 
> we can provide a url but not sure how is that unique when multiple provider with the same url are configured ?
> 
> I think there still some confusion here:
> 
>     1. url is not enough, provider can configure multiple upstream with the same url (we have that already)
> 
>     2. id is unique for the current state, and true when delete and recreate the id would be reused but that's an easy fix, a migration should fix that
> 
>     3. as mentioned above when we use something like this `X-ROUTSTR-MODEL-PATH` to address a specific upstream and model url isn't correct, using a provider id and a model id would be the proper approach long term

The client doesn't care and doesn't need to know about multiple API keys for the same upstream. My LLM also brought this up as an issue haha.

PR #591, opened 2026-07-07
fix(wallet): return 425 when refund is pending, not 404

sh1ftred:
## Summary

Fixes the refund race condition tracked as `refund-race-condition`.

The X-Cashu refund endpoint (`POST /v1/wallet/refund`) raised `404 "Refund not found"` when the `in` transaction existed with a `request_id` but the `out` (refund) transaction had not been written yet. This is a timing race: the endpoint is polled while the upstream request is still in flight, before `send_refund()` has minted and stored the refund token.

The `404` was indistinguishable from a genuinely-missing refund, so clients had no signal that retrying would succeed — leading to stranded refunds when clients gave up after the first `404`.

## Changes

- **`routstr/balance.py`** — `refund_wallet_endpoint`:
  - The third not-found branch (`in` row exists with `request_id`, but `out` row missing) now returns **`425 Too Early`** with `Retry-After: 2` instead of `404`. This signals "refund is pending, retry shortly."
  - The two earlier `404` branches (no `in` row; `in` row has no `request_id`) remain `404` — those genuinely mean no refund will ever exist.
  - Adds `logger.debug` on all three not-found/pending branches so the race is no longer invisible to operators (middleware does not log request headers; the `404` path previously wrote nothing).

- **`tests/unit/test_balance.py`**:
  - `test_refund_x_cashu_pending_raises_425` — new test for the race window (in row exists, out row not yet created → 425 + `Retry-After`).
  - `test_refund_x_cashu_in_tx_without_request_id_raises_425` — new test confirming the no-`request_id` case stays a genuine `404`.

## Why 425

- Semantically correct (RFC 9110 §15.6.4: "server is unwilling to process the request because the resource ... is not yet available").
- Carries an explicit `Retry-After` so well-behaved clients back off deterministically.
- Distinct from `404` (genuine "no refund will ever exist") so client logic can branch.

## Test plan

- [x] `uv run pytest tests/unit/test_balance.py` — 16 passed
- [x] `uv run pytest tests/unit/` — 437 passed, 1 skipped
- [x] `uv run pytest tests/integration/test_wallet_refund.py` — 14 passed, 1 skipped

## Out of scope (tracked for follow-up)

These were identified during research and are **not** addressed here to keep this PR minimal:
- Stop swallowing `store_cashu_transaction` failures in `send_refund` and the `in`-row write (`except Exception: pass`) — currently a transient DB blip can permanently strand a refund.
- Index `(request_id, type)` on `cashu_transactions` (the `out` lookup is unindexed).
- Restore request-header logging in middleware (at least `X-Cashu`).
- Client-side timeout guidance (`forward_x_cashu_request` uses `httpx.AsyncClient(timeout=None)`).
- Unify the duplicated `out`-row write across call sites (refs `unify-refund-flows`).

Context event with full codebase findings: `fa32e0d921789e5ae217c7863333f067166cd937565b9191904b0925602a1a30` on Nostr (kind 3638, linked to task `refund-race-condition`).

### DIFF
--- a/routstr/balance.py
+++ b/routstr/balance.py
@@ -274,7 +274,20 @@ async def refund_wallet_endpoint(
         )
         out_tx = out_tx_result.first()
         if out_tx is None:
-            raise HTTPException(status_code=404, detail="Refund not found")
+            # The "in" row exists with a request_id, but the "out" (refund)
+            # row hasn't been written yet — the upstream request is still in
+            # flight and the refund will be minted once it completes. Tell the
+            # client to retry instead of 404ing permanently (race condition
+            # where /v1/wallet/refund is polled before the refund exists).
+            logger.debug(
+                "refund_wallet_endpoint: refund pending (in row exists, out row not yet created)",
+                extra={"request_id": in_tx.request_id},
+            )
+            raise HTTPException(
+                status_code=425,
+                detail="Refund is pending; retry shortly.",
+                headers={"Retry-After": "2"},
+            )
         if out_tx.swept:
             raise HTTPException(status_code=410, detail="Refund has been swept")
 
--- a/tests/unit/test_balance.py
+++ b/tests/unit/test_balance.py
@@ -104,6 +104,64 @@ async def test_refund_x_cashu_not_found_raises_404() -> None:
 
 
 @pytest.mark.asyncio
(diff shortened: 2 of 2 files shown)


9qeklajc:
LGTM

PR #596, opened 2026-07-09
enforce out token creation

9qeklajc:


9qeklajc:
this could be needed if after this https://github.com/Routstr/routstr-core/pull/600 we still have issues

PR #597, opened 2026-07-10
fix: per-mint rate limiting, trusted-mint fallback, and retry factory…

9qeklajc:


routstr/payment/helpers.py
```diff
@@ -18,6 +18,17 @@
...
 
+# Interim policy: when Routstr must move value to another trusted mint, the
+# cross-mint Lightning round trip can consume fees that are not visible to the
+# client. Reserve 5% headroom until the fee-payer policy is made explicit.
+_MINT_FEE_ALLOWANCE = 0.05
+
+
+def apply_mint_fee_allowance(cost_msat: int) -> int:
+    """Reserve headroom for possible trusted-mint fallback fees."""
+    adjusted = math.ceil(cost_msat * (1 - _MINT_FEE_ALLOWANCE))
```
jeroenubbink: `ceil(cost * 0.95)` reduces the reservation; headroom would be `* 1.05`. This replaces the `max(max_cost_for_model, min_request_msat)` clamp at `proxy.py:335` and the failover branch `:471`, so every request on every payment method now reserves 95% of the documented maximum, fallback or not:

- bearer keys: admitted at `balance >= 95% of max`, and nothing clamps `total_msats` in `auth.py`, so a request that actually costs the max drives the balance negative;
- X-Cashu: `check_token_balance` accepts a token 5% short, and settlement's `if refund_amount > 0` silently drops the shortfall — the node absorbs it.

No test covers a request whose actual cost reaches the model max, which is the case this breaks.

Please split it into its own PR with an explicit fee-payer statement; if it stays here, apply it only on the path that can actually incur a cross-mint fee.

routstr/wallet.py
```diff
@@ -591,20 +1281,35 @@ async def swap_to_primary_mint(
...
+                    proofs=token_obj.proofs,
+                    invoice=mint_quote.request,
+                    fee_reserve_sat=melt_quote.fee_reserve,
+                    quote_id=melt_quote.quote,
+                ),
+                op_name="swap_melt",
+                mint_url=token_obj.mint,
+                retry_timeouts=False,
             )
         except Exception as e:
```
jeroenubbink, line 1160: `_mint_operation` wraps each attempt in `asyncio.wait_for(..., 30)`, and `wait_for` **cancels** the in-flight request. A cancelled melt may well have settled, so `httpx.TimeoutException` landing here → `SourceMintConnectionError` → 503 *"the token cannot be redeemed at another mint"* tells the client their proofs are intact while the value sits at the destination as a paid-but-unminted quote. `TokenConsumedError` already models exactly that state.

30s is also below the Lightning attempt window the mint itself works to: CLN `pay` retries for 60s by default and LND's attempt window is the same. And no timeout bounds the *ambiguity* — an in-flight HTLC can stay pending to its CLTV expiry, which is why NUT-05 has `UNPAID → PENDING → PAID`. Meanwhile nginx cuts at 60s (`proxy_read_timeout`) and Cloudflare at 100s (524, not raisable below Enterprise), so a synchronous client request can never wait a melt out. Any number picked here is wrong.

Suggestion: stop inferring the outcome from the exception type and ask the mint, with two calls that already exist — `get_melt_quote(quote_id)` (NUT-05 state) and `check_proof_state(proofs)` (NUT-07, already used at `wallet.py:1643`). `UNSPENT` is the only state that may say "retry with another token"; `SPENT`/`PAID` → `TokenConsumedError` and carry on minting at the destination; `PENDING` deserves its own outcome (durable record + reconciler, client sees "in progress", never "unspent"). Persisting the melt quote id and source mint before the melt is the same pattern this PR introduces with `lightning_invoices.mint_url`. Then the timeout value stops being load-bearing.

Refs: https://docs.corelightning.org/reference/pay , https://github.com/lightningnetwork/lnd/discussions/8324

Smaller version of the same thing: `lnurl.py`'s inner 30s now fires before its outer `MELT_TIMEOUT_SECONDS = 60`, so that path reports "timed out after 60s" at 30s.

routstr/wallet.py
```diff
@@ -123,6 +155,311 @@ class TokenConsumedError(Exception):
...
+        while True:
+            deadline = self._cooldown_until
+            wait = max(0.0, deadline - time.monotonic())
+            if wait <= 0:
+                return
+            logger.debug(
+                "Mint rate guard: cooling down",
+                extra={"mint_url": self._mint_url, "wait_seconds": round(wait, 2)},
+            )
+            await asyncio.sleep(wait)
```
jeroenubbink: This sleep can be up to `_MINT_RATE_LIMIT_MAX_COOLDOWN_SECONDS` = 7h (line 147, reached after ~9 consecutive 429s), and it is taken **while holding the node-wide wallet lock**.

`wallet_operation_guard()` (line 49) is an exclusive `fcntl.flock` serialising proof mutation across *all* worker processes, acquired by a `sleep(0.05)` spin with no timeout. `credit_balance` (X-Cashu header and `/v1/wallet/topup`) holds it and reaches `_redeem_same_mint` → `_mint_operation(load_mint, ...)` at line 608 with no cooldown pre-check. So one mint answering 429 nine times stops every payment operation on the node, in every worker process, for up to 7 hours, with the waiters spinning at 20 Hz. `validate_bearer_key` (`auth.py:179`) and `check_invoice_payment` (`lightning.py:356`) hold the same lock.

`retry_on_rate_limit=False` doesn't avoid it — that only suppresses in-place retry. Nor does adding a pre-check at the call site: the `flock` is acquired *before* the guarded code is reached.

Suggestion: `run()` raises `MintRateLimitedError` with `retry_after = cooldown_remaining()` instead of entering `_run_probe`'s wait — the guard never sleeps. Probing belongs to something that already loops (the invoice watcher runs every 10s) or to one small background prober. That also lets `classify_redemption_error`'s existing `mint_rate_limited` 503 finally surface with a real `Retry-After`, and removes the `_probe_lock` convoy.

I checked every Lightning/ecash path against a raise here and fail-fast is safe on all of them: same-mint redeem and the cross-mint swap both refuse before anything is sent; `create_invoice` and the admin balance endpoints already pre-check; `check_invoice_payment` swallows the exception and leaves the invoice `pending`, where the new CAS fence keeps the retry exactly-once; `periodic_payout` retries next iteration. Two things need to come along with it:

- `get_invoice_status` still flips a pending invoice to `"expired"` past `expires_at`. Fail-fast makes "we couldn't reach the mint" a normal outcome, so a user who really paid gets told their invoice expired.
- `select_to_send(set_reserved=True)` at `wallet.py:718`, `wallet.py:2287` and `lnurl.py:234` bypasses the guard entirely. cashu falls through to `swap_to_send` → `POST /v1/swap` whenever offline coin selection can't hit the exact amount, i.e. the normal case — so the send/refund/payout path keeps hitting a mint already known to be rate-limited, and a 429 there never feeds back, since only `_mint_operation` calls `apply_rate_limit_cooldown`. Bounded at cashu's 60s client default so it isn't a hang, but it wants one choke point.

Separately: the retry ladder is 4 attempts × 30s plus `1+2+4` backoff ≈ 127s per operation against a merely slow mint, and a cross-mint redeem performs five in sequence — past nginx's 60s with no cooldown involved at all. Worth an overall cap.

routstr/wallet.py
```diff
@@ -112,8 +116,36 @@ class TokenConsumedError(Exception):
...
+    """Cashu wallet adapter that preserves rate-limit status information.
+
+    Cashu's default response adapter converts JSON error bodies into plain
+    ``Exception`` instances before calling ``raise_for_status``. Intercept 429
+    here so Routstr's fallback and cooldown policy can use the real status
+    without unreliable message matching.
+    """
+
+    @staticmethod
+    def raise_on_error_request(resp: httpx.Response) -> None:
```
jeroenubbink, line 120: The whole feature rides on this hook being reached. It is today — cashu's `LedgerAPI` calls `self.raise_on_error_request`, and `raise_on_unsupported_version` delegates to it for non-404 — but both tests call it directly, so nothing proves the dispatch. A cashu bump that stops calling it disables 429 detection, cooldowns and fallback **silently**: every 429 becomes a plain `Exception`. One `httpx.MockTransport` test driving a 429 through `wallet.mint_quote` would pin it.

routstr/lightning.py
```diff
@@ -12,12 +15,59 @@
...
+    wallet_operation_guard,
+)
 
 logger = get_logger(__name__)
 
 lightning_router = APIRouter(prefix="/lightning")
 
+# Avoid duplicate work within one process. Cross-process credit fencing is done
+# by the conditional pending -> paid update in _finalize_invoice_settlement().
+_invoice_settlement_locks: dict[str, asyncio.Lock] = {}
```
jeroenubbink: Never pruned: one `asyncio.Lock` per invoice id ever seen by the process, `setdefault` only, nothing removes entries — while the invoices themselves stop mattering once paid or expired. Growth is monotonic with traffic and the only remedy is a restart.

Popping the entry on terminal state is safe: the new CAS fence, not this lock, is what makes settlement exactly-once across processes, and `_mint_invoice_quote` is independently idempotent.

routstr/lightning.py
```diff
@@ -223,158 +352,180 @@ async def recover_invoice(
...
+    lock = _invoice_settlement_locks.setdefault(invoice.id, asyncio.Lock())
+    async with lock, wallet_operation_guard():
+        minted = False
+        try:
+            # Snapshot the row and end the caller's read transaction before any
+            # potentially slow mint I/O. All final DB mutations use owned,
+            # short-lived sessions below.
+            await session.refresh(invoice)
+            if invoice.status != "pending":
+                await session.commit()
```
jeroenubbink, line 451: Still commits a session it borrows from the caller (also `:367`). The rollback half is fixed at this head, and `session.refresh(invoice)` + `_reload_invoice_view` repair the watcher's per-iteration expiry — but `_reload_invoice_view` is itself wrapped in `except Exception: pass` (`:447`), so when it fails (DB down, which is the reason you're in that branch) `get_invoice_status` reads `invoice.expires_at` on an expired instance → `MissingGreenlet` → 500.

This same function was rewritten in #634 to own its session and return the outcome instead of mutating the caller's ORM object; worth taking that shape here rather than solving it twice.

routstr/core/settings.py
```diff
@@ -53,6 +53,18 @@ def parse_env_var(cls, field_name: str, raw_value: str) -> Any:  # type: ignore[
     payout_interval_seconds: int = Field(
         default=900, gt=0, env="PAYOUT_INTERVAL_SECONDS"
     )
+    # Timeout (seconds) for individual mint API operations (melt, mint, swap,
+    # checkstate). When a mint is slow or rate-limiting, operations are
+    # cancelled after this delay instead of hanging indefinitely.
+    mint_operation_timeout_seconds: int = Field(
```
jeroenubbink, line 59: `MINT_OPERATION_TIMEOUT_SECONDS`, `MINT_MAX_CONCURRENCY` and `MINT_RETRY_MAX_ATTEMPTS` are operator-facing knobs on the money path but appear nowhere in `.env.example` or the docs. Whatever survives the melt-timeout discussion should be documented.

routstr/wallet.py
```diff
@@ -643,15 +1348,25 @@ async def swap_to_primary_mint(
 
         break
 
-    logger.info(
-        "swap_to_primary_mint: melt succeeded, minting on primary",
-        extra={"minted_amount": minted_amount, "mint_quote_id": mint_quote.quote},
+    logger.warning(
+        "Source melt succeeded; minting on destination",
```
jeroenubbink, line 1223: Happy-path events logged at `warning` — 26 in this file, including "top-up started/completed", "Trying same-mint Cashu redemption" and this one. At the default `LOG_LEVEL=INFO` every paid request emits several WARNINGs, which buries the real ones. `info` for these; keep `warning` for actual degradations.

routstr/lightning.py
```diff
@@ -12,12 +15,59 @@
 from .core.db import ApiKey, LightningInvoice, create_session, get_session
 from .core.logging import get_logger
 from .core.settings import settings
-from .wallet import get_wallet, wallet_operation_guard
+from .wallet import (
```
jeroenubbink, line 25: `_is_mint_rate_limited`, `_mint_cooldown_remaining` and `_mint_operation` are private names imported across modules, and `lnurl.py:225` imports `_mint_operation` *inside a function* to dodge a circular import. That circularity is the signal: the guard wants its own module with a public API — which is also where the cooldown rewrite naturally lands.

Minor, same area:

- two different functions named `_request_mint_with_fallback` (this one returns `(bolt11, quote, mint_url)`, `wallet.py`'s returns `(wallet, mint_url, quote)`);
- `op_name="swap_mint_on_primary"` and the `swap_to_primary_mint:` log strings survive on a path where the destination is explicitly *not* primary — misleading in exactly the incident you'd read them for;
- `set_reserved_for_send` is a local wallet-DB write but runs through `_mint_operation` (`wallet.py:1655`), so it queues behind mint concurrency and can wait on a mint cooldown for nothing.

routstr/wallet.py
```diff
@@ -343,6 +734,61 @@ async def send_token(amount: int, unit: str, mint_url: str | None = None) -> str
...
+                balances[mint_url] = 0
+                continue
+            raise
+
+        proofs = get_proofs_per_mint_and_unit(wallet, mint_url, unit, not_reserved=True)
+        balances[mint_url] = sum(proof.amount for proof in proofs)
+        if balances[mint_url] >= amount:
+            return mint_url
+
+    raise ValueError(
```
jeroenubbink, line 559: `send()` now routes through here, and this requires a **single** mint to cover the amount — so a node with funds split across two trusted mints fails before any attempt, where the old path would have worked. Easy to hit; I'd call that a regression rather than a question.

Two behaviour changes I'd like confirmed as intended:

- `preferred_mint` is skipped when it isn't in `cashu_mints`, and auto-topup passes the provider row's `mint_url` — so a provider configured against a non-trusted mint now gets topped up with a token from a different mint (the new `token_mint_url` suggests you're aware);
- `get_wallet` throttles reloads to once per 30s (`_WALLOAD_RELOAD_MIN_INTERVAL_SECONDS`), so `fetch_all_balances` and anything else reading proofs through it can report a balance up to 30s stale.

tests/unit/test_wallet.py
```diff
@@ -1389,3 +1541,887 @@ async def test_swap_melt_transport_error_raises_mint_connection_error() -> None:
...
+        assert code != "cashu_mint_rate_limited"
+
+
+# ---------------------------------------------------------------------------
+# _MintRateGuard — probe does NOT escalate cooldown counter
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.asyncio
+async def test_probe_does_not_escalate_consecutive_rate_limits() -> None:
```
jeroenubbink: This and `test_probe_recovery_resets_consecutive_rate_limits` never call `_run_probe` — they hand-simulate what it does and reassign its internals, so they'd keep passing if `_run_probe` regressed. Same shape in `test_non_pending_invoice_is_not_minted` (`tests/unit/test_lightning_settlement.py:148`), which passes a bare `AsyncMock()` session, so the `session.refresh` assertion is vacuous.

jeroenubbink (changes requested):
Three blockers, all on the money path, commented inline. The rest is non-blocking.

1. `apply_mint_fee_allowance` reduces the upfront reservation by 5% on every request, not just on the fallback path.
2. A timed-out melt is classified as "token unspent, retry with another one".
3. The cooldown sleep is unbounded (7h) and is taken while holding the node-wide wallet `flock`.

Verified at `443c910`: `make lint` clean (ruff + mypy), `make test-fast` 1169 passed / 10 skipped.

PR #613, opened 2026-07-14
fix overflow

9qeklajc:


jeroenubbink (commented):
The change itself is correct — `dvh` is the right fix for mobile viewport overflow in dialogs. `vh` ignores browser chrome (address bar), so `90vh` extends behind it on mobile; `90dvh` respects the currently visible area.

But this same bug exists in **11 other places** across `ui/`. They should all get the same treatment while we are here.

### Dialog `max-h-[Xvh]` — same bug, same fix

| File | Current | Should be |
|------|---------|-----------|
| `components/ui/dialog.tsx:133` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/cost-calculator-dialog.tsx:28` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/edit-group-form.tsx:103` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/provider-card.tsx:216` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/add-provider-model-dialog.tsx:441` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/add-model-form.tsx:89` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/providers/ProviderBalance.tsx:115` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/providers/RoutstrProviderCard.tsx:242` | `max-h-[90vh]` | `max-h-[90dvh]` |
| `components/collect-models-dialog.tsx:180` | `max-h-[80vh]` | `max-h-[80dvh]` |

### Base drawer component

| File | Current | Should be |
|------|---------|-----------|
| `components/ui/drawer.tsx:59` | `max-h-[80vh]` (×2: bottom + top) | `max-h-[80dvh]` |

### Chart height (`100vh` → `100dvh`)

| File | Current |
|------|---------|
| `components/usage-metrics-chart.tsx:294` | `h-[calc(100vh-220px)] … sm:h-[calc(100vh-260px)]` |
| `components/top-models-usage-chart.tsx:629` | same pattern |

### Already correct (no action needed)

- `app/logs/log-details-dialog.tsx:70` — uses `svh` for mobile (`70svh`), which already handles dynamic chrome. Desktop `sm:h-[75vh]` is fine (no address bar issue).

### Why fixing just `dialog.tsx` is not enough

The base `DialogContent` merges with `cn(...base, className)` — the prop wins the conflict. If the base fixes `max-h-[90dvh]` but a consumer passes `className='max-h-[90vh]…'`, tailwind-merge keeps the **last** conflicting class → the override wins and the bug stays.

---

Two options:
1. **Scope-expand this PR** to include all 13 fixes — one search-and-replace pass through `ui/`
2. **Merge this as-is** and we file a follow-up issue

Small preference for option 1 since the grep is deterministic and the changes are mechanical. What do you think?

jeroenubbink:
I think fixing all vh's is a simple win but it lgtm like it is.

jeroenubbink (approved):
LGTM!

PR #617, opened 2026-07-16
feat: pricing provenance — track where each model's price came from

jeroenubbink:
## What

Makes a price's **origin** a first-class, queryable fact and uses it to handle free/untrusted prices safely.

Every model now carries a `pricing_source` — `native`, `litellm`, `openrouter`, `manual`, or `unresolved` — in decreasing order of trust, plus a `pricing_checked_at` timestamp and, for the static litellm cost map, its dist version as a freshness anchor. The tag is stamped at each provider's fetch site and survives the fee/sats carrier rebuilds that `refresh` performs.

## Why

Today an unpriceable or free-at-$0 model can end up billing every request at nothing, and there's no way to tell a deliberately-free model from one no source could price. Provenance gives us the signal to fail closed on the latter while letting an operator vouch for the former.

## Behaviour

**Resolution on write (admin):**
- Editing a price flips the row to `manual` — the operator owns it.
- An unchanged save adopts the payload's provenance (a "save as fetched") or preserves the row's.
- A hand-created model is `manual` when priced, `unresolved` when free — a client that omits the tag can't tell a deliberate free import from an unpriced one, so it is never laundered into a billable `manual` $0.

**Money safety:**
- A zero-price row whose source is **not** `manual` is **force-disabled** on write rather than billed at nothing.
- A `manual` price — including an operator zeroing a rate — keeps exactly the `enabled` state requested.
- Price comparison tolerates numeric strings, so a string-typed edit still flips to `manual`.
- An unknown `pricing_source` is rejected (422) instead of persisted as a tag that reads back as `None`.

**Migration:** three nullable columns on `models`; existing rows backfill to `unresolved` (their true origin is unknowable, and a relabelled trusted tag would slip past the re-enable guard). `pricing_source` is plain text so a future source needs no schema change.

## Notes / follow-ups (out of scope here)
- The admin UI doesn't yet round-trip `pricing_source`, so today it relies on price-edit detection to infer `manual`. A follow-up will make the source client-authoritative and add an explicit "this price is manual" checkbox, which also lets an operator keep a genuinely-free model enabled without editing the price.
- `ppqai` deep-copies the matched OpenRouter model before overlaying its price (two PPQ ids can tail-match one OR entry). The broader model-identity consolidation (keeping `vendor/model` separated for pricing and collapsing to one user-facing entry late) is tracked separately.

## Testing
`make lint` clean · `make test-fast` 810 passed. New unit + integration coverage for the source tag through every provider, the migration backfill, and the admin persist rules (price-edit-only `manual`, false-flip regression, zero-price force-disable, string-typed edits, invalid-source rejection, PPQ tail-match).

### DIFF
--- a/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
+++ b/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
@@ -40,16 +40,6 @@ def upgrade() -> None:
             "models",
             sa.Column("pricing_source", sa.String(), nullable=True),
         )
-    if "pricing_checked_at" not in columns:
-        op.add_column(
-            "models",
-            sa.Column("pricing_checked_at", sa.Integer(), nullable=True),
-        )
-    if "pricing_source_version" not in columns:
-        op.add_column(
-            "models",
-            sa.Column("pricing_source_version", sa.String(), nullable=True),
-        )
 
     _backfill_pricing_source(conn)
 
@@ -59,10 +49,5 @@ def downgrade() -> None:
     inspector = sa.inspect(conn)
     columns = {c["name"] for c in inspector.get_columns("models")}
 
-    for column in (
-        "pricing_source_version",
-        "pricing_checked_at",
-        "pricing_source",
-    ):
-        if column in columns:
-            op.drop_column("models", column)
+    if "pricing_source" in columns:
+        op.drop_column("models", "pricing_source")
--- a/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
+++ b/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
@@ -8,6 +8,7 @@ Create Date: 2026-07-16 00:00:00.000000
 from __future__ import annotations
 
 import json
+import math
 
 import sqlalchemy as sa
 from alembic import op
@@ -46,10 +47,19 @@ def _backfill_pricing_source(conn: sa.Connection) -> None:
 
 
 def _row_is_chargeable(pricing_json: object) -> bool:
-    """True if the stored pricing JSON has any positive billable rate.
+    """True if every stored billable rate is usable and at least one is positive.
 
-    Unparseable / missing pricing fails closed (treated as unchargeable), the
-    safe direction for a money guard.
+    The frozen counterpart of ``routstr.payment.models.has_chargeable_price``,
+    and it must answer identically: a row this migration leaves enabled is one
+    the running node will bill on. Returning ``True`` on the first positive rate
+    is what let a malformed row through — a positive ``completion`` hid a
+    negative ``prompt``, and ``Infinity`` (which ``json.loads`` accepts as a bare
+    literal) is greater than zero, so both survived the very sweep meant to
+    fail-close them.
+
+    Every field is therefore checked before any of them can vouch for the row.
+    Unparseable, missing or non-numeric pricing fails closed — the safe
+    direction for a money guard.
     """
     if not isinstance(pricing_json, str):
         return False
@@ -59,13 +69,18 @@ def _row_is_chargeable(pricing_json: object) -> bool:
         return False
     if not isinstance(pricing, dict):
         return False
+    has_positive = False
     for field in _BILLABLE_FIELDS:
         try:
-            if float(pricing.get(field, 0) or 0) > 0:
-                return True
-        except (TypeError, ValueError):
-            continue
-    return False
+            # An oversized JSON integer raises OverflowError, not ValueError.
+            rate = float(pricing.get(field, 0) or 0)
+        except (TypeError, ValueError, OverflowError):
+            return False
+        if not math.isfinite(rate) or rate < 0:
+            return False
+        if rate > 0:
+            has_positive = True
+    return has_positive
 
 
 def _disable_unchargeable_enabled_rows(conn: sa.Connection) -> None:
--- a/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
+++ b/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
@@ -1,4 +1,4 @@
-"""add pricing provenance columns to models
+"""add pricing provenance to models and fail-close unchargeable rows
 
 Revision ID: f0e1d2c3b4a5
 Revises: 9c4d8e2f1a6b
@@ -6,6 +6,8 @@ Create Date: 2026-07-16 00:00:00.000000
 """
 
 from __future__ import annotations
+
+import json
 
 import sqlalchemy as sa
 from alembic import op
@@ -15,18 +17,72 @@ down_revision = "9c4d8e2f1a6b"
 branch_labels = None
 depends_on = None
 
+# Frozen snapshot of routstr.payment.models.BILLABLE_PRICING_FIELDS — the rates
+# whose positive value makes a request chargeable. Copied (not imported) so the
+# migration stays stable if the app's field set later changes.
+_BILLABLE_FIELDS = (
+    "prompt",
+    "completion",
+    "request",
+    "image",
+    "web_search",
+    "internal_reasoning",
+    "input_cache_read",
+    "input_cache_write",
+)
+
 
 def _backfill_pricing_source(conn: sa.Connection) -> None:
     # Existing rows predate provenance tracking, so their true origin is
     # unknowable. Backfill "unresolved" rather than assert a trusted source:
     # a persisted fail-closed zero-price row relabelled "openrouter" would slip
-    # past the re-enable guard. Freshness columns stay NULL — we never resolved
-    # these prices ourselves.
+    # past the re-enable guard.
     conn.execute(
         sa.text(
             "UPDATE models SET pricing_source = 'unresolved' "
             "WHERE pricing_source IS NULL"
         )
+    )
+
+
+def _row_is_chargeable(pricing_json: object) -> bool:
+    """True if the stored pricing JSON has any positive billable rate.
+
+    Unparseable / missing pricing fails closed (treated as unchargeable), the
+    safe direction for a money guard.
+    """
+    if not isinstance(pricing_json, str):
+        return False
+    try:
+        pricing = json.loads(pricing_json)
+    except (TypeError, ValueError):
+        return False
+    if not isinstance(pricing, dict):
+        return False
+    for field in _BILLABLE_FIELDS:
+        try:
+            if float(pricing.get(field, 0) or 0) > 0:
+                return True
+        except (TypeError, ValueError):
+            continue
+    return False
+
+
+def _disable_unchargeable_enabled_rows(conn: sa.Connection) -> None:
+    # The legacy rows that motivated provenance: an enabled model priced at
+    # nothing bills every request at zero. list_models() gates only on
+    # `enabled`, so fail-close them here rather than wait for an admin re-save.
+    rows = conn.execute(
+        sa.text("SELECT rowid, pricing FROM models WHERE enabled")
+    ).all()
+    stale = [rowid for rowid, pricing_json in rows if not _row_is_chargeable(pricing_json)]
+    if not stale:
+        return
+    conn.execute(
+        sa.text("UPDATE models SET enabled = 0 WHERE rowid IN :ids").bindparams(
+            sa.bindparam("ids", expanding=True)
+        ),
+        {"ids": stale},
     )
 
 
@@ -42,6 +98,7 @@ def upgrade() -> None:
         )
 
     _backfill_pricing_source(conn)
+    _disable_unchargeable_enabled_rows(conn)
 
 
 def downgrade() -> None:
--- a/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
+++ b/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
@@ -1,0 +1,68 @@
+"""add pricing provenance columns to models
+
+Revision ID: f0e1d2c3b4a5
+Revises: 9c4d8e2f1a6b
+Create Date: 2026-07-16 00:00:00.000000
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "f0e1d2c3b4a5"
+down_revision = "9c4d8e2f1a6b"
+branch_labels = None
+depends_on = None
+
+
+def _backfill_pricing_source(conn: sa.Connection) -> None:
+    # Existing rows predate provenance tracking, so their true origin is
+    # unknowable. Backfill "unresolved" rather than assert a trusted source:
+    # a persisted fail-closed zero-price row relabelled "openrouter" would slip
+    # past the re-enable guard. Freshness columns stay NULL — we never resolved
+    # these prices ourselves.
+    conn.execute(
+        sa.text(
+            "UPDATE models SET pricing_source = 'unresolved' "
+            "WHERE pricing_source IS NULL"
+        )
+    )
+
+
+def upgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+    columns = {c["name"] for c in inspector.get_columns("models")}
+
+    if "pricing_source" not in columns:
+        op.add_column(
+            "models",
+            sa.Column("pricing_source", sa.String(), nullable=True),
+        )
+    if "pricing_checked_at" not in columns:
+        op.add_column(
+            "models",
+            sa.Column("pricing_checked_at", sa.Integer(), nullable=True),
+        )
+    if "pricing_source_version" not in columns:
+        op.add_column(
+            "models",
+            sa.Column("pricing_source_version", sa.String(), nullable=True),
+        )
+
+    _backfill_pricing_source(conn)
+
+
+def downgrade() -> None:
+    conn = op.get_bind()
+    inspector = sa.inspect(conn)
+    columns = {c["name"] for c in inspector.get_columns("models")}
+
+    for column in (
+        "pricing_source_version",
+        "pricing_checked_at",
+        "pricing_source",
+    ):
+        if column in columns:
+            op.drop_column("models", column)
--- a/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
+++ b/migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
@@ -75,7 +75,9 @@ def _disable_unchargeable_enabled_rows(conn: sa.Connection) -> None:
     rows = conn.execute(
         sa.text("SELECT rowid, pricing FROM models WHERE enabled")
     ).all()
-    stale = [rowid for rowid, pricing_json in rows if not _row_is_chargeable(pricing_json)]
+    stale = [
+        rowid for rowid, pricing_json in rows if not _row_is_chargeable(pricing_json)
+    ]
     if not stale:
         return
     conn.execute(
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -13,9 +13,12 @@ from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from ..payment.models import (
+    BILLABLE_PRICING_FIELDS,
     Model,
+    Pricing,
     PricingSource,
     _row_to_model,
+    has_chargeable_price,
     list_models,
 )
 from ..proxy import refresh_model_maps, reinitialize_upstreams
@@ -524,21 +527,6 @@ class ModelCreate(BaseModel):
             raise ValueError(f"pricing_source must be one of: {allowed}")
 
 
-# The money-bearing pricing fields whose change flips a model to ``manual``.
-# Derived fields (max_*_cost) and ``sats_pricing`` are excluded — they are
-# computed/reset on every upsert and would false-flip provenance.
-_PRICING_RATE_FIELDS = (
-    "prompt",
-    "completion",
-    "request",
-    "image",
-    "web_search",
-    "internal_reasoning",
-    "input_cache_read",
-    "input_cache_write",
-)
-
-
 def _as_price(value: object) -> float | None:
     """Coerce a payload price to ``float``, tolerating numeric strings.
 
@@ -561,32 +549,47 @@ def _as_price(value: object) -> float | None:
     return None
 
 
-def _pricing_edited(payload: "ModelCreate", existing: Model) -> bool:
-    """True if the payload changes any money-bearing rate vs ``existing``.
+def _canonical_pricing(payload: "ModelCreate") -> Pricing:
+    """One complete ``Pricing`` from the payload, used for compare AND persist.
+
+    The write replaces the entire stored pricing JSON, which ``Pricing`` later
+    reparses with missing rates defaulting to 0. Building that same canonical
+    object here — and comparing/persisting *it* rather than the raw payload dict
+    — closes the gap where a replacement payload that omits a priced field reads
+    as "unchanged" yet silently drops the rate to zero. Numeric strings are
+    coerced (``_as_price``) so a string-typed edit is interpreted consistently.
+    """
+    values = {
+        field: (_as_price(payload.pricing.get(field)) or 0.0)
+        for field in BILLABLE_PRICING_FIELDS
+    }
+    return Pricing(**values)
+
+
+def _pricing_edited(canonical: Pricing, existing: Model) -> bool:
+    """True if the canonical price differs from ``existing`` on any billable rate.
 
     ``existing`` is the fee-free ``_row_to_model`` view the admin UI was shown
     (not the raw stored JSON): that view backfills litellm cache rates on read
     and the UI round-trips per-1M ↔ per-token, so a faithful "save as fetched"
     can legitimately differ from the stored JSON by cache rates or float noise.
-    Comparing against that view with ``isclose`` avoids false ``manual`` flips;
-    only a genuine operator price edit trips it.
+    Comparing the canonical object against that view with ``isclose`` avoids
+    false ``manual`` flips; only a genuine change to a stored rate trips it.
     """
-    payload_pricing = payload.pricing
-    for field in _PRICING_RATE_FIELDS:
-        new_value = _as_price(payload_pricing.get(field))
-        if new_value is None:
-            continue
-        if not math.isclose(
-            new_value,
+    return any(
+        not math.isclose(
+            getattr(canonical, field),
             getattr(existing.pricing, field),
             rel_tol=1e-9,
             abs_tol=0.0,
-        ):
-            return True
-    return False
+        )
+        for field in BILLABLE_PRICING_FIELDS
+    )
 
 
-def _resolve_provenance(payload: "ModelCreate", existing: Model | None) -> str | None:
+def _resolve_provenance(
+    canonical: Pricing, payload: "ModelCreate", existing: Model | None
+) -> str | None:
     """The ``pricing_source`` to persist for a write.
 
     - Update, price edited → ``manual`` (the operator owns this price).
@@ -599,7 +602,7 @@ def _resolve_provenance(payload: "ModelCreate", existing: Model | None) -> str |
       into a billable ``manual`` $0.
     """
     if existing is not None:
-        if _pricing_edited(payload, existing):
+        if _pricing_edited(canonical, existing):
             return PricingSource.MANUAL.value
         if payload.pricing_source is not None:
             return payload.pricing_source
@@ -607,37 +610,28 @@ def _resolve_provenance(payload: "ModelCreate", existing: Model | None) -> str |
         return source.value if source is not None else None
     if payload.pricing_source is not None:
         return payload.pricing_source
-    if _both_prices_zero(payload):
+    if not has_chargeable_price(canonical):
         return PricingSource.UNRESOLVED.value
     return PricingSource.MANUAL.value
 
 
-def _both_prices_zero(payload: "ModelCreate") -> bool:
-    """True when the payload prices both prompt and completion at exactly 0."""
-    prompt = _as_price(payload.pricing.get("prompt", 0))
-    completion = _as_price(payload.pricing.get("completion", 0))
-    if prompt is None or completion is None:
-        return False
-    return prompt == 0 and completion == 0
-
-
-def _effective_enabled(payload: "ModelCreate", source: str | None) -> bool:
+def _effective_enabled(
+    canonical: Pricing, requested_enabled: bool, source: str | None
+) -> bool:
     """The ``enabled`` flag to persist, force-disabling untrustworthy free rows.
 
-    A model priced at 0 whose price no source vouches for (``unresolved``/None)
+    An unchargeable model whose price no source vouches for (``unresolved``/None)
     would bill every request at nothing, so it is kept disabled regardless of
     the requested flag — the operator must give it a real price (which flips it
     to ``manual``) to enable it. A ``manual`` price is a deliberate declaration,
     and an operator editing a price *down to* zero owns that choice, so those
     rows keep exactly the ``enabled`` state the write requested.
     """
-    if not payload.enabled:
+    if not requested_enabled:
         return False
     if source == PricingSource.MANUAL.value:
         return True
-    if _both_prices_zero(payload):
-        return False
-    return True
+    return has_chargeable_price(canonical)
 
 
 @admin_router.post(
@@ -657,19 +651,22 @@ async def upsert_provider_model(
         # Try to get existing model
         existing_row = await session.get(ModelRow, (payload.id, provider_pk))
 
+        # One canonical price, compared and persisted identically.
+        canonical = _canonical_pricing(payload)
+
         if existing_row:
             # Update existing model
             logger.info(f"Updating existing model: {payload.id}")
             # Snapshot provenance from the fee-free view before mutating so a
             # price edit can be detected against what the UI was shown.
             existing_model = _row_to_model(existing_row, apply_provider_fee=False)
-            source = _resolve_provenance(payload, existing_model)
+            source = _resolve_provenance(canonical, payload, existing_model)
             existing_row.name = payload.name
             existing_row.description = payload.description
             existing_row.created = int(payload.created)
             existing_row.context_length = int(payload.context_length)
             existing_row.architecture = json.dumps(payload.architecture)
-            existing_row.pricing = json.dumps(payload.pricing)
+            existing_row.pricing = json.dumps(canonical.dict())
             existing_row.sats_pricing = None
             existing_row.per_request_limits = (
                 json.dumps(payload.per_request_limits)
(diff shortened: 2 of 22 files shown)


migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
```diff
@@ -0,0 +1,68 @@
...
+
+def _backfill_pricing_source(conn: sa.Connection) -> None:
+    # Existing rows predate provenance tracking, so their true origin is
+    # unknowable. Backfill "unresolved" rather than assert a trusted source:
+    # a persisted fail-closed zero-price row relabelled "openrouter" would slip
+    # past the re-enable guard. Freshness columns stay NULL — we never resolved
+    # these prices ourselves.
+    conn.execute(
+        sa.text(
+            "UPDATE models SET pricing_source = 'unresolved' "
```
9qeklajc, line 43: This gives the old rows an honest source, but it doesn't apply the fail-closed behavior to them. An existing row with `enabled = true` and zero pricing stays enabled because `list_models()` still gates only on `enabled`; it won't be disabled until somebody saves it through the admin endpoint. That leaves the exact pre-existing condition this PR is meant to fix in place after migration.

Could we disable existing unchargeable rows here, then use the same shared predicate when building the served model map? Something along the lines of:

```python
def has_chargeable_price(pricing: Pricing) -> bool:
    return any(getattr(pricing, field) > 0 for field in BILLABLE_FIELDS)
```

The served-map guard would also cover legacy data and writers that don't use this endpoint.

routstr/core/admin.py
```diff
@@ -461,6 +468,164 @@ class ModelCreate(BaseModel):
...
+    ``existing`` is the fee-free ``_row_to_model`` view the admin UI was shown
+    (not the raw stored JSON): that view backfills litellm cache rates on read
+    and the UI round-trips per-1M ↔ per-token, so a faithful "save as fetched"
+    can legitimately differ from the stored JSON by cache rates or float noise.
+    Comparing against that view with ``isclose`` avoids false ``manual`` flips;
+    only a genuine operator price edit trips it.
+    """
+    payload_pricing = payload.pricing
+    for field in _PRICING_RATE_FIELDS:
+        new_value = _as_price(payload_pricing.get(field))
```
9qeklajc: Skipping a missing field is unsafe here because the write below replaces the entire pricing JSON. I reproduced this with an existing `request=0.5`: posting unchanged prompt/completion without `request` makes `_pricing_edited()` return `False`, but reparsing the replacement payload resets `request` to `0.0`, so the old trusted provenance survives a real price change.

I'd canonicalize one complete `Pricing` object and use that for both comparison and persistence, or reject incomplete replacement payloads. That also gives us one definition of the billable fields instead of having comparison and storage interpret the payload differently.

routstr/upstream/ollama.py
```diff
@@ -148,6 +155,7 @@ async def fetch_models(self) -> list[Model]:
                             enabled=True,
                             upstream_provider_id=None,
                             canonical_slug=None,
+                            **pricing_metadata(PricingSource.NATIVE),
```
9qeklajc: I don't think `native` is truthful for this price. The rates just above are hard-coded by Routstr; they didn't come from Ollama's `/api/tags` response, while `PricingSource.NATIVE` is documented as provider-supplied pricing. That makes the new audit field confidently report the wrong origin.

Can we run Ollama models through the shared fallback resolver and preserve its source, or mark them unresolved when no source knows the model? I'd avoid adding an Ollama-specific interpretation of `native`.

routstr/upstream/ppqai.py
```diff
@@ -158,6 +170,15 @@ async def fetch_models(self) -> list[Model]:
                             if output_price is not None:
                                 or_model.pricing.completion = output_price / 1_000_000
 
+                            # When PPQ published its own price we overwrote the
+                            # OpenRouter number, so the price is now native;
+                            # otherwise the OR feed's price (and its tag) stand.
+                            if input_price is not None or output_price is not None:
```
9qeklajc: This changes the source for the whole price when PPQ supplies either side. If PPQ supplies input only, completion is still OpenRouter-derived, but the model now reads as fully native. In the unmatched branch, the same `or` rule can leave one side at zero while still calling the model native.

With model-wide provenance, I'd require both PPQ rates before assigning `NATIVE`:

```python
has_complete_ppq_price = input_price is not None and output_price is not None
```

Otherwise keep OpenRouter provenance for a matched model and mark an unmatched partial price unresolved/disabled. If mixed-source prices are intentional, we'll eventually need provenance per component rather than one tag for the whole `Pricing`.

9qeklajc (changes requested):
I like the direction here. Making price origin queryable is useful, and the care around numeric strings, fee/sats rebuilds, malformed tags, and PPQ aliasing shows up in the tests.

I do think this needs another pass before merging. The main safety goal is smaller than the machinery added here—an enabled model must have a chargeable price unless an operator explicitly approves it as free—but the migration doesn't apply that rule to the existing rows that motivated the change. There are also a few places where `native` doesn't describe where the full price actually came from.

My preference would be to centralize a `has_chargeable_price()` invariant, enforce it on writes and when building the served model map, add an explicit operator `allow_free` control, and keep `pricing_source` if auditability is a requirement. I'd defer `pricing_checked_at` and `pricing_source_version` until a freshness policy actually consumes them. That gets the money-safety rule in place with less propagation and makes the provenance semantics easier to trust.

I left the concrete issues inline. Focused provenance tests pass locally (11 unit, 13 integration), and CI is green. Once the existing-row gap and source mismatches are fixed, the core idea will be on much firmer ground.

jeroenubbink:
Thanks — this was the right push. The money-safety rule is now centralized and enforced in both places, the freshness columns are deferred, and every `native` mislabel you flagged is corrected. Rundown:

**`has_chargeable_price()` invariant, enforced on writes and the served map** — done in [`95a82b5`](https://github.com/Routstr/routstr-core/pull/617/commits/95a82b5). One `has_chargeable_price(pricing)` over a shared `BILLABLE_PRICING_FIELDS` (all 8 billable rates, not just prompt/completion), enforced when building the served model map *and* in the migration, so legacy rows and writers that bypass the admin endpoint are both covered.

**Existing-row gap** — the migration now fail-closes: enabled rows with no chargeable price and a non-`manual` source are disabled in place ([`95a82b5`](https://github.com/Routstr/routstr-core/pull/617/commits/95a82b5)), so the pre-existing condition doesn't survive the upgrade.

**Defer `pricing_checked_at` / `pricing_source_version`** — agreed and done in [`48942b2`](https://github.com/Routstr/routstr-core/pull/617/commits/48942b2): the migration now adds only `pricing_source`. Both freshness columns move to the freshness-policy PR that actually reads them, so state lands with its consumer.

**`allow_free` control** — I'd rather not add a second free-approval axis. `manual` already carries "an operator vouched for this price, including a deliberate $0": `_effective_enabled` keeps a `manual` row exactly as the write requested (free or not), and force-disables only untrusted (`unresolved`/None) unchargeable rows. So the money-safety rule is "enabled ⇒ chargeable **unless** manual", which is the same gate `allow_free` would give without a new boolean to keep in sync. An explicit "this is a manual price" UI affordance is a reasonable follow-up, but as UI over the existing source, not a new stored flag. Happy to reconsider if you see a case `manual` can't express.

Inline replies on each thread. While self-reviewing the fixes I also caught and fixed three related issues — noted them inline too.

migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
jeroenubbink: Fixed in [`95a82b5`](https://github.com/Routstr/routstr-core/pull/617/commits/95a82b5). The migration now disables enabled-but-unchargeable non-`manual` rows in place (frozen field copy of `BILLABLE_PRICING_FIELDS`), and `list_models()` gates the served map on the same `has_chargeable_price()` predicate — so legacy data and non-admin writers are both covered, exactly as suggested.

routstr/core/admin.py
jeroenubbink: Good catch, this was a real hole. Fixed in [`c0c4bfa`](https://github.com/Routstr/routstr-core/pull/617/commits/c0c4bfa): `_canonical_pricing(payload)` builds one complete `Pricing` used for **both** the edit comparison and the persisted JSON, so an omitted `request` can no longer compare-equal while being reset to 0 on write.

Self-review follow-up: the compare then had the opposite asymmetry — it checked the raw canonical against the fee-free `existing` view, which backfills litellm cache rates on read, so a non-UI client omitting those backfilled rates falsely flipped a trusted source to `manual`. [`704a00d`](https://github.com/Routstr/routstr-core/pull/617/commits/704a00d) backfills both sides before comparing, so they're like-for-like; a stored rate with no backfilled twin (e.g. `request`) still trips as the genuine edit it is.

routstr/upstream/ollama.py
jeroenubbink: Agreed — `/api/tags` carries no price, so `native` was a lie. Fixed in [`614be86`](https://github.com/Routstr/routstr-core/pull/617/commits/614be86): Ollama models now run through the shared `FallbackPricingResolver` and wear whatever source it resolves, or `unresolved` + disabled when no source knows the model. No Ollama-specific reading of `native` anymore.

routstr/upstream/ppqai.py
jeroenubbink: Fixed in [`c086228`](https://github.com/Routstr/routstr-core/pull/617/commits/c086228) using your `has_complete_ppq_price` shape. A matched model now keeps OpenRouter provenance unless PPQ priced *both* sides; an unmatched partial price is `unresolved` + `enabled=False`. Whole-`Pricing` single-tag stands for now, with per-component provenance noted as the future work if we keep mixed-source prices.

Self-review follow-up on the same branch: the matched overlay gated on `is not None`, so a one-sided PPQ `0` (absent-as-zero, not free) overwrote the matched OpenRouter rate with $0 and billed those tokens at nothing while the model stayed enabled and tagged openrouter. [`43040fd`](https://github.com/Routstr/routstr-core/pull/617/commits/43040fd) gates the overlay (and the both-sides native tag) on a truthy price, matching the unmatched branch.

Also worth flagging: single upsert returns the row so a force-disable is visible, but batch-override was silent. [`29031ad`](https://github.com/Routstr/routstr-core/pull/617/commits/29031ad) returns `force_disabled: [ids]` from batch so a bulk caller can tell which entries didn't land enabled.

9qeklajc (commented):
Three small follow-ups from another pass; the first one still leaves the original served-map safety request open.

routstr/payment/models.py
```diff
@@ -309,21 +395,32 @@ async def list_models(
...
-            and providers_by_id[r.upstream_provider_id].enabled
-        )
-    ]
+        # Served-map money-safety backstop for legacy rows and writers that
+        # bypass the admin upsert: never serve an unchargeable price unless an
+        # operator vouched for it as free (``manual``); it would bill at nothing.
+        if (
+            not include_disabled
+            and model.pricing_source != PricingSource.MANUAL
+            and not has_chargeable_price(model.pricing)
```
9qeklajc: I think this guard is on the wrong path: `refresh_model_maps()` reads enabled rows directly, so an unresolved $0 override is still routable. Could we enforce this while building candidates too?

routstr/core/admin.py
```diff
@@ -473,6 +483,140 @@ class ModelCreate(BaseModel):
...
+
+    The write replaces the entire stored pricing JSON, which ``Pricing`` later
+    reparses with missing rates defaulting to 0. Building that same canonical
+    object here — and comparing/persisting *it* rather than the raw payload dict
+    — closes the gap where a replacement payload that omits a priced field reads
+    as "unchanged" yet silently drops the rate to zero. Numeric strings are
+    coerced (``_as_price``) so a string-typed edit is interpreted consistently.
+    """
+    values = {
+        field: (_as_price(payload.pricing.get(field)) or 0.0)
```
9qeklajc, line 595: This turns a present invalid value (`"oops"`, `NaN`, or `-1`) into enabled manual pricing. Can we reject malformed, non-finite, and negative rates with a 422 instead?

routstr/upstream/ppqai.py
```diff
@@ -163,7 +175,12 @@ async def fetch_models(self) -> list[Model]:
                             elif ppqai_model.pricing.input_per_1M_tokens:
                                 input_price = ppqai_model.pricing.input_per_1M_tokens
 
-                            if input_price is not None:
+                            # A price of 0 means PPQ did not really price this
+                            # side (absent-as-zero), not that the tokens are
+                            # free: overlaying it would zero the matched
+                            # OpenRouter rate and bill those tokens at nothing.
+                            # Only a positive PPQ price overrides OpenRouter's.
+                            if input_price:
```
9qeklajc: Negative floats are truthy here, so PPQ can still overwrite OpenRouter with a negative rate and later mark it native. Could we require a finite value `> 0` first?

routstr/payment/models.py
jeroenubbink: You're right — the guard was only on the served catalog (`list_models`), while routing built candidates straight off `enabled` in `create_model_mappings`, so an unresolved-$0 override was still routable. Fixed in [`15b1935`](https://github.com/Routstr/routstr-core/pull/617/commits/15b1935): a `_override_unroutable_free` helper mirrors the served-catalog rule (`source != manual and not has_chargeable_price`) and is applied at both override-consumption sites in `create_model_mappings`. Scoped to override rows so discovered-provider models are untouched. Two unit tests: unresolved-$0 override excluded from the maps; `manual`-$0 still routes.

routstr/core/admin.py
jeroenubbink: Confirmed — `_as_price` coerces a non-numeric rate to `0.0` and `NaN`/`inf`/negatives are truthy, so a malformed manual price could land on an enabled, chargeable-looking model. Fixed in [`587844d`](https://github.com/Routstr/routstr-core/pull/617/commits/587844d) with a `@field_validator("pricing")` on `ModelCreate` (the single carrier for the upsert, legacy-update and batch endpoints): any *present* billable rate that is non-numeric, non-finite or negative is rejected with 422. Absent rates and numeric strings still pass. Integration tests cover negative → 422, malformed string → 422, and `NaN`/`±inf` → validation error.

routstr/upstream/ppqai.py
jeroenubbink: Fixed in [`53b4947`](https://github.com/Routstr/routstr-core/pull/617/commits/53b4947). Added a `_positive_rate` helper (finite `> 0`, else `None`) and routed both rates through it. Applied to both this overlay branch and the sibling standalone branch, which had the same negative→`native`/enabled bug. Native tagging now requires both sides positive-finite; the standalone path falls to `unresolved`/disabled otherwise. Two unit tests: a negative overlay leaves the matched OpenRouter rate intact and keeps `openrouter`; a negative standalone imports `unresolved` + disabled.

While here I checked the other price-ingestion paths for the same class of bug: the Venice native reader, both resolver rungs (litellm + the `_has_valid_pricing`-filtered OpenRouter feed) and the Ollama path already guard negative/non-finite/both-zero. The one remaining unguarded spot is the Tinfoil catalog reader, which builds pricing straight from its raw per-1M rates — it predates this PR and doesn't touch provenance, so I've left it for a separate follow-up rather than widen scope here.

9qeklajc:
can u pls check these too:

 1. Provider-discovered zero-price models can remain routable, notably Tinfoil models with missing pricing.
 2. Shared pricing validation accepts NaN, +inf, and mixed positive/invalid rates.
 3. Tinfoil native pricing still lacks provenance and gets persisted as manual.
 4. Oversized JSON integers raise OverflowError instead of returning 422.


pls recreate the migration

jeroenubbink:
> can u pls check these too:
> 
>     1. Provider-discovered zero-price models can remain routable, notably Tinfoil models with missing pricing.

Fixed. Now holds back unchargeable free models

>     2. Shared pricing validation accepts NaN, +inf, and mixed positive/invalid rates.

Fixed

>     3. Tinfoil native pricing still lacks provenance and gets persisted as manual.

TInfoil was rolled in while this PR was already open. I thought about leaving it out but it is now covered!

>     4. Oversized JSON integers raise OverflowError instead of returning 422.

Should work now
 
> pls recreate the migration

done.

jeroenubbink:
Two more commits since the above, plus one note on the fourth point.

**Non-finite rates (item 2), follow-up** — [`a2573af`](https://github.com/Routstr/routstr-core/pull/617/commits/a2573af). The validator was already rejecting `NaN`/`inf` correctly, but a bare `Infinity`/`NaN` JSON literal still came back as a **500**: pydantic echoes the rejected value in the error's `input` field, and `JSONResponse` encodes with `allow_nan=False`, so serializing the 422 raised "Out of range float values are not JSON compliant". Handled `RequestValidationError` explicitly and render non-finite floats as text, so it covers every endpoint rather than the pricing edge alone. Verified against a running node: `Infinity`, `NaN`, `1e999`, mixed good+`NaN`, and the batch-override path all return 422 now. Nothing was ever persisted on those requests — it failed closed, so this was a status-code defect, not a billing one.

That commit also makes the integration app copy the main app's exception handlers (`tests/integration/conftest.py`). It previously copied routes only, so integration tests exercised an app that handled errors differently from production — which is why this hole survived. Full suite green with the change.

**`tests/unit/test_fee_payout_migration.py`** — [`c631c3e`](https://github.com/Routstr/routstr-core/pull/617/commits/c631c3e) changes a test that came in with the re-apply-migration PR, so flagging it deliberately: it asserted the alembic head against a hardcoded literal, which fails on any branch that adds a migration. It now derives the head from `alembic heads` and asserts there's exactly one, keeping the single-head invariant. Worth knowing it would have broken every future migration PR, not just this one.

routstr/payment/models.py
```diff
@@ -58,6 +111,45 @@ class Pricing(BaseModel):
...
+    from upstream catalogs as well as the admin edge (``json.loads`` accepts the
+    bare ``NaN``/``Infinity`` literals and overflows ``1e999`` to ``inf``), so
+    the shared predicate rejects them wherever they came from. One bad rate
+    disqualifies the model even alongside a valid one, since a request can bill
+    on the bad field.
+    """
+    rates = [getattr(pricing, field) for field in BILLABLE_PRICING_FIELDS]
+    if any(not math.isfinite(rate) for rate in rates):
+        return False
+    return any(rate > 0 for rate in rates)
```
9qeklajc: This red test makes the remaining shared-predicate hole explicit:

```python
def test_has_chargeable_price_false_when_one_rate_is_negative() -> None:
    pricing = Pricing(prompt=-1.0, completion=1.0)
    assert has_chargeable_price(pricing) is False
```

It currently fails with `assert True is False`. Since upstream/foreign DB writes bypass the admin validator, one negative billable field needs to invalidate the whole price before the final `any(rate > 0)` check.

9qeklajc (commented):
The latest fixes close the earlier admin/routing gaps nicely. I wrote four focused regression tests against the remaining pricing-safety cases so the failure modes are concrete.

All four tests are intentionally red on the current PR head (`a2573aff`):

- mixed negative/positive runtime pricing is still considered chargeable;
- the migration accepts mixed negative/positive pricing;
- the migration accepts `Infinity`;
- a PPQ model can be tagged `native` while retaining every auxiliary OpenRouter charge.

Focused run: **4 failed in 1.44s**, each at the intended assertion. Ruff and `git diff --check` are clean. These are test-only changes; nothing was committed or pushed to the branch.

<details><summary>Test patch (expected to fail until the production fixes land)</summary>

```diff
diff --git a/tests/unit/test_pricing_provenance.py b/tests/unit/test_pricing_provenance.py
index 965de35b..f33c734b 100644
--- a/tests/unit/test_pricing_provenance.py
+++ b/tests/unit/test_pricing_provenance.py
@@ -157,6 +157,19 @@ def test_has_chargeable_price_false_for_non_finite_rates() -> None:
     assert has_chargeable_price(Pricing(prompt=1e-06, completion=float("inf"))) is False
 
 
+def test_has_chargeable_price_false_when_one_rate_is_negative() -> None:
+    """One positive field must not hide a malformed negative billable field.
+
+    Upstream catalogs and foreign/direct database writers bypass the admin
+    validator. If this shared predicate accepts the row, listing and routing use
+    it as their money-safety backstop and a request can still bill on the
+    negative component.
+    """
+    pricing = Pricing(prompt=-1.0, completion=1.0)
+
+    assert has_chargeable_price(pricing) is False
+
+
 @pytest.mark.asyncio
 async def test_openrouter_rung_rejects_non_finite_feed_price() -> None:
     """``json.loads`` accepts bare ``NaN``/``Infinity`` and overflows ``1e999`` to
@@ -453,6 +466,83 @@ async def test_ppqai_matched_partial_price_keeps_openrouter_provenance() -> None
     assert model.pricing_source == PricingSource.OPENROUTER
 
 
+@pytest.mark.asyncio
+async def test_ppqai_native_price_does_not_retain_openrouter_auxiliary_rates() -> None:
+    """Whole-``Pricing`` provenance cannot say ``native`` while billable fields
+    still come from OpenRouter.
+
+    PPQ publishes only input/output token prices. Once both PPQ rates replace the
+    matched token rates and the model is relabelled ``native``, unrelated
+    OpenRouter request/image/cache charges must not survive on that price.
+    Otherwise Routstr can bill fees PPQ never published while the audit tag says
+    every rate is provider-native.
+    """
+    from routstr.upstream.ppqai import PPQAIUpstreamProvider
+
+    payload = {
+        "data": [
+            {
+                "id": "gpt-4o",
+                "name": "GPT-4o",
+                "created_at": 0,
+                "context_length": 8192,
+                "pricing": {
+                    "api": {"input_per_1M": 5.0, "output_per_1M": 15.0}
+                },
+            }
+        ]
+    }
+    or_feed = [
+        {
+            "id": "gpt-4o",
+            "name": "GPT-4o",
+            "created": 0,
+            "description": "d",
+            "context_length": 8192,
+            "architecture": {
+                "modality": "text",
+                "input_modalities": ["text"],
+                "output_modalities": ["text"],
+                "tokenizer": "unknown",
+                "instruct_type": None,
+            },
+            "pricing": {
+                "prompt": 0.000001,
+                "completion": 0.000002,
+                "request": 0.25,
+                "image": 0.5,
+                "web_search": 0.75,
+                "internal_reasoning": 0.0000003,
+                "input_cache_read": 0.0000001,
+                "input_cache_write": 0.0000002,
+            },
+            "pricing_source": "openrouter",
+        }
+    ]
+
+    provider = PPQAIUpstreamProvider(api_key="k")
+    with patch(
+        "routstr.upstream.ppqai.httpx.AsyncClient",
+        lambda *a, **k: _PPQClient(payload),
+    ):
+        with patch(
+            "routstr.upstream.ppqai.async_fetch_openrouter_models",
+            AsyncMock(return_value=or_feed),
+        ):
+            models = await provider.fetch_models()
+
+    model = _model_by_id(models, "gpt-4o")
+    assert model.pricing_source == PricingSource.NATIVE
+    assert model.pricing.prompt == pytest.approx(5.0 / 1_000_000)
+    assert model.pricing.completion == pytest.approx(15.0 / 1_000_000)
+    assert model.pricing.request == 0.0
+    assert model.pricing.image == 0.0
+    assert model.pricing.web_search == 0.0
+    assert model.pricing.internal_reasoning == 0.0
+    assert model.pricing.input_cache_read == 0.0
+    assert model.pricing.input_cache_write == 0.0
+
+
 @pytest.mark.asyncio
 async def test_ppqai_matched_zero_side_does_not_zero_openrouter_price() -> None:
     """A PPQ price of 0 on one side means PPQ did not really price that side, not
diff --git a/tests/unit/test_pricing_provenance_migration.py b/tests/unit/test_pricing_provenance_migration.py
index 0ad4c01f..b873748d 100644
--- a/tests/unit/test_pricing_provenance_migration.py
+++ b/tests/unit/test_pricing_provenance_migration.py
@@ -55,6 +55,18 @@ def test_backfill_assigns_unresolved_and_preserves_existing() -> None:
     assert rows == [("a", "unresolved"), ("b", "manual")]
 
 
+def test_row_is_chargeable_rejects_mixed_negative_and_positive_rates() -> None:
+    """A positive field must not hide a negative field that can still be billed."""
+    assert (
+        migration._row_is_chargeable('{"prompt": -1, "completion": 1}') is False
+    )
+
+
+def test_row_is_chargeable_rejects_infinity() -> None:
+    """Python's JSON parser accepts ``Infinity``, but it is not a usable price."""
+    assert migration._row_is_chargeable('{"prompt": Infinity}') is False
+
+
 def test_disables_enabled_unchargeable_rows_only() -> None:
     """Legacy rows that motivated the change — enabled but priced at nothing —
     are fail-closed by the migration itself, not only when re-saved through the

```
</details>

The migration itself is real and active in this branch: `f0e1d2c3b4a5` is the sole Alembic head and a copied developer DB upgraded successfully from `7f2843d3f4e4`. The actual developer DB was left untouched and is still on `7f2843d3f4e4`.

migrations/versions/f0e1d2c3b4a5_add_pricing_provenance_to_models.py
```diff
@@ -0,0 +1,112 @@
...
+        return False
+    try:
+        pricing = json.loads(pricing_json)
+    except (TypeError, ValueError):
+        return False
+    if not isinstance(pricing, dict):
+        return False
+    for field in _BILLABLE_FIELDS:
+        try:
+            if float(pricing.get(field, 0) or 0) > 0:
```
9qeklajc: Two red migration tests pin down the fail-open behavior here:

```python
def test_row_is_chargeable_rejects_mixed_negative_and_positive_rates() -> None:
    assert migration._row_is_chargeable(
        '{"prompt": -1, "completion": 1}'
    ) is False


def test_row_is_chargeable_rejects_infinity() -> None:
    assert migration._row_is_chargeable('{"prompt": Infinity}') is False
```

Both currently return `True`. Please validate every billable field as finite and non-negative first, then require at least one positive field.

routstr/upstream/ppqai.py
```diff
@@ -174,29 +209,57 @@ async def fetch_models(self) -> list[Model]:
...
 
+                            output_price = _positive_rate(output_price)
                             if output_price is not None:
                                 or_model.pricing.completion = output_price / 1_000_000
 
+                            # Only a model PPQ priced on *both* sides is fully
+                            # native; if PPQ supplied one side, the other is
+                            # still OpenRouter-derived, so the whole-Pricing tag
+                            # stays whatever the OR feed carried (openrouter).
+                            if input_price is not None and output_price is not None:
```
9qeklajc, line 231: The added PPQ red test feeds nonzero OpenRouter values for every auxiliary billable field (`request`, `image`, `web_search`, reasoning, and both cache rates), then supplies complete PPQ token rates. The result is `PricingSource.NATIVE`, but all six OpenRouter-derived charges survive.

The test expects those fields not to survive under a whole-`Pricing` native tag. It currently fails first at `request == 0.0` with `request=0.25`; the failure object shows the other five retained values too. Clearing/rebuilding the complete PPQ price—or retaining OpenRouter provenance—would satisfy the invariant.

jeroenubbink:
Thanks — all four of your tests are in and green. Since this class of bug keeps coming
back, I also audited the other paths a rate can take into the money math, which turned
up six more.

**Your four**

- [`bb938b2`](https://github.com/Routstr/routstr-core/pull/617/commits/bb938b2) — a
  positive field could vouch for a negative one. Extracted `is_usable_rate` (finite and
  non-negative) so the guards stop each defining "usable" differently.
- [`e0900c1`](https://github.com/Routstr/routstr-core/pull/617/commits/e0900c1) — the
  migration predicate had neither guard, so negative and `Infinity` rows survived the
  fail-close sweep meant to catch them. Oversized ints (`OverflowError`) now fail closed
  too.
- [`95e95d3`](https://github.com/Routstr/routstr-core/pull/617/commits/95e95d3) — PPQ's
  native price is rebuilt from the two rates PPQ publishes instead of overlaid on the
  matched OpenRouter entry.

**Found by the audit** — each with a test that fails without the fix:

- [`79fc271`](https://github.com/Routstr/routstr-core/pull/617/commits/79fc271) — the
  "no token pricing" gate tested truthiness, and `NaN`/negatives are truthy, so they
  reached the token math and raised *after* the response was served (request unbilled).
- [`14ea10d`](https://github.com/Routstr/routstr-core/pull/617/commits/14ea10d) — a
  non-finite figure in `cost_details` poisons the split (`inf / inf` is `NaN`), the
  exception is swallowed, and a request with a valid $0.01 total bills 2 msats. Measured.
- [`9f09999`](https://github.com/Routstr/routstr-core/pull/617/commits/9f09999) — the
  BTC/USD feed takes `min()`, so a `0`/negative/`NaN` quote doesn't join the sample, it
  *wins* it. Booleans too: `float(True)` passes every numeric guard and prices the node
  at $1/BTC.
- [`1f781ee`](https://github.com/Routstr/routstr-core/pull/617/commits/1f781ee) — the
  served-catalog backstop exempted `manual` from malformed as well as from free. Zero is
  a price an operator can vouch for; a negative or non-finite rate isn't a price at all.
- [`094f1e2`](https://github.com/Routstr/routstr-core/pull/617/commits/094f1e2) — one row
  with unparseable stored pricing raised out of the whole listing, so the node advertised
  nothing.
- [`05e87e1`](https://github.com/Routstr/routstr-core/pull/617/commits/05e87e1) — the
  admin listing rendered a non-finite rate as `null`, indistinguishable from a missing
  one, on the one view that still shows the held-back row.

The last two had exact twins in the routing path that I'd missed —
[`28025d4`](https://github.com/Routstr/routstr-core/pull/617/commits/28025d4) (a
`manual` row with a `NaN` rate was hidden from `/v1/models` but still routable, billing
every request at the full max-cost reservation) and
[`d347081`](https://github.com/Routstr/routstr-core/pull/617/commits/d347081) (one
unparseable override row emptied the whole routing map).

**One thing to check me on.** I edited migration `f0e1d2c3b4a5` **in place** rather than
recreating it. Your earlier "recreate it" ask was about a migration amended across rounds until its
id no longer described one schema step; this touches only the data guard inside a migration that has
never been deployed (your DB is still on `7f2843d3f4e4`) and its schema step is unchanged. Happy to
recreate it if you would rather have the clean id.

**Follow-ups, separate PRs.** The root cause is that the node has ten independent "is
this a number I can bill on" implementations with different rules; the last non-finite
sweep updated five of them, which is why the negative case was still live. This PR gives
them one predicate to delegate to and converts the paths it touches — a follow-up would
route the rest through `is_usable_rate` and delete the duplicates, as a pure
consolidation with no behaviour change. A second PR covers the settlement side
(unvalidated `provider_fee`, no clamp on `total_msats` so a negative cost credits the
caller, and a magnitude floor on exchange quotes) — that one needs a policy decision
first, so we want to agree on it before writing it.

9qeklajc:
**Request-only billing regression**

```python
# tests/unit/test_algorithm.py
from routstr.payment.cost_calculation import calculate_cost  # noqa: E402
    _update_model_sats_pricing,
from routstr.upstream.generic import GenericUpstreamProvider  # noqa: E402


@pytest.mark.asyncio
async def test_request_only_price_routes_but_settles_for_advertised_charge() -> None:
    """The PR's fail-closed predicate must mean a routed price is collectible.

    ``has_chargeable_price`` currently vouches for a request-only price. Drive
    that model through the same fee/max-cost and sats rebuilds used by refresh,
    route it, then settle an ordinary token response. The charge must cover the
    advertised per-request rate rather than collapsing to the minimum request
    reservation.
    """
    model = create_test_model(
        "generic/request-priced",
        prompt_price=0.0,
        completion_price=0.0,
        request_price=0.5,
    )
    model.pricing_source = PricingSource.NATIVE
    fee_provider = GenericUpstreamProvider(base_url="https://example.test/v1")
    priced = fee_provider._apply_provider_fee_to_model(model)
    priced = _update_model_sats_pricing(priced, sats_to_usd=0.001)
    provider = create_test_provider(
        "generic",
        "https://example.test/v1",
        db_id=9,
        models=[priced],
    )

    _, provider_map, _ = create_model_mappings(
        upstreams=[provider],
        overrides_by_key={},
        disabled_model_keys=set(),
    )
    routed_model, _ = provider_map["generic/request-priced"][0]
    assert routed_model.sats_pricing is not None
    reserved_msats = int(routed_model.sats_pricing.max_cost * 1000)
    advertised_request_msats = int(routed_model.sats_pricing.request * 1000)

    settled = await calculate_cost(
        {
            "model": routed_model.id,
            "usage": {"prompt_tokens": 100, "completion_tokens": 20},
        },
        reserved_msats,
        model_obj=routed_model,
    )

    assert settled.total_msats >= advertised_request_msats
```

9qeklajc:
**Cache provenance regression**

```python
# tests/unit/test_pricing_provenance.py
def test_cache_backfill_does_not_leave_whole_price_tagged_native() -> None:
    """Refresh must not attach LiteLLM rates to a wholly ``native`` price.

    PPQ's fetch path now clears OpenRouter auxiliary rates before assigning its
    whole-``Pricing`` native tag. The shared provider-fee enrichment then fills
    missing cache rates from LiteLLM. Once that happens the resulting price can
    no longer truthfully claim that every advertised rate came from PPQ.
    """
    from routstr.upstream.ppqai import PPQAIUpstreamProvider

    model = _model_with_source(PricingSource.NATIVE)
    model.id = "gpt-4o"
    assert model.pricing.input_cache_read == 0.0

    provider = PPQAIUpstreamProvider(api_key="k")
    rebuilt = provider._apply_provider_fee_to_model(model)

    assert rebuilt.pricing.input_cache_read > 0.0
    assert rebuilt.pricing_source != PricingSource.NATIVE
```

9qeklajc:
**Admin repair regression**

```python
# tests/integration/test_admin_pricing_provenance.py
from pydantic.v1 import ValidationError
@pytest.mark.integration
@pytest.mark.asyncio
async def test_admin_upsert_repairs_an_unreadable_stored_price(
    integration_client: AsyncClient,
    integration_session: AsyncSession,
) -> None:
    """An operator must be able to replace a legacy/foreign malformed price.

    The served-catalog guard correctly drops a row that cannot become a
    ``Pricing`` object, but the admin write path still reads the broken value
    before replacing it. A valid full upsert should be the recovery path rather
    than returning 500 and leaving the row permanently unrepairable.
    """
    provider_id = await _make_provider(integration_session)
    integration_session.add(
        ModelRow(
            id="repair-me",
            name="repair-me",
            description="d",
            created=0,
            context_length=8192,
            architecture=json.dumps(
                {
                    "modality": "text",
                    "input_modalities": ["text"],
                    "output_modalities": ["text"],
                    "tokenizer": "unknown",
                    "instruct_type": None,
                }
            ),
            pricing=json.dumps({"prompt": "not-a-number", "completion": 2e-06}),
            upstream_provider_id=provider_id,
            enabled=False,
            forwarded_model_id="repair-me",
            pricing_source="unresolved",
        )
    )
    await integration_session.commit()

    try:
        resp = await integration_client.post(
            f"/admin/api/upstream-providers/{provider_id}/models",
            headers=_admin_headers(),
            json=_payload(
                provider_id,
                model_id="repair-me",
                prompt=1e-06,
                completion=2e-06,
                enabled=True,
            ),
        )
        repair_status = resp.status_code
    except ValidationError:
        # The integration transport re-raises the exception that production
        # turns into a 500. Normalize it so the regression fails on the API
        # contract below, rather than looking like fixture/setup breakage.
        repair_status = 500

    assert repair_status == 200
    integration_session.expire_all()
    row = await integration_session.get(ModelRow, ("repair-me", provider_id))
    assert row is not None
    assert json.loads(row.pricing)["prompt"] == pytest.approx(1e-06)
    assert row.pricing_source == "manual"
    assert row.enabled is True
```

Small migration note: fixing the first case also changes the frozen chargeability rule in the migration. Since #634 needs to land and already adds a migration from the same current head, could we please create/regenerate this migration on top of #634 once it merges, so the Alembic chain stays linear?

PR #619, opened 2026-07-17
test: vulnerability-reproducing RED tests + coverage (10 fail, 857 pass)

9qeklajc:
## Summary

**10 RED tests that assert CORRECT behavior — they FAIL against current buggy main.**

These are TRUE TDD RED tests: they document bugs by asserting the FIXED behavior.
When the bugs are fixed, these tests go green — serving as regression guards.

### RED Tests (fail against main — document live bugs)

| Test | Bug Documented |
|---|---|
| `test_store_cashu_raises_on_db_failure` | DB errors silently return False — callers never know |
| `test_retry_wrapper_exists` | Retry wrapper merged (#600) then reverted (#604) |
| `test_emergency_refund_no_try_except_pass` | Emergency refund mints token then `except: pass` → silent fund loss |
| `test_fee_payout_has_crash_guard` | Pay-then-reset crash window can double-pay fees |
| `test_billing_error_must_not_hardcode_zero_cost` | `base.py:1026` hardcodes `total_msats: 0` → free service |
| `test_billing_error_must_release_reserved_balance` | Reserved balance never released on billing error |
| `test_billing_error_catch_is_too_broad` | `except Exception` catches transient errors → free inference |
| `test_billing_error_must_log_critical` | Money loss logged at ERROR not CRITICAL |

### Coverage Tests (45 pass, zero regressions)

| Module | Before | After | Tests |
|---|---|---|---|
| `middleware.py` | 38% | **90%** | 11 (new) |
| `payment/helpers.py` | 52% | **60%** | 10 (existing) |
| `base.py` | 41% | **41%** | 17 (new) |
| `proxy.py` | 47% | **51%** | 13 (new) |
| `admin.py` | 35% | **36%** | 11 (new) |

### Test Results

```
10 failed (RED — correct), 857 passed, 14 skipped
```

All RED failures are intentional — they assert the behavior the code SHOULD have.
All GREEN tests are new coverage — zero regressions.

PR #620, opened 2026-07-18
fix: guard fee payouts against crash double-payments

9qeklajc:
Checkpoints fee payouts before sending so 2 of PR #619’s 10 red tests move toward green.


9qeklajc:
Independent review complete; restart and migration safety are covered, moving 2 of PR #619’s 10 failures to green.

PR #621, opened 2026-07-18
fix: propagate Cashu storage errors

9qeklajc:
Stops hiding failed money-path writes so 2 more of PR #619’s 10 red tests move to green.


9qeklajc:
Independent review is clean after caller recovery checks, moving 2 more of PR #619’s 10 failures to green.

PR #622, opened 2026-07-18
fix: retry critical Cashu storage writes

9qeklajc:
Stacks on #621 and retries every money-path write so 2 more of PR #619’s 10 failures move to green.


9qeklajc:
Independent review is clean after idempotency checks, moving 2 more of PR #619’s 10 failures to green.

PR #623, opened 2026-07-18
fix: fail safely on streaming billing errors

9qeklajc:
Stacks on #620 and makes billing cleanup idempotent so the final 4 of PR #619’s 10 red tests move to green.

### DIFF
--- a/migrations/versions/a9bc1d633fa0_add_reservation_release_idempotency_.py
+++ b/migrations/versions/a9bc1d633fa0_add_reservation_release_idempotency_.py
@@ -1,8 +1,8 @@
 """add reservation release idempotency records
 
-Revision ID: ac10fd366795
+Revision ID: a9bc1d633fa0
 Revises: d7e8f9a0b1c2
-Create Date: 2026-07-22 22:24:09.482339
+Create Date: 2026-07-24 00:20:27.967658
 """
 
 from __future__ import annotations
@@ -10,7 +10,7 @@ from __future__ import annotations
 import sqlalchemy as sa
 from alembic import op
 
-revision = "ac10fd366795"
+revision = "a9bc1d633fa0"
 down_revision = "d7e8f9a0b1c2"
 branch_labels = None
 depends_on = None
--- a/routstr/proxy.py
+++ b/routstr/proxy.py
@@ -7,7 +7,13 @@ from fastapi.responses import Response, StreamingResponse
 from sqlmodel import select
 
 from .algorithm import create_model_mappings
-from .auth import pay_for_request, revert_pay_for_request, validate_bearer_key
+from .auth import (
+    ReservationSnapshot,
+    get_reservation_snapshot,
+    pay_for_request,
+    revert_pay_for_request,
+    validate_bearer_key,
+)
 from .core import get_logger
 from .core.db import (
     ApiKey,
@@ -433,8 +439,10 @@ async def proxy(
             "upstream_error", "All upstreams failed", 502, request=request
         )
 
+    reservation_snapshot: ReservationSnapshot | None = None
     if is_ehbp or request_body_dict:
         await pay_for_request(key, max_cost_for_model, session)
+        reservation_snapshot = await get_reservation_snapshot(key, session)
 
     # Tracks request params already removed in response to upstream rejections,
     # shared across providers so a stripped param stays stripped on failover and
@@ -468,6 +476,7 @@ async def proxy(
                             max_cost_for_model=max_cost_for_model,
                             session=session,
                             model_obj=model_obj,
+                            reservation_snapshot=reservation_snapshot,
                         )
                     elif is_responses_api:
                         response = await upstream.forward_responses_request(
@@ -490,6 +499,7 @@ async def proxy(
                             max_cost_for_model,
                             session,
                             model_obj,
+                            reservation_snapshot,
                         )
                 except UpstreamError:
                     # Let the outer UpstreamError handler manage retry/revert
@@ -506,7 +516,9 @@ async def proxy(
                             "max_cost_for_model": max_cost_for_model,
                         },
                     )
-                    await revert_pay_for_request(key, session, max_cost_for_model)
+                    await revert_pay_for_request(
+                        key, session, max_cost_for_model, reservation_snapshot
+                    )
                     raise
 
                 # Reactive recovery: some models reject one specific request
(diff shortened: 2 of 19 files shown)


9qeklajc:
Independent review completed; combined validation now passes all 26 PR #619 tests, including the 10 red tests.

routstr/upstream/base.py
```diff
@@ -1435,6 +1465,7 @@ def _process_event(
                                 max_cost_for_model,
                                 model_obj,
                                 self.provider_fee,
+                                reservation_snapshot,
                             )
                             usage_finalized = True
                         except Exception as e:
```
jeroenubbink: **Medium — the streaming Responses handler still has the zero-cost bug this PR fixes for chat.**

On any finalization error this still catches `except Exception`, logs at ERROR, and hardcodes the zero `cost_data` block just below (`total_msats: 0`, `total_usd: 0.0`), then emits it. `usage_finalized` stays `False`, so the `finally` calls `finalize_db_only`, whose own `except Exception: pass` swallows the retry — so the reservation is never released either.

Confirmed by driving this handler with the same `SQLAlchemyError("database unavailable")` that makes the chat handler (correctly) raise + release. The client receives:
```json
"usage": {"cost": 0.0, "cost_sats": 0, "remaining_balance_msats": 1000},
"cost":  {"total_msats": 0, "total_usd": 0.0}
```
No exception propagates and `release_reservation` is never called → free inference + a permanently stuck reservation. Needs the same treatment as the chat path (rollback → `release_reservation` → log CRITICAL → `raise`), plus a behavioral test.

routstr/upstream/base.py
```diff
@@ -1810,6 +1841,7 @@ async def finalize_without_usage() -> bytes | None:
                             max_cost_for_model,
                             model_obj,
                             self.provider_fee,
+                            reservation_snapshot,
```
jeroenubbink, line 1875: **Medium — the messages streaming path swallows billing errors and never fails safe.**

Two swallow sites:
- `finalize_without_usage` (lines 1848–1850) catches the error and **sets `usage_finalized = True`** — it launders a failed charge into apparent success, so the `finally` won't retry.
- The token-based branch (~line 2006) has a bare `except Exception: pass` right after `adjust_payment_for_tokens`.

Driving this handler with the same injected DB error: it emits only `data: [DONE]`, never releases the reservation, and never raises — silent free inference.

Note the guard test intended to prevent this (`test_messages_streaming_no_silent_billing_failure`) passes against this code even though the `except Exception: pass` is still present — it's a `getsource()` substring check whose split doesn't line up with this occurrence, so it gives no real protection. Suggest replacing it with a behavioral test alongside the fix, matching what the chat path now does.

routstr/upstream/base.py
```diff
@@ -1018,48 +1031,63 @@ def _process_event(
...
+                                    # Release is a terminal billing state. Do not
+                                    # enqueue finalize_db_only from the generator's
+                                    # finally block and charge this request later.
+                                    usage_finalized = True
+                                else:
+                                    logger.critical(
+                                        "Billing reservation could not be released",
+                                        extra={
+                                            "key_hash": key.hashed_key[:8] + "...",
+                                            "reserved_balance": fresh_key.reserved_balance,
```
jeroenubbink: **Low — this diagnostic will misfire on exactly the path it's meant to help debug.**

`fresh_key.reserved_balance` is read after `await session.rollback()` above. `rollback()` expires ORM attributes regardless of `expire_on_commit=False`, so this access triggers an async lazy-load and raises `MissingGreenlet` — which the surrounding `except Exception as release_error` catches and mis-logs as *"Billing reservation release failed: greenlet_spawn…"*. The `raise` still fires (no fund loss), but the intended `reserved_balance` value is never emitted. Cheap fix: log `reservation_snapshot.reserved_msats` (already in hand), or snapshot the value before the rollback.

routstr/proxy.py
```diff
@@ -638,8 +657,10 @@ async def proxy(
                     "max_cost_for_model": max_cost_for_model,
                 },
             )
-            await asyncio.shield(
-                revert_pay_for_request(key, session, max_cost_for_model)
+            # The cancellation has been caught, so complete exact cleanup in
```
jeroenubbink, line 660: **Low / question.** Dropping the `asyncio.shield` here means the cancellation-path revert can now itself be interrupted by a re-delivered cancellation. I think this is intentional and fine given the durable reservation row + stale-reservation sweep make the revert recoverable — but since it's a real semantics change on the cancel path, just confirming it's a deliberate trade-off rather than incidental to the refactor.

migrations/versions/a9bc1d633fa0_add_reservation_release_idempotency_.py
```diff
@@ -0,0 +1,62 @@
...
+Create Date: 2026-07-22 22:24:09.482339
+"""
+
+from __future__ import annotations
+
+import sqlalchemy as sa
+from alembic import op
+
+revision = "ac10fd366795"
+down_revision = "d7e8f9a0b1c2"
```
jeroenubbink: Heads-up, not a change request: this chains onto `d7e8f9a0b1c2`, the same head a couple of other in-flight PRs branch from, so whichever of them merges after the others will need a quick `down_revision` re-chain to avoid multiple alembic heads. No action needed on this PR — just flagging the merge-order dependency. (This migration itself is a clean single-purpose add-table, so it doesn't need the recreate-amended-in-place treatment requested elsewhere.)

jeroenubbink (commented):
The durable `ReservationRelease` design is a solid fix — one row per reservation with a single terminal state makes release/charge exact and idempotent across crashes, concurrent finalize, and rolling upgrades. Migration is clean (single head, symmetric), lint is green, and the new behavioral tests in `test_streaming_billing_finalization.py` are good coverage. Verified locally: the red billing tests go green and the full fast suite passes with no regressions.

One blocking concern before merge: the fail-safe finalization was only applied to `handle_streaming_chat_completion`. The **Responses** and **messages** streaming handlers still carry the exact zero-cost / silent-swallow bug this PR set out to kill. I drove all three handlers with the same injected DB error — chat now raises + releases (correct), but Responses emits `total_msats: 0` and messages emits nothing, neither releases the reservation, neither raises. Details inline. The plumbing is already there (`reservation_snapshot` is threaded into both), so it should be a small extension. Two low-severity notes and a merge-ordering heads-up are also inline.

PR #624, opened 2026-07-18
test: extract passing coverage tests from #619

9qeklajc:
## Summary

- extract the passing coverage additions from #619 into a standalone PR
- cover admin, upstream base, middleware, payment helpers, proxy, and wallet money paths
- exclude the vulnerability-reproducing RED tests and two admin password assertions that do not pass on current `main`
- isolate the wallet balance test from the module-level wallet cache

## Validation

- `uv run pytest -q` — 934 passed, 14 skipped
- `uv run ruff check <added test files>` — passed
- `uv run mypy <added test files>` — passed

Source: #619

### DIFF
--- a/tests/unit/test_coverage_admin.py
+++ b/tests/unit/test_coverage_admin.py
@@ -1,0 +1,136 @@
+"""Coverage tests for admin.py (currently 35%).
+
+Tests admin endpoints that are testable without full app setup:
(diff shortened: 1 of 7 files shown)


9qeklajc:
these will guaratee behaviour even when we refactor and clean up code

PR #627, opened 2026-07-21
fix(wallet): recover stale Cashu reservations safely

9qeklajc:
## Summary

- select only liquid, unreserved proofs when issuing Cashu refunds
- report liquid and reserved proof totals separately in wallet diagnostics
- preserve negative owner equity when the wallet is empty instead of hiding liabilities
- add a dry-run-by-default reconciliation tool for stale reserved proofs
- add a bounded retry helper that only restarts Routstr after reconciliation and a solvency check
- add regression coverage for outstanding liabilities with an empty wallet

## Incident context

A production node accumulated millions of sats in locally reserved proofs while only a few hundred sats remained liquid. Many reserved proofs were already spent at their mints, while a significant unspent subset remained locked. Refund selection consequently raised `BalanceTooLowError` despite misleading logs showing a large total proof balance.

The reconciliation script:

1. verifies both SQLite databases before doing any work
2. queries each mint's NUT-07 proof state in retrying batches
3. moves confirmed `SPENT` proofs out of the active proof table
4. preserves `UNSPENT` proofs backing tracked pending refunds
5. releases only untracked, confirmed `UNSPENT` proofs
6. marks outgoing transactions collected only when every known proof is spent
7. writes a restricted JSON audit report
8. defaults to dry-run; mutation requires `--apply`

`--apply` must only be used with the node stopped and after verified database backups.

## Tests

- targeted wallet/balance suite: **104 passed**
- full suite: **939 passed, 13 skipped, 2 failed**
  - the two failures are pre-existing admin-withdraw coverage tests that attempt to contact the currently unavailable Minibits mint (`httpx.ConnectError`)
- `git diff --check`: clean

## Operational note

The retry helper is self-disabling. It keeps the node stopped while the primary mint is unavailable or liquid assets remain below positive API balances + pending refunds + accrued fees.

### DIFF
--- a/routstr/wallet.py
+++ b/routstr/wallet.py
@@ -219,8 +219,10 @@ async def send(amount: int, unit: str, mint_url: str | None = None) -> tuple[int
     """Internal send function - returns amount and serialized token"""
     effective_mint_url = mint_url or settings.primary_mint
     wallet: Wallet = await get_wallet(effective_mint_url, unit)
-    proofs = get_proofs_per_mint_and_unit(wallet, effective_mint_url, unit)
+    all_proofs = get_proofs_per_mint_and_unit(wallet, effective_mint_url, unit)
+    proofs = [proof for proof in all_proofs if not proof.reserved]
     proofs_for_mint = sum(p.amount for p in proofs)
+    reserved_for_mint = sum(p.amount for p in all_proofs if p.reserved)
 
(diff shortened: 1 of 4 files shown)


9qeklajc:
Closing and recreating from a clean branch based on current main using the maintainer checkout and git identity.

PR #628, opened 2026-07-22
fix(wallet): recover stale Cashu reservations safely

9qeklajc:
## Summary

- select only liquid, unreserved proofs when issuing Cashu refunds
- report liquid and reserved proof totals separately in wallet diagnostics
- preserve negative owner equity when the wallet is empty instead of hiding liabilities
- add a dry-run-by-default reconciliation tool for stale reserved proofs
- add a bounded retry helper that restarts Routstr only after reconciliation and a solvency check
- add regression coverage for outstanding liabilities with an empty wallet

## Incident context

A node accumulated millions of sats in locally reserved proofs while only a few hundred sats remained liquid. Many reserved proofs were already spent at their mints, while a significant unspent subset remained locked. Refund selection consequently raised `BalanceTooLowError` despite misleading logs showing a large total proof balance.

The reconciliation script:

1. verifies both SQLite databases before doing any work
2. queries each mint's NUT-07 proof state in retrying batches
3. moves confirmed `SPENT` proofs out of the active proof table
4. preserves `UNSPENT` proofs backing tracked pending refunds
5. releases only untracked, confirmed `UNSPENT` proofs
6. marks outgoing transactions collected only when every known proof is spent
7. writes a restricted JSON audit report
8. defaults to dry-run; mutation requires `--apply`

`--apply` must only be used with the node stopped and after verified database backups.

## Tests

- targeted wallet/balance suite: **104 passed**
- full suite: **939 passed, 13 skipped, 2 failed**
  - the two failures are existing admin-withdraw tests that attempt to contact the currently unavailable Minibits mint (`httpx.ConnectError`)
- `git diff --check`: clean

## Operational note

The retry helper is self-disabling. It keeps the node stopped while the primary mint is unavailable or liquid assets remain below positive API balances + pending refunds + accrued fees.


routstr/wallet.py
```diff
@@ -219,8 +219,10 @@ async def send(amount: int, unit: str, mint_url: str | None = None) -> tuple[int
     """Internal send function - returns amount and serialized token"""
     effective_mint_url = mint_url or settings.primary_mint
     wallet: Wallet = await get_wallet(effective_mint_url, unit)
-    proofs = get_proofs_per_mint_and_unit(wallet, effective_mint_url, unit)
+    all_proofs = get_proofs_per_mint_and_unit(wallet, effective_mint_url, unit)
+    proofs = [proof for proof in all_proofs if not proof.reserved]
```
jeroenubbink, line 223: Nit: `get_proofs_per_mint_and_unit` already takes `not_reserved=True` (used in `fetch_balance`). The manual filter is justified here because you also need `all_proofs` for the `reserved_for_mint` log line — but a one-line comment on *why* the fallback threshold must count liquid-only would stop a future "simplification" from folding reserved proofs back into `proofs_for_mint` and quietly reintroducing this bug.

routstr/wallet.py
```diff
@@ -219,8 +219,10 @@ async def send(amount: int, unit: str, mint_url: str | None = None) -> tuple[int
...
     proofs_for_mint = sum(p.amount for p in proofs)
```
jeroenubbink, line 226: This is where the behaviour actually changes. `select_to_send` already excludes reserved proofs internally (via `active_proofs` → `not p.reserved`), so filtering here doesn't change *which* proofs get sent — it changes `proofs_for_mint`, which gates the fallback a few lines down. Before this, a non-primary mint with e.g. 500 liquid + 600 stale-reserved and a 1000-sat refund computed `1100 >= 1000`, skipped the fallback, then hit `BalanceTooLowError` on selection. Now it correctly falls back to the primary mint. Good fix.

The gap: this path has no test. `send`/`send_token` are mocked in every test, and `test_stale_reservations` covers the api_key `reserved_balance`, not Cashu proofs — so a regression that counted reserved proofs in the threshold again would stay green. Could you add a unit test with a mixed reserved/liquid inventory at a non-primary mint that asserts the fallback fires (and that a fully-reserved primary still raises)? Happy to hand you one if it helps.

scripts/reconcile_reserved_proofs.py
```diff
@@ -0,0 +1,235 @@
...
+    batch_size: int,
+) -> dict[str, str]:
+    states: dict[str, str] = {}
+    endpoint = mint_url.rstrip("/") + "/v1/checkstate"
+    for offset in range(0, len(proofs), batch_size):
+        batch = proofs[offset : offset + batch_size]
+        last_error: Exception | None = None
+        for attempt in range(4):
+            try:
+                response = await client.post(endpoint, json={"Ys": [proof.Y for proof in batch]})
```
jeroenubbink: `proof.Y` is computed inside this `try`, but a malformed secret raises `ValueError`, not `Timeout`/`TransportError`, so it isn't caught by the retry `except` and propagates up to the per-mint `except Exception` in `run()`. That records the whole `(mint, unit)` group as errored and skips *all* its healthy proofs too. Worth validating/skipping individual bad proofs so one poison row can't block a mint's reconciliation.

scripts/reconcile_reserved_proofs.py
```diff
@@ -0,0 +1,235 @@
...
+        "keys": len(negative_rows),
+        "amount_msat": sum(row["balance"] for row in negative_rows),
+        "key_prefixes": [row["hashed_key"][:12] for row in negative_rows],
+    }
+    if args.apply:
+        keys.execute("UPDATE api_keys SET balance = 0 WHERE balance < 0")
+
+    if args.apply:
+        wallet.commit()
+        keys.commit()
```
jeroenubbink: Minor: the two connections commit sequentially, so if `keys.commit()` throws after `wallet.commit()` has already succeeded you get a partial reconciliation (wallet mutated, keys not). Low-probability for a one-shot offline tool, but since the failure mode is a wallet/keys divergence, worth either noting in the docstring or logging loudly so the operator knows a re-run is needed.

jeroenubbink (commented):
Reviewed both halves — the wallet.py runtime change and the reconciliation tooling. The logic is correct: I traced the reserved-proof filter against cashu internals, confirmed the owner_balance number is display-only, and ran the touched suites (fetch_all_balances + the send/refund/wallet set, all green). The script is appropriately offline and dry-run-first.

One thing I'd like addressed before merge: a test for the `send()` fallback behaviour (details inline). The rest are minor notes. Nice catch on the owner_balance zero-masking hiding real liabilities.

jeroenubbink:
In case it was unclear from clanker slop: It is ok as is, feel free to have a look at the commented sections if you think they are worth it.

PR #629, opened 2026-07-22
fix: bill and forward failover requests as the provider that actually served

jeroenubbink:
## Summary

When a request fails over from its best-ranked provider to a fallback, the node was still forwarding, echoing, and **billing** it as the *routing winner* rather than the provider that actually served. This PR threads the served candidate's identity (model + provider fee) through the whole settlement path so a failover serve is priced, forwarded, and refunded at the serving provider's rate — and it re-sizes the balance reservation to the serving candidate's envelope so admission stays honest.

## The problem

Model resolution collapses aliases to a single best-ranked `Model`, and settlement re-derived pricing from the response's `model` string back through that same alias map. So when two providers expose the same model tail at different prices and the winner 502s:

- the fallback was asked to serve the **winner's** model id;
- the response was billed at the **winner's** per-token pricing and provider fee, not the one that served;
- with identical model ids across providers, the mispricing was completely invisible on the wire — only the settled amount was wrong;
- the reservation was sized once to the winner's max cost, so a failover to a pricier candidate could settle beyond the admitted envelope and eat balance reserved by other in-flight requests.

## What changed

- **Thread the served model into settlement (bearer + x-cashu).** The alias map now holds `(model, provider)` candidate pairs; the proxy rebinds both per failover attempt; forwarding, max-cost echo, and settlement all use the candidate actually being tried. `adjust_payment_for_tokens` / `get_x_cashu_cost` / `calculate_cost` accept the routed `Model` and bill its pricing directly. On a failover serve the response's `model` field now names the serving candidate.
- **Bill the serving provider's fee on the USD-cost path.** The USD-cost and litellm fallback paths resolved the fee via the best-ranked provider for the alias; settlement callers now pass their own `provider_fee` through. Configured model pricing is unaffected (the fee is already baked into cached pricing).
- **Re-reserve the serving candidate's envelope on failover.** Before trying a fallback, raise the reservation to its own max cost; reject candidates the key can't cover — exactly as admission would have, had that candidate been ranked first.
- **Version-suffix routing fix.** The unified candidate lookup now applies the `-YYYYMMDD` strip that model resolution already had, so version-suffixed requests no longer resolve a model yet 400 with "no provider found".
- **Type-safety at the seams.** `model_obj` / `provider_fee` are required (nullable) on the settlement entry points, so a call site that forgets to thread the served candidate is a type error rather than a silent fallback to alias re-derivation. `calculate_cost` keeps its defaults as the one documented fallback seam.
- **Cleanup.** Dropped the now-dead write-only model-instance alias map; fixed a concurrent `patch()`-leak in the overrun finalization tests that was leaking a mock into later tests.

## Testing

- New `tests/integration/test_failover_billing.py`: fallback serve billed at the fallback rate; identical-id failover settles at the serving price (the wire-invisible case); USD-cost path carries the serving fee; version-suffixed id routes; over-envelope fallback rejected pre-serve; affordable pricier fallback re-reserved, served, and refunded.
- New `tests/unit/test_settlement_identity.py` for the threading contract at the billing seams.
- `make lint`, `make test-fast`, and integration suite green.

### DIFF
--- a/routstr/algorithm.py
+++ b/routstr/algorithm.py
@@ -89,7 +89,9 @@ def create_model_mappings(
     overrides_by_key: dict[tuple[str, int], tuple],
     disabled_model_keys: set[tuple[str, int]],
 ) -> tuple[
-    dict[str, "Model"], dict[str, list["BaseUpstreamProvider"]], dict[str, "Model"]
+    dict[str, "Model"],
+    dict[str, list[tuple["Model", "BaseUpstreamProvider"]]],
+    dict[str, "Model"],
 ]:
     """Create optimal model mappings based on cost and provider preferences.
 
@@ -97,7 +99,9 @@ def create_model_mappings(
     and creates three mappings based on cost optimization:
 
     1. model_instances: alias -> Model (all model aliases mapped to their Model objects)
-    2. provider_map: alias -> List[UpstreamProvider] (sorted list of providers for each alias)
+    2. provider_map: alias -> List[(Model, UpstreamProvider)] (sorted candidate
+       list for each alias; each provider is paired with ITS OWN model so
+       failover can forward and bill the candidate that actually serves)
     3. unique_models: base_id -> Model (unique models without provider prefixes)
 
     The algorithm:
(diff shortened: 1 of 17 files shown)


9qeklajc:
LGTM.. I would refactor this later to enforce that model_id and provider_fee couldn't be None ..

PR #630, opened 2026-07-24
fix: stop sharing AsyncSession across tasks and release DB connections during external I/O

9qeklajc:
## Summary

Fixes SQLAlchemy connection pool exhaustion (`QueuePool limit of size 5 overflow 10 reached`) and `concurrent operations are not permitted` errors that degraded the node until a container restart.

### Root cause
- `fetch_all_balances()` shared a single `AsyncSession` across `asyncio.gather()` tasks — `AsyncSession` is not safe for concurrent use.
- Several paths held DB sessions open across slow external mint/LNURL calls (invoice checks, periodic payout, refund sweep, fee payout), pinning pool connections.
- Streaming proxy responses kept the request-scoped session alive for the full stream duration.

### Changes
- **`fetch_all_balances` / `periodic_payout`**: fetch all user liabilities in one short `GROUP BY` query (`balances_by_mint_and_unit`), close the session, then run mint checks — no session sharing, no session held during external I/O. Mint fan-out bounded by `MINT_OPERATION_CONCURRENCY` semaphore.
- **`check_invoice_payment`**: commit before the slow mint call so it cannot pin a connection; credit + conditional `UPDATE ... WHERE status='pending'` finalize in one atomic transaction (lost race rolls back the credit; duplicate mints rejected by the mint's single-use quotes). No schema change needed.
- **`_refund_sweep_once`**: read candidates in one session, then per-item short sessions with idempotent conditional updates around the external redeem.
- **`periodic_routstr_fee_payout`**: checkpoint in a short session, perform external send outside it, complete checkpoint in a fresh session.
- **`proxy()`**: explicitly close the request session before returning, so long LLM streams no longer hold DB connections.
- **Pool config**: `DATABASE_POOL_SIZE / MAX_OVERFLOW / POOL_TIMEOUT / POOL_RECYCLE / POOL_PRE_PING` env vars (in-memory SQLite keeps its pool untouched); checkout/checkin listeners warn when a connection is held longer than `DATABASE_POOL_HOLD_WARN_SECONDS`. All settings pass through unchanged to a future postgres/asyncpg engine.

### Review follow-ups
- **`periodic_payout` stale liability**: liability is now fetched in a short session per mint/unit right before computing the available balance — a concurrent top-up can only shrink the payout, never send customer-backed funds as profit.
- **Refund sweep race**: each refund is atomically claimed (`swept=True WHERE swept=false AND collected=false`) before redemption; failures release the claim for retry, so concurrent sweeps can no longer misreport a sweep as client-collected.
- **Cancellation safety**: `check_invoice_payment` catches `BaseException`, so task cancellation after a successful mint still emits the reconciliation alert.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -22,6 +22,13 @@ ROUTSTR_SECRET_KEY=
 
 # Database
 # DATABASE_URL=sqlite+aiosqlite:///keys.db
+# Keep total pool capacity across all workers below the database connection limit.
+# DATABASE_POOL_SIZE=5
+# DATABASE_MAX_OVERFLOW=0
+# DATABASE_POOL_TIMEOUT=5
+# DATABASE_POOL_RECYCLE=1800
+# DATABASE_POOL_PRE_PING=true
+# DATABASE_POOL_HOLD_WARN_SECONDS=10
 
 # Node Information
 # NAME=My Routstr Node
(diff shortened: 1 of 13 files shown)


9qeklajc:
close in favor of https://github.com/Routstr/routstr-core/pull/634

Issue #611, opened 2026-07-13
Prevent payouts from consuming refund liabilities and handle uncertain refunds safely


9qeklajc:
## Summary

The payout and refund paths do not currently share one authoritative solvency check. This can let operator withdrawals consume funds needed to refund users, and ambiguous payment outcomes may restore a user's balance even when the remote payment actually succeeded.

## Confirmed and suspected problems

- **Confirmed over-withdraw risk:** `POST /admin/withdraw` validates against all unreserved wallet proofs rather than owner surplus. The withdrawal UI also permits using user-backed funds.
- **Confirmed payout bypass:** Routstr fee payout spends from the primary wallet without first protecting customer refund liabilities.
- **High-confidence over-refund risk:** the refund flow restores the API-key balance after any exception. An LNURL/Cashu melt can succeed remotely but time out locally, causing the balance to be restored and a retry to potentially pay twice.
- Periodic payout, manual withdrawal, refunds, and fee payout use separate non-transactional calculations and process-local coordination.
- Existing tests do not cover solvency across these paths, independent SQLite connection races, durable idempotency, or paid-but-timed-out reconciliation.

## Expected behavior

- Calculate protected refund liability and withdrawable owner surplus per `(mint_url, unit)`.
- Permit manual and automatic payouts only from server-calculated withdrawable surplus.
- Preserve enough assets to refund all users, including while requests or outbound refunds are pending.
- Represent outbound operations durably with idempotency and atomic claims.
- Distinguish `rejected`, terminal `failed`, `pending_confirmation`, and `paid/issued` outcomes.
- Never restore a balance or retry as a new payment while the external result is uncertain.
- Show only the actually withdrawable amount in the UI; separately display wallet assets, refund liability, pending outbound funds, deficit, and withdrawable surplus.

## Likely implementation areas

- `routstr/core/admin.py` — withdrawal limit and structured statuses/errors
- `routstr/balance.py` — durable refund liability handoff and guarded restoration
- `routstr/payment/lnurl.py` — pending/unknown melt outcomes and reconciliation
- `routstr/wallet.py` — shared accounting for periodic, admin, and fee payouts
- `routstr/core/db.py` plus a migration — durable outbound-operation state and idempotency
- `ui/components/withdraw-modal.tsx` and wallet API/types — withdrawable-only UX

## Acceptance criteria

- [ ] No operator payout path can claim more than stage-correct withdrawable surplus.
[...]

PR #632, opened 2026-07-24
fix: stop DB connection-pool exhaustion in balance reads (+ env-configurable pool sizing)

jeroenubbink:
## Problem

Under load, admin/API endpoints start returning 500/502 and the logs fill with SQLAlchemy connection-pool exhaustion:

```
sqlalchemy.exc.TimeoutError: QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00
```

often alongside:

```
This session is provisioning a new connection; concurrent operations are not permitted
```

Once it starts, background tasks (invoice watcher, dead-key prune, stale-reservation sweep) and the balances page all fail with the same `QueuePool limit ... reached` error, and only a container restart clears the wedged connections — until it recurs.

## Root cause

`fetch_all_balances()` created **one** `AsyncSession` and shared it across the tasks it ran with `asyncio.gather`:

```python
async with db.create_session() as session:
    tasks = [
        fetch_balance(session, mint_url, unit)   # each awaits a query on `session`
        for mint_url in mint_urls
        for unit in units
    ]
    balance_details = list(await asyncio.gather(*tasks))
```

`AsyncSession` is **not safe for concurrent use**. Overlapping queries on the shared session raise *"concurrent operations are not permitted"* and leave the underlying connection checked-out and wedged. Repeated over time, the `QueuePool` (5 + 10 overflow, 30s timeout) is drained and every DB path 500s.

## What this PR does

Two independent, separately-reviewable commits.

### 1. `fix:` stop AsyncSession pool exhaustion in balance reads

- **One up-front grouped read.** All outstanding user liabilities are read once, in a single short-lived session with a single `GROUP BY` query (`balances_by_mint_and_unit`), *before* the concurrent fan-out. The per-mint balance checks then run concurrently with **no DB session in scope**, so nothing shares a session across `gather`.
- **Graceful degradation.** If the liability read fails, the balances page no longer 500s. It still reports the known **wallet custody** and blanks only the unknowable per-user/owner split, tagging each mint with the error. The wallet total is reported truthfully; `owner` is reported as `0` so custody is never mis-reported as owner profit; and a specific per-mint failure message is preserved rather than being overwritten by the generic liability-read error.
- **Payout money-safety.** `periodic_payout` now reads each liability **fresh, immediately before the payout decision**, instead of from a single pre-loop snapshot. The per-mint round trip is slow (a `sleep(5)` plus mint I/O per iteration); a user top-up during the cycle would otherwise let a later mint/unit compute `available_balance = fresh_proofs − stale_liability` and **over-send funds owed to users**. (Related to the payout-safety concern in issue #611.)

### 2. `feat:` make the DB connection pool env-configurable

- Adds `DATABASE_POOL_SIZE` / `DATABASE_MAX_OVERFLOW` / `DATABASE_POOL_TIMEOUT`, consumed like every other typed env var through the pydantic `Settings` (constrained `Field`s; defaults **5 / 10 / 30** match SQLAlchemy's own baseline, so leaving them unset is behaviour-neutral).
- **Fail-loud on bad config.** An out-of-range or non-integer value fails validation and **refuses to boot** — the same behaviour a malformed `DATABASE_URL` already has — instead of silently starting up misconfigured.
- `create_db_engine` sizes the pool from these values and **logs the effective config at startup** (`Database pool configured: ...`) so it can be confirmed from the boot output during an incident. In-memory SQLite uses `StaticPool` (which rejects the pool kwargs) and is detected and built without them.
- **`pool_pre_ping` is intentionally not exposed.** The default backend is a local SQLite file with no network peer to drop idle connections, so it would add a `SELECT 1` per checkout for no benefit — and it detects *dead* connections, not the live-but-wedged ones behind this exhaustion.
- **Env-only, never DB-configured.** These knobs are infrastructure the node needs *before* it can open a DB session, so they can never be sourced from the DB (chicken-and-egg). A new `ENV_ONLY_FIELDS` set keeps them out of the persisted settings blob and stops a DB value from shadowing env in both `SettingsService.initialize` and `.update` — env stays authoritative.
- `.env.example` gains a documented **"Database tuning"** section (defaults, the exact `QueuePool` symptom to watch for, the boot log line, the fail-loud note, and the caveat that SQLite serialises writes so a bigger pool only helps concurrent readers).

## Testing

- `balances_by_mint_and_unit`: real-DB coverage for grouping/summing, mint+unit filtering, NULL-key exclusion, and empty inputs.
- `fetch_all_balances`: a session-concurrency guard asserting the liabilities are read in exactly one session with one query and never shared across the fan-out; graceful-degradation asserting custody is still reported while the user/owner split is blanked; and that a specific mint error survives a concurrent liability-read failure.
- `periodic_payout`: asserts the liability is read fresh once per `(mint, unit)`, after the inner sleep — not from a pre-loop snapshot.
- Pool config: defaults match the baseline; overrides flow through; in-memory SQLite skips the kwargs; pre-ping stays off; the effective config is logged; and non-integer / out-of-range / negative values are rejected at boot (`max_overflow=0` is allowed).
- Settings: the pool fields are never written to the blob and a stale/injected blob value never shadows env, on both `initialize` and `update`.

`make lint` clean (ruff + mypy). `make test-fast`: 1063 passed, 10 skipped.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -23,6 +23,38 @@ ROUTSTR_SECRET_KEY=
 # Database
 # DATABASE_URL=sqlite+aiosqlite:///keys.db
 
+# Database tuning (advanced — you almost certainly do not need these)
+# ---------------------------------------------------------------------------
+# These size the SQLAlchemy connection pool. The defaults below match
+# SQLAlchemy's own baseline, so leaving them unset changes nothing. Touch them
+# ONLY if the logs show the pool running out of connections:
+#
+#     QueuePool limit of size 5 overflow 10 reached, connection timed out, timeout 30.00
+#
+# The effective values are printed at startup ("Database pool configured: ...")
+# so you can confirm what is actually in effect while diagnosing an incident.
+# An invalid value (non-integer, or a pool size below 1) stops the node at boot
+# with a clear error rather than starting up misconfigured.
+#
+# DATABASE_POOL_SIZE      Connections kept open in the pool.        (default 5)
+# DATABASE_MAX_OVERFLOW   Extra connections opened under load,      (default 10)
+#                         beyond pool_size, then discarded.
+# DATABASE_POOL_TIMEOUT   Seconds a request waits for a free        (default 30)
+#                         connection before failing with the error above.
+#
+# Note for the default SQLite backend: SQLite serialises writes behind a single
+# database-level lock, so a bigger pool does NOT buy more write throughput — it
+# only lets more readers run at once (WAL mode) and trades pool-timeout errors
+# for "database is locked" errors. If you exhaust the pool on SQLite, the cause
+# is almost always connections being held too long, not the pool being too
+# small. These knobs come into their own if you point DATABASE_URL at a
+# networked database (e.g. Postgres), where connections genuinely run in
+# parallel; there, size the pool against the server's max_connections.
+#
+# DATABASE_POOL_SIZE=5
+# DATABASE_MAX_OVERFLOW=10
+# DATABASE_POOL_TIMEOUT=30
+
 # Node Information
 # NAME=My Routstr Node
 # DESCRIPTION=Fast AI API access with Bitcoin payments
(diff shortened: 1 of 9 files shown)


9qeklajc:
combined here https://github.com/Routstr/routstr-core/pull/634

9qeklajc:
LGTM

jeroenubbink:
closed due to this being merged into #634

PR #634, opened 2026-07-24
fix: combine DB pool-exhaustion and session-lifecycle fixes

9qeklajc:
## Summary

This supersedes the implementation approach in #630 and #632 while preserving both PR histories as merge parents.

It uses #630's broader connection-lifecycle work as the base and integrates #632's stronger typed configuration and balance-query behavior.

## Combined behavior

- Stops sharing `AsyncSession` across concurrent balance checks.
- Keeps DB sessions short around mint, LNURL, refund, invoice, and streaming I/O.
- Bounds mint-side concurrency.
- Uses a filtered grouped liability query.
- Degrades balance reporting safely when liabilities cannot be read: custody remains visible, but unknown user/owner splits are never presented as owner profit.
- Uses validated, environment-only pool controls while retaining pre-ping, recycle, and long-checkout diagnostics.
- Logs effective pool configuration at startup.
- Preserves fail-fast bounded pool defaults.

## Additional fixes found while combining

- Refund sweeps now use a recoverable claim lease (`sweep_started_at`). Cancellation releases the lease, and stale claims can be retried after a crash.
- Fee payout prepares the wallet and proofs before creating the durable payout checkpoint, so a preparation failure cannot strand a payout that was never attempted.
- A lost concurrent fee-payout claim never sends funds.

## History

The branch contains explicit merge commits for both PR heads:

- #630: `2410a4a6ce0729a6e9f0bb471cb17a51adb4a7cf`
- #632: `a2cedd676934fc73e405c0d9aee57ff1d736b845`

Both original PRs are intentionally left open for comparison.

## Validation

- `uv run pytest -q` — **1078 passed, 14 skipped**
- `uv run mypy routstr` — **no issues**
- `uv run ruff check .` — **passed**
- Fresh SQLite Alembic upgrade reaches `8a1b2c3d4e5f (head)`.

## Known limitation

When reclaiming a stale refund-sweep lease, an `already spent` response cannot distinguish a prior worker that redeemed before crashing from a client that collected during the stale window. The code conservatively records the ambiguous outcome as swept; funds are not at risk, but the audit classification may be imperfect.

### DIFF
--- a/routstr/core/settings.py
+++ b/routstr/core/settings.py
@@ -100,6 +100,14 @@ class Settings(BaseSettings):
     refund_cache_ttl_seconds: int = Field(default=3600, env="REFUND_CACHE_TTL_SECONDS")
     refund_sweep_ttl_seconds: int = Field(default=604800, env="REFUND_SWEEP_TTL_SECONDS")
 
+    # Database connection-pool sizing (advanced). Defaults match SQLAlchemy's own
+    # baseline, so unset is behaviour-neutral — see the "Database tuning" section
+    # of .env.example. A pool_size of 0 would mean an unbounded pool in
+    # SQLAlchemy, so it is rejected (ge=1); max_overflow=0 (no overflow) is valid.
+    database_pool_size: int = Field(default=5, ge=1, env="DATABASE_POOL_SIZE")
+    database_max_overflow: int = Field(default=10, ge=0, env="DATABASE_MAX_OVERFLOW")
+    database_pool_timeout: int = Field(default=30, ge=1, env="DATABASE_POOL_TIMEOUT")
+
     # Logging
     log_level: str = Field(default="INFO", env="LOG_LEVEL")
     enable_console_logging: bool = Field(default=True, env="ENABLE_CONSOLE_LOGGING")
@@ -144,10 +152,25 @@ def _normalize_settings_data(data: dict[str, Any]) -> dict[str, Any]:
 # ``routstr.core.vault``.
 SECRET_FIELDS = frozenset({"admin_password", "nsec"})
 
+# Infrastructure the node needs *before* it can open a DB session — so it can
+# never be configured from the DB (chicken-and-egg) and stays env-only. Unlike
+# secrets (owned by bootstrap), these are excluded so the DB settings blob can
+# neither store nor shadow them; env is always authoritative.
+ENV_ONLY_FIELDS = frozenset(
+    {"database_pool_size", "database_max_overflow", "database_pool_timeout"}
+)
+
+_NON_PERSISTED_FIELDS = SECRET_FIELDS | ENV_ONLY_FIELDS
+
 
 def _strip_secret_fields(data: dict[str, Any]) -> dict[str, Any]:
-    """Return a copy of ``data`` without any secret fields (for persistence)."""
-    return {k: v for k, v in data.items() if k not in SECRET_FIELDS}
+    """Return a copy of ``data`` without secret or env-only fields.
+
+    Both are kept out of the persisted settings blob: secrets for confidentiality,
+    env-only fields (e.g. DB pool sizing) because they must never be sourced from
+    the database.
+    """
+    return {k: v for k, v in data.items() if k not in _NON_PERSISTED_FIELDS}
 
 
 def _apply_to_live_settings(data: dict[str, Any]) -> None:
@@ -327,7 +350,13 @@ class SettingsService:
             valid_fields = set(env_resolved.dict().keys())
             merged_dict: dict[str, Any] = dict(env_resolved.dict())
             merged_dict.update(
-                {k: v for k, v in db_json.items() if v not in (None, "", [], {}) and k in valid_fields}
+                {
+                    k: v
+                    for k, v in db_json.items()
+                    if v not in (None, "", [], {})
+                    and k in valid_fields
+                    and k not in ENV_ONLY_FIELDS
+                }
             )
             merged_dict = Settings(**merged_dict).dict()
 
@@ -394,8 +423,13 @@ class SettingsService:
                 )
             )
             await db_session.commit()
-            # Update in-place
+            # Update in-place. Env-only fields (e.g. DB pool sizing) are never
+            # applied here: the engine pool is already built at boot from env,
+            # so letting an update mutate the live value would only make it
+            # diverge from the running pool.
             for k, v in candidate.dict().items():
+                if k in ENV_ONLY_FIELDS:
+                    continue
                 setattr(settings, k, v)
             cls._current = settings
             return settings
--- a/routstr/core/settings.py
+++ b/routstr/core/settings.py
@@ -41,6 +41,9 @@ class Settings(BaseSettings):
     receive_ln_address: str = Field(default="", env="RECEIVE_LN_ADDRESS")
     primary_mint: str = Field(default="", env="PRIMARY_MINT_URL")
     primary_mint_unit: str = Field(default="sat", env="PRIMARY_MINT_UNIT")
+    mint_operation_concurrency: int = Field(
+        default=4, ge=1, env="MINT_OPERATION_CONCURRENCY"
+    )
 
     # Lightning payout configuration
     # Minimum available balance (in satoshis) before profit is paid out over
--- a/routstr/core/settings.py
+++ b/routstr/core/settings.py
@@ -106,14 +106,17 @@ class Settings(BaseSettings):
         default=900, gt=0, env="REFUND_SWEEP_CLAIM_TIMEOUT_SECONDS"
     )
 
-    # Database connection-pool controls (advanced). Keep the default pool
-    # bounded and fail quickly: increasing it can hide leaked connections and
-    # makes SQLite lock contention worse. These fields are env-only below.
+    # Database connection-pool controls (advanced). Capacity defaults match
+    # SQLAlchemy's established queue-pool behavior. Pre-ping is enabled by the
+    # engine factory for networked backends; SQLite can explicitly opt in.
+    # These fields are env-only below.
     database_pool_size: int = Field(default=5, ge=1, env="DATABASE_POOL_SIZE")
-    database_max_overflow: int = Field(default=0, ge=0, env="DATABASE_MAX_OVERFLOW")
-    database_pool_timeout: float = Field(default=5.0, gt=0, env="DATABASE_POOL_TIMEOUT")
+    database_max_overflow: int = Field(default=10, ge=0, env="DATABASE_MAX_OVERFLOW")
+    database_pool_timeout: float = Field(
+        default=30.0, gt=0, env="DATABASE_POOL_TIMEOUT"
+    )
     database_pool_recycle: int = Field(default=1800, ge=0, env="DATABASE_POOL_RECYCLE")
-    database_pool_pre_ping: bool = Field(default=True, env="DATABASE_POOL_PRE_PING")
+    database_pool_pre_ping: bool = Field(default=False, env="DATABASE_POOL_PRE_PING")
     database_pool_hold_warn_seconds: float = Field(
         default=10.0, gt=0, env="DATABASE_POOL_HOLD_WARN_SECONDS"
     )
--- a/routstr/core/settings.py
+++ b/routstr/core/settings.py
@@ -102,6 +102,21 @@ class Settings(BaseSettings):
     enable_models_refresh: bool = Field(default=True, env="ENABLE_MODELS_REFRESH")
     refund_cache_ttl_seconds: int = Field(default=3600, env="REFUND_CACHE_TTL_SECONDS")
     refund_sweep_ttl_seconds: int = Field(default=604800, env="REFUND_SWEEP_TTL_SECONDS")
+    refund_sweep_claim_timeout_seconds: int = Field(
+        default=900, gt=0, env="REFUND_SWEEP_CLAIM_TIMEOUT_SECONDS"
+    )
+
+    # Database connection-pool controls (advanced). Keep the default pool
+    # bounded and fail quickly: increasing it can hide leaked connections and
+    # makes SQLite lock contention worse. These fields are env-only below.
+    database_pool_size: int = Field(default=5, ge=1, env="DATABASE_POOL_SIZE")
+    database_max_overflow: int = Field(default=0, ge=0, env="DATABASE_MAX_OVERFLOW")
+    database_pool_timeout: float = Field(default=5.0, gt=0, env="DATABASE_POOL_TIMEOUT")
+    database_pool_recycle: int = Field(default=1800, ge=0, env="DATABASE_POOL_RECYCLE")
+    database_pool_pre_ping: bool = Field(default=True, env="DATABASE_POOL_PRE_PING")
+    database_pool_hold_warn_seconds: float = Field(
+        default=10.0, gt=0, env="DATABASE_POOL_HOLD_WARN_SECONDS"
+    )
 
     # Logging
     log_level: str = Field(default="INFO", env="LOG_LEVEL")
@@ -147,10 +162,32 @@ def _normalize_settings_data(data: dict[str, Any]) -> dict[str, Any]:
 # ``routstr.core.vault``.
 SECRET_FIELDS = frozenset({"admin_password", "nsec"})
 
+# Infrastructure the node needs *before* it can open a DB session — so it can
+# never be configured from the DB (chicken-and-egg) and stays env-only. Unlike
+# secrets (owned by bootstrap), these are excluded so the DB settings blob can
+# neither store nor shadow them; env is always authoritative.
+ENV_ONLY_FIELDS = frozenset(
+    {
+        "database_pool_size",
+        "database_max_overflow",
+        "database_pool_timeout",
+        "database_pool_recycle",
+        "database_pool_pre_ping",
+        "database_pool_hold_warn_seconds",
+    }
+)
+
+_NON_PERSISTED_FIELDS = SECRET_FIELDS | ENV_ONLY_FIELDS
+
 
 def _strip_secret_fields(data: dict[str, Any]) -> dict[str, Any]:
-    """Return a copy of ``data`` without any secret fields (for persistence)."""
-    return {k: v for k, v in data.items() if k not in SECRET_FIELDS}
+    """Return a copy of ``data`` without secret or env-only fields.
+
+    Both are kept out of the persisted settings blob: secrets for confidentiality,
+    env-only fields (e.g. DB pool sizing) because they must never be sourced from
+    the database.
+    """
+    return {k: v for k, v in data.items() if k not in _NON_PERSISTED_FIELDS}
 
 
 def _apply_to_live_settings(data: dict[str, Any]) -> None:
@@ -330,7 +367,13 @@ class SettingsService:
             valid_fields = set(env_resolved.dict().keys())
             merged_dict: dict[str, Any] = dict(env_resolved.dict())
             merged_dict.update(
-                {k: v for k, v in db_json.items() if v not in (None, "", [], {}) and k in valid_fields}
+                {
+                    k: v
+                    for k, v in db_json.items()
+                    if v not in (None, "", [], {})
+                    and k in valid_fields
+                    and k not in ENV_ONLY_FIELDS
+                }
             )
             merged_dict = Settings(**merged_dict).dict()
 
@@ -397,8 +440,13 @@ class SettingsService:
                 )
             )
             await db_session.commit()
-            # Update in-place
+            # Update in-place. Env-only fields (e.g. DB pool sizing) are never
+            # applied here: the engine pool is already built at boot from env,
+            # so letting an update mutate the live value would only make it
+            # diverge from the running pool.
             for k, v in candidate.dict().items():
+                if k in ENV_ONLY_FIELDS:
+                    continue
                 setattr(settings, k, v)
             cls._current = settings
             return settings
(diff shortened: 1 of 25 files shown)


jeroenubbink (commented):
Reviewed the combined branch end-to-end — the merge is done carefully: the balance-read fan-out fix and the broader connection-lifecycle work are both preserved, the grouped-liability query + graceful degradation survive intact, and the added money-safety hardening (conditional-UPDATE invoice finalize, fee-payout prep-before-checkpoint, refund-sweep lease) is real improvement. Migrations chain onto a single head; `pytest -q` is green locally. Approval in principle.

One correctness issue I'd like resolved before merge (the borrowed-session `rollback`, inline below) — it's funds-safe but degrades the exact background task this PR touches, under normal concurrency, and isn't covered by the new tests. Everything else is non-blocking.

One more that couldn't be anchored inline (the file isn't in this diff): **`sweep_started_at` now surfaces in the admin transactions API.** Because it's a new `CashuTransaction` column, `admin.py`'s `[tx.dict() for tx in transactions]` serializes it into the `/transactions` response. It's an internal worker-lease timestamp, not transaction data — worth excluding from the serialized shape (or confirming clients tolerate the extra key).

routstr/lightning.py
```diff
@@ -222,79 +222,129 @@ async def recover_invoice(
...
+                "api_key_hash": invoice.api_key_hash[:8] + "..."
+                if invoice.api_key_hash
+                else None,
+            },
+        )
+    except BaseException as e:
+        # BaseException so task cancellation (e.g. client disconnect) after a
+        # successful mint still triggers the reconciliation alert.
+        try:
+            await asyncio.shield(session.rollback())
```
jeroenubbink: **`check_invoice_payment` rolls back a session it doesn't own, which expires the caller's ORM objects.** This function receives `session` from the caller — an HTTP request session in `get_invoice_status`/`recover_invoice`, and in `periodic_invoice_watcher` a *single* session shared across the whole 100-invoice batch. Under async SQLAlchemy (the reason `expire_on_commit=False` is set), a `rollback()` that follows a DB write expires **every** object in that session; the next attribute access — even a primary key — raises `MissingGreenlet`. I reproduced this directly.

Two reachable consequences:
- **Watcher batch (normal operation):** a client polling `/invoice/{id}/status` marks an invoice paid first; the watcher then hits `finalized.rowcount != 1` and rolls back at line 266 (after flushing the api-key insert + finalize UPDATE). That expires the *sibling* invoices still in the shared session, so the rest of the batch fails for the cycle (each logged "Invoice watcher failed") and is deferred to the next tick.
- **Endpoint 500:** the exception path here (unlike line 266) doesn't `refresh`, so a `database is locked` during finalize — plausible under the pool pressure this PR targets — leaves the caller's `invoice` expired and `get_invoice_status`/`recover_invoice` 500 on `invoice.status`.

Fix direction: don't roll back a borrowed session (let the caller own its transaction), or give the watcher a fresh session per invoice. Worth a test that drives the lost-race and a write-then-fail path through the watcher and an endpoint.

routstr/wallet.py
```diff
@@ -935,108 +942,164 @@ async def periodic_payout() -> None:
...
-                        extra={
-                            "id": refund.id,
-                            "error": str(e),
-                        },
-                    )
-        await session.commit()
+            else:
+                # Cancellation is a known-no-send outcome at this boundary, so
+                # release the lease under shield and let a later sweep retry.
+                await asyncio.shield(
```
jeroenubbink: **The `else` branch releases the claim lease even when the token was already redeemed.** It runs for any non-"already spent" exception — including one raised by `_set_refund_sweep_state(swept=True, …)` *after* `recieve_token` already swept the funds into the node wallet. Releasing `sweep_started_at=None` then loses the "mid-sweep" signal: the next run re-claims, gets "already spent", but `reclaimed_stale_claim` is now `False`, so it books the row `collected=True` — a node sweep permanently misrecorded as a client collection. Funds are safe; owner/liability accounting is wrong. This is a different direction than the known-limitation in the PR description. Distinguishing "already redeemed by us" from "cancelled before send" (only the latter should release the lease) would close it.

routstr/lightning.py
```diff
@@ -222,79 +222,129 @@ async def recover_invoice(
...
+
+        # The mint enforces single-use quotes, so a concurrent checker that
+        # races us here fails inside wallet.mint rather than double-minting.
+        await wallet.mint(invoice.amount_sats, quote_id=invoice.payment_hash)
+        minted = True
+
+        if invoice_purpose == "create":
+            api_key = await _create_api_key_record(invoice, session)
+            invoice.api_key_hash = api_key.hashed_key
+        elif invoice_purpose == "topup":
```
jeroenubbink: Minor: the mint at line 240 now runs unconditionally before purpose dispatch, and this `elif` dropped the old `and invoice.api_key_hash` guard. Reachability is low (creation enforces `api_key_hash` for topups), but if the target key is pruned between creation and payment, funds get minted and then `_credit_topup_record` raises → CRITICAL "reconciliation required", invoice stuck pending. The CRITICAL outcome is arguably better than the old silent no-op, so flagging rather than blocking — just noting the guard removal is load-bearing.

routstr/wallet.py
```diff
@@ -1083,52 +1146,58 @@ async def periodic_routstr_fee_payout() -> None:
...
+                )
+            except Exception:
+                logger.critical(
+                    "Routstr fee payout outcome is unknown; manual reconciliation required",
+                    extra={"payout_in_progress_msats": paid_msats},
+                    exc_info=True,
+                )
+                continue
+
+            async with db.create_session() as session:
```
jeroenubbink: After `raw_send_to_lnurl` succeeds, opening this fresh session for the completion checkpoint can itself fail (pool timeout / locked). That exception falls to the outer generic `except Exception`, not the specific `"sent but checkpoint not completed"` CRITICAL just below — so the "money left, checkpoint didn't" signal is lost (the `payout_in_progress` guard still surfaces it next cycle, but with the ambiguous message). Consider guarding the completion session-acquire so it still routes to the specific CRITICAL.

routstr/wallet.py
```diff
@@ -935,108 +942,164 @@ async def periodic_payout() -> None:
...
+                        proofs = get_proofs_per_mint_and_unit(
+                            wallet, mint_url, unit, not_reserved=True
+                        )
+                        proofs = await slow_filter_spend_proofs(proofs, wallet)
+                        await asyncio.sleep(5)
+                        # Fetch the liability AFTER the proofs snapshot and the
+                        # settle delay: a concurrent top-up then only inflates
+                        # the liability, shrinking the payout — never sending
+                        # customer-backed funds as profit.
+                        async with db.create_session() as session:
```
jeroenubbink, line 987: Moving `create_session()` into the per-`(mint, unit)` try means a DB-session failure now logs `"Error sending payout"` (twice) in the same bucket as mint/network errors, instead of the previous once-per-cycle `"Error in periodic payout cycle"`. Any alert keyed on the DB-specific string goes quiet. Minor, but if that string is monitored it's a silent regression.

routstr/core/settings.py
```diff
@@ -99,6 +102,21 @@ def parse_env_var(cls, field_name: str, raw_value: str) -> Any:  # type: ignore[
...
+    )
+
+    # Database connection-pool controls (advanced). Keep the default pool
+    # bounded and fail quickly: increasing it can hide leaked connections and
+    # makes SQLite lock contention worse. These fields are env-only below.
+    database_pool_size: int = Field(default=5, ge=1, env="DATABASE_POOL_SIZE")
+    database_max_overflow: int = Field(default=0, ge=0, env="DATABASE_MAX_OVERFLOW")
+    database_pool_timeout: float = Field(default=5.0, gt=0, env="DATABASE_POOL_TIMEOUT")
+    database_pool_recycle: int = Field(default=1800, ge=0, env="DATABASE_POOL_RECYCLE")
+    database_pool_pre_ping: bool = Field(default=True, env="DATABASE_POOL_PRE_PING")
```
jeroenubbink: `pool_pre_ping=True` as the default is pure overhead on the SQLite default backend. Pre-ping issues a `SELECT 1` on every checkout to catch connections severed *while idle* (a DB server's idle-timeout, an LB/firewall dropping idle TCP). The default backend is `sqlite+aiosqlite:///keys.db` — a local, in-process file with no server, socket, peer, or idle-reaper that can invalidate a connection between checkouts; barring the file being deleted underneath a running process (which `SELECT 1` wouldn't recover from), a checked-in SQLite connection is always usable next checkout. So it adds a guaranteed round-trip per checkout that can never return "stale", on the hot path. It also doesn't address this incident: exhaustion came from connections *checked out and wedged*, not *checked in and dead*, and pre-ping only inspects idle-pool connections at checkout. Keeping the field is right (it earns its cost on a networked backend), but suggest `default=False`, auto-enabled when the URL backend isn't sqlite.

routstr/core/settings.py
```diff
@@ -99,6 +102,21 @@ def parse_env_var(cls, field_name: str, raw_value: str) -> Any:  # type: ignore[
...
     refund_sweep_ttl_seconds: int = Field(default=604800, env="REFUND_SWEEP_TTL_SECONDS")
+    refund_sweep_claim_timeout_seconds: int = Field(
+        default=900, gt=0, env="REFUND_SWEEP_CLAIM_TIMEOUT_SECONDS"
+    )
+
+    # Database connection-pool controls (advanced). Keep the default pool
+    # bounded and fail quickly: increasing it can hide leaked connections and
+    # makes SQLite lock contention worse. These fields are env-only below.
+    database_pool_size: int = Field(default=5, ge=1, env="DATABASE_POOL_SIZE")
+    database_max_overflow: int = Field(default=0, ge=0, env="DATABASE_MAX_OVERFLOW")
```
jeroenubbink: Flipping these two defaults isn't behaviour-neutral — it changes the runtime envelope of every existing node on upgrade with no config edit on their side. Until now "unset" meant the SQLAlchemy baseline (`max_overflow=10`, `pool_timeout=30`). `max_overflow 10→0` hard-caps the pool at `pool_size` (5) with zero burst headroom → `TimeoutError`/503 at the 6th concurrent checkout; `pool_timeout 30→5s` makes a transient burst that used to queue-and-drain fail fast. To an operator that reads as "the node started throwing pool-timeout errors after upgrading" though their workload never changed. The fail-fast posture is reasonable (and the new hold-time diagnostic makes leaks visible) — but how aggressively to fail is a per-operator call. Suggest keeping the shipped defaults at the baseline and documenting `0 / 5s` in `.env.example` as recommended fail-fast tuning; the hold-warn + `pool_recycle` can stay on by default (observability, not a capacity cut).

routstr/wallet.py
```diff
@@ -935,108 +942,164 @@ async def periodic_payout() -> None:
...
+    async with db.create_session() as session:
+        await session.exec(  # type: ignore[call-overload]
+            update(db.CashuTransaction)
+            .where(col(db.CashuTransaction.id) == refund_id)
+            .values(**values)
+        )
+        await session.commit()
+
+
 async def _refund_sweep_once(cutoff: int) -> None:
```
jeroenubbink, line 1062: Design thought, not blocking: this refund-sweep lease (`sweep_started_at` timestamp + time-cutoff predicate + conditional UPDATE) hand-rolls a distributed lease on the same table, right after the codebase added a durable idempotency-record table (`reservation_releases`, migration `7f2843d3f4e4`) for the analogous "did this side-effect already happen" question. Two parallel idempotency mechanisms now coexist, and the time-lease is what forces the ambiguous "already spent" classification. A shared "durable operation attempt" primitive would unify both and disambiguate the two financially-distinct outcomes. Fine to defer, worth noting for the direction.

routstr/lightning.py
```diff
@@ -222,79 +222,129 @@ async def recover_invoice(
...
     invoice: LightningInvoice, session: AsyncSession
 ) -> None:
+    minted = False
+    invoice_id = invoice.id
+    invoice_purpose = invoice.purpose
     try:
-        wallet = await get_wallet(settings.primary_mint, "sat")
+        # A preceding invoice lookup starts a transaction. End it before the
+        # potentially slow mint request so it cannot pin a pool connection.
+        await session.commit()
```
jeroenubbink, line 235: `check_invoice_payment`, `periodic_payout`, `periodic_routstr_fee_payout`, and `_refund_sweep_once` each now hand-roll the same shape: release the connection → do mint/LNURL I/O with no session held → re-open a session for a conditional-UPDATE checkpoint. The invariant "never hold a pool connection across external I/O" (this PR's whole point) is enforced only by each author remembering it — the only guardrail is the after-the-fact hold-time *warning*. A shared context manager / helper for the two-phase flow would make a future streaming route unable to silently reintroduce the pin. Defer-able.

routstr/wallet.py
```diff
@@ -935,108 +942,164 @@ async def periodic_payout() -> None:
...
+                            wallet, mint_url, unit, not_reserved=True
+                        )
+                        proofs = await slow_filter_spend_proofs(proofs, wallet)
+                        await asyncio.sleep(5)
+                        # Fetch the liability AFTER the proofs snapshot and the
+                        # settle delay: a concurrent top-up then only inflates
+                        # the liability, shrinking the payout — never sending
+                        # customer-backed funds as profit.
+                        async with db.create_session() as session:
+                            balances = await db.balances_by_mint_and_unit(
```
jeroenubbink: A few non-blocking cleanups I noticed (grouping them here):
- **This line:** `periodic_payout` calls `balances_by_mint_and_unit` with singleton `[mint]`/`[unit]` per loop iteration — an `IN(…)+GROUP BY` per pair where the helper was built to answer all pairs in one query. A single-key scalar read fits better here (the settle-delay ordering blocks a full up-front hoist).
- **`lightning.py` `topup_api_key_from_invoice`** now has no callers (prod or tests); `create_api_key_from_invoice` is test-only — both duplicate the `get_wallet+mint` sequence inlined into `check_invoice_payment`. Candidates for deletion.
- **`wallet.py` (lines ~841 and ~942)** the `mint_urls` assembly is duplicated verbatim — one `_mints_to_inspect()` helper.
- **`wallet.py:1039`** the inline claim UPDATE duplicates `_set_refund_sweep_state`'s session/update/commit ceremony (add an optional predicate + return rowcount).
- **`wallet.py`** the `//1000` / `*1000` sat↔msat conversion is inlined 4+ times — a tiny `_to_sats(amount, unit)` helper would de-risk the over/undercharge class.
- **`db.py` (63-91)** the pool hold-time listeners bind to the module-global `engine` (so a test-patched or future second engine gets no instrumentation) and need a load-bearing `# noqa: E402` deferred import; attaching them inside the engine factory would be cleaner and drops the double `settings` import.

routstr/wallet.py
```diff
@@ -1017,26 +1110,78 @@ async def _refund_sweep_once(cutoff: int) -> None:
...
-                        extra={
-                            "id": refund.id,
-                            "error": str(e),
-                        },
+                        "Refund claim ownership changed before spent-token checkpoint",
+                        extra={"id": refund.id},
                     )
-        await session.commit()
+            else:
+                # Cancellation or a transient pre-redemption failure is a
```
jeroenubbink: The `else` branch still releases the claim on failures that happen *after* the token was spent. `redeemed = True` is only set once `recieve_token` has returned, so anything raised from inside it lands here instead:

- `swap_to_primary_mint` melts on the source mint before minting on the primary. Past the "melt succeeded, minting on primary" log (`wallet.py:600`) the refund token's proofs are gone. A `primary_wallet.mint()` failure that is not "outputs already signed", a recovery scan that raises, and the `balance_gained < minted_amount` refusal all produce messages with no "already spent" in them.
- `_redeem_same_mint` has the same window: if the mint processes `wallet.split()` but the response is lost, the token is spent and what surfaces is a transport error.

From there the lease is released, the next sweep re-claims with `reclaimed_stale_claim = False`, `recieve_token` now genuinely reports "already spent", and the row is booked `collected=True` - a node sweep permanently recorded as a client collection. That is the outcome the earlier lease finding was about; the `redeemed` flag closes the half *after* `recieve_token` returns but not the half inside it. Funds are safe either way, owner/liability accounting is not.

The comment is the load-bearing part: "a known-no-send outcome" is not knowable at this boundary - all this branch really knows is that the error text did not contain "already spent". `MintConnectionError` is not a safe signal either, since a connection failure at the melt step can equally mean the melt landed and the response did not.

Suggest inverting the default: retain the claim for anything raised out of `recieve_token` (what the `redeemed` path already does), and release only for failures that provably precede it - cancellation before the call, the deserialize and keyset validation. The cost of retaining is one `refund_sweep_claim_timeout_seconds` before a retry, weighed against a permanently wrong audit row.

routstr/lightning.py
```diff
@@ -222,80 +222,147 @@ async def recover_invoice(
...
+
+        # Do not redeem a paid top-up quote if its target has already been
+        # pruned. This validation owns a short-lived session and releases its
+        # connection before any mint I/O starts.
+        if invoice_purpose == "topup":
```
jeroenubbink, line 245: Two things about this block.

**Ordering.** It runs before `get_mint_quote` and its `if not mint_status.paid: return`, so every pending top-up invoice costs an extra `create_session()` plus a SELECT on every watcher poll - up to 100 invoices every 5 seconds - even when the quote was never paid. That is steady-state pool churn in the change that exists to remove it. Moving the block below the `mint_status.paid` check keeps the guarantee (never redeem a quote whose target is gone) and costs nothing for unpaid invoices.

**The early return is a dead end.** When the target key is gone the invoice keeps `status="pending"` indefinitely: the watcher re-logs this same error every 5 seconds until `expires_at`, the paid quote is never redeemed, and `get_invoice_status` (`lightning.py:166`) eventually flips a genuinely paid invoice to `"expired"` - telling someone who paid that their invoice expired. `recover_invoice`, the endpoint whose whole purpose is "I paid, give me my key back", returns `pending` with no signal at all.

It is broader than the pruned-key case: `check_invoice_payment` swallows every `Exception` into `logger.error` and returns `None`, so no caller can distinguish "not paid yet" from "paid, but we could not credit you". Note the asymmetry with the rest of this change - every other money-stranding path here gets a `critical`, while this one, where a user's sats sit unredeemed at the mint, gets a repeating `error`.

jeroenubbink (commented):
Re-reviewed at head `39f8015`. The headline issue from the last round is properly fixed, and every other item is addressed - thanks for taking both config asks as suggested.

Verified fixed:

- **Borrowed-session rollback.** Finalization now owns its own `create_session()`, the caller's session is never rolled back, and the lost-CAS branch re-reads the winning row in the owned session instead of refreshing the caller's. The new sibling-invoice assertions (`inspect(obj).expired is False`) in `test_post_mint_db_failure_keeps_invoice_pending_for_reconciliation` and `test_db_guard_credits_once_when_both_mints_succeed` drive exactly the failure mode I described - good coverage.
- Refund-sweep lease: claim-ownership predicates plus a rowcount check on every checkpoint (one window survives - first comment below).
- Fee-payout completion now routes to the specific "sent but checkpoint not completed" critical.
- `periodic_payout`'s DB-specific cycle log restored by isolating the liability read.
- Top-up target validated before minting; `sweep_started_at` out of the admin transactions payload.
- Pool defaults back to the SQLAlchemy baseline with the fail-fast values documented in `.env.example`, and pre-ping auto-enabling only for networked backends - upgrading a running node is behaviour-neutral again.
- All six cleanup items from the roll-up.

Independently verified on the branch: single alembic head `aa50fde387a2` chaining onto main, `ruff check` and `mypy` clean, 789 unit and 300 integration tests green.

Four comments below. The first is the one I would want closed before merge: it is the same accounting invariant as the last round's lease finding, reaching the same wrong outcome through a different path. The other three are small.

Nit, not worth its own comment: `tests/unit/test_admin_transactions.py` has no module docstring, unlike the other test files touched here.

routstr/lightning.py
```diff
@@ -222,80 +222,147 @@ async def recover_invoice(
...
+                    # read transaction; never refresh the caller's session.
+                    invoice.api_key_hash = committed_invoice.api_key_hash
+                    invoice.status = committed_invoice.status
+                    invoice.paid_at = committed_invoice.paid_at
+                return
+            await finalization_session.commit()
+
+        # Only publish finalized values to the caller-owned object after the
+        # owned transaction has committed successfully.
+        invoice.api_key_hash = finalized_api_key_hash
```
jeroenubbink: Small structural point, and it happens to fix the second half of the previous comment too: consider returning the outcome from `check_invoice_payment` rather than writing it onto the caller's `invoice`.

Assigning a mapped column on an object owned by someone else's session puts it in `session.dirty`, so the caller's next `commit()` emits `UPDATE lightning_invoices SET status=?, api_key_hash=?, paid_at=? WHERE id=?` with no predicate - quietly sidestepping the conditional update this function is careful to use a few lines above. In practice `get_session` never commits at teardown, so request handlers never flush it; `periodic_invoice_watcher`'s shared session does, on the next iteration's `await session.commit()`. The values match what was just written, so nothing breaks today - but the "exactly one conditional write" property holds by accident rather than by construction.

Returning the state (or raising a typed error) also gives the endpoints something to answer with, which is what the pending/expired dead end above needs.

routstr/wallet.py
```diff
@@ -1082,53 +1227,69 @@ async def periodic_routstr_fee_payout() -> None:
...
 
+            try:
+                amount_received = await raw_send_to_lnurl(
+                    wallet,
+                    proofs,
+                    ROUTSTR_LN_ADDRESS,
+                    "sat",
+                    amount=accumulated_sats,
+                )
+            except Exception:
```
jeroenubbink: This one stayed `except Exception` while the checkpoint block below was widened to `except BaseException` in this round. A `CancelledError` during `raw_send_to_lnurl` - shutdown while a payout is in flight, which is exactly when it matters - escapes both this handler and the outer `except Exception`, so the task dies with `payout_in_progress_msats` set and without the "outcome is unknown" critical; the operator finds out at the next process start instead. Same one-word change as the block below.

PR #636, opened 2026-07-26
ppq-auto-topup

9qeklajc:


routstr/wallet.py
```diff
@@ -282,6 +283,203 @@ async def send_token(amount: int, unit: str, mint_url: str | None = None) -> str
...
+            except Exception as e:
+                logger.debug(
+                    "Cashu mint cannot fund BOLT11 invoice",
+                    extra={"mint_url": mint_url, "unit": unit, "error": str(e)},
+                )
+
+    if not candidates:
+        raise ValueError("No configured Cashu mint has enough balance to pay invoice")
+
+    _, wallet, proofs, quote, mint_url, unit = max(
```
jeroenubbink: **Blocker.** This picks the mint with the highest *raw* proof balance, and `execute_bolt11_payment` then melts `quote.amount + fee_reserve + input fees` with no liability read anywhere. `_payout_mint_and_unit` sends only `proofs_balance - total_user_liability` ("Send only conservatively proven owner funds").

A melt to PPQ is irreversible, unlike the Routstr top-up whose ecash the provider card can refund. Same money leaving the node — why does it not walk the same path? Either subtract `total_user_liability` here or say why this one is exempt.

routstr/upstream/auto_topup.py
```diff
@@ -156,47 +261,635 @@ async def _check_and_topup(row: UpstreamProviderRow) -> None:
...
-                "new_balance_approx": balance + amount,
+                "invoice_id": topup.invoice_id,
+                "admin_action": f"POST /admin/api/upstream-providers/{row.id}/ppq-auto-topup/release",
             },
+            exc_info=True,
+        )
+        raise
+
+    settled = False
+    for attempt in range(PPQ_SETTLEMENT_ATTEMPTS):
```
jeroenubbink: **Blocker.** `check_topup_status` calls `raise_for_status`, so a non-2xx raises rather than returning `False`. This loop is unwrapped, so one PPQ 502 in the ~8s window — *after* the melt already spent the proofs — propagates out: `_mark_ppq_reconcile` never runs, the CRITICAL at line 869 never fires, and the operator sees only "Auto top-up failed for provider", indistinguishable from a pre-payment failure. The claim also stays `in_flight`, so no admin release until the lease expires.

Wrap the call — a raise should count as "not settled this attempt". Once melt returned paid, the reconcile flag and the CRITICAL have to be guaranteed.

routstr/upstream/auto_topup.py
```diff
@@ -156,47 +261,635 @@ async def _check_and_topup(row: UpstreamProviderRow) -> None:
...
+            result = await session.exec(  # type: ignore[call-overload]
+                update(CashuTransaction)
+                .where(
+                    col(CashuTransaction.id) == state_id,
+                    or_(
+                        CashuTransaction.collected == True,  # noqa: E712
+                        CashuTransaction.swept == True,  # noqa: E712
+                    ),
+                )
+                .values(
```
jeroenubbink, line 661: **Blocker (1/2).** This overwrites `token`, `amount`, `unit`, `mint_url` and `created_at` on the one row that exists per provider (`ppq-auto-topup-{id}`), so a completed top-up survives exactly until the next attempt.

Combined with `get_ppq_auto_topup_state` returning a bare `{"active": False}` once `collected` (line 376) and the unconditional exclusion in the admin transaction list, there is no way for an operator to answer "how many sats did this node melt to PPQ, to which mint, when" — for the only irreversible outward payment it makes. `_check_and_topup_routstr` does the opposite: a separate row per top-up (line 253), `source="auto_topup"`, not filtered.

Suggestion: keep `ppq-auto-topup-{id}` a pure lock and hide only the lock; write a separate terminal row for the spend at line 875. The refund sweep will not touch it — it only selects `collected == False and swept == False`.

routstr/core/admin.py
```diff
@@ -1633,7 +1790,10 @@ async def get_transactions_api(
     async with create_session() as session:
         from sqlmodel import col, func
 
-        base = select(CashuTransaction)
+        base = select(CashuTransaction).where(
```
jeroenubbink, line 1820: **Blocker (2/2).** This is applied to `base` before the `type`/`source`/`status`/`search` filters, so a PPQ row can never be listed — even `?source=ppq_auto_topup` returns nothing. Right for the lock row, wrong once that row is also the only record of the spend. See the `auto_topup.py` comment.

routstr/upstream/auto_topup.py
```diff
@@ -156,47 +261,635 @@ async def _check_and_topup(row: UpstreamProviderRow) -> None:
...
+        extra={
+            "provider_id": row.id,
+            "balance_usd": balance,
+            "threshold_usd": threshold_usd,
+            "topup_usd": amount_usd,
+        },
+    )
+
+    try:
+        topup = await provider.initiate_topup(amount_usd)
```
jeroenubbink, line 950: **Blocker.** `initiate_topup` creates a real invoice on PPQ's side, and only then do the local preconditions run — `prepare_bolt11_payment` (line 782, raises "No configured Cashu mint has enough balance") and `sats_usd_price()` (line 784, raises while the price is uninitialised). Both are pure local reads.

Neither is exotic: `update_prices_periodically` returns before its first `_update_prices()` when `enable_pricing_refresh` is false, so the price stays `None` for the process lifetime; and a wallet that cannot cover the amount is the normal state of a node that needs a top-up. In either case this loops at 60s creating one fresh unpayable invoice a minute, indefinitely (~1400/day), plus a melt quote at every configured mint each pass. No funds are lost — the claim is released each time — but it is unbounded churn against a third party, and a rate limit on their side would break top-ups outright.

Read the price and pick the funding mint before calling PPQ.

routstr/upstream/auto_topup.py
```diff
@@ -41,19 +56,19 @@ async def periodic_auto_topup() -> None:
...
                 extra={"error": str(e), "error_type": type(e).__name__},
             )
-
         await asyncio.sleep(AUTO_TOPUP_INTERVAL_SECONDS)
 
 
 async def _run_auto_topup_cycle() -> None:
-    """Single cycle: check all eligible providers and top up if needed."""
+    """Check all eligible providers once, isolating failures by provider."""
+    await _reconcile_all_ppq_claims()
```
jeroenubbink: **Blocker, pending an answer.** If this settles a claim (`collected=True`), the provider loop below then runs `_check_and_topup_ppq`, whose own `_reconcile_ppq_state` now returns `False`, so `get_balance()` is called microseconds later. `check_topup_status` returning `"Settled"` is a statement about the invoice, not about the balance endpoint having been credited.

Does PPQ credit `/credits/balance` synchronously with `/topup/status` returning `Settled`? If it can lag, this ordering buys a second top-up in the same cycle. We are asking them as well.

routstr/upstream/auto_topup.py
```diff
@@ -63,69 +78,164 @@ async def _run_auto_topup_cycle() -> None:
...
     if balance is None:
         logger.warning(
-            "Could not fetch balance for auto top-up",
+            "Could not fetch Routstr balance for auto top-up",
             extra={"provider_id": row.id, "base_url": row.base_url},
         )
         return
-
-    if balance >= threshold * 1000:
+    if balance >= threshold:
```
jeroenubbink: This changes the *existing* Routstr auto-topup inside a PPQ PR, unmentioned. The new comparison is right — `get_balance()` returns sats and the UI says "(Sats)" — so the old one was a 1000x bug. But a stored threshold of `5` used to fire below 5000 sats and now fires below 5: auto top-up effectively stops until the provider runs dry. Do stored thresholds need rescaling, and can this go in its own PR?

routstr/upstream/auto_topup.py
```diff
@@ -63,69 +78,164 @@ async def _run_auto_topup_cycle() -> None:
...
+    if _invalid_number(settings.get("topup_amount_limit"), integer=True):
+        return "Auto top-up amount must be a positive whole number"
+
+    if provider_type == "routstr":
+        mint_url = settings.get("topup_mint_url")
+        if not isinstance(mint_url, str) or not mint_url.strip():
+            return "Auto top-up requires a mint URL"
+    elif provider_type == "ppqai":
+        amount = int(settings["topup_amount_limit"])
+        if not PPQ_MIN_TOPUP_USD <= amount <= PPQ_MAX_TOPUP_USD:
```
jeroenubbink: `topup_threshold`/`topup_amount_limit` live in one blob but mean sats for `routstr` and USD for `ppqai`. The validator cannot catch reinterpretation — `500` is legal under both readings — so a type change via the admin API with settings carried over silently turns 500 sats into $500. Type-scoped keys or an explicit unit field would close it.

routstr/wallet.py
```diff
@@ -282,6 +283,203 @@ async def send_token(amount: int, unit: str, mint_url: str | None = None) -> str
...
+        result = await asyncio.wait_for(
+            plan.wallet.melt(
+                proofs=selected,
+                invoice=plan.invoice,
+                fee_reserve_sat=plan.quote.fee_reserve,
+                quote_id=plan.quote.quote,
+            ),
+            timeout=60,
+        )
+    except Exception as e:
```
jeroenubbink: `except Exception` misses `CancelledError` (`BaseException` since 3.8). On shutdown the auto top-up task is cancelled, so the `set_reserved_for_melt(..., quote_id=...)` recovery below is skipped — the exact case `test_wallet_melt_restart.py` proves is needed for `get_melt_quote()` to find the proofs. `except BaseException`, re-reserve, re-raise.

tests/integration/test_ppq_auto_topup_claim.py
```diff
@@ -0,0 +1,516 @@
...
+        is False
+    )
+
+
+async def test_ppq_claim_rows_are_excluded_from_the_admin_transaction_list(
+    patched_db_engine: Any,
+) -> None:
+    from routstr.core.admin import get_transactions_api
+
+    await _claim_ppq_topup(_row())
```
jeroenubbink, line 237: This test passes without exercising the filter. There is no `_seed_provider()` here, so `_claim_ppq_topup` hits `_ppq_provider_is_claimable` → `session.get(UpstreamProviderRow, 1)` → `None` and returns before inserting anything; the final assert holds because the row was never created. `test_claim_is_refused_when_the_provider_row_is_gone` asserts exactly that, and every other claim test in this file seeds first.

Adding `await _seed_provider()` here gives the admin-list exclusion its only coverage.

tests/unit/test_auto_topup.py
```diff
@@ -4,14 +4,32 @@
...
+from routstr.upstream.auto_topup import (
+    _check_and_topup,
+    _parse_ppq_request_id,
+    validate_auto_topup_settings,
+)
+from routstr.upstream.ppqai import PPQAIUpstreamProvider
+
+
+def test_ppq_claim_parser_rejects_invalid_expiry() -> None:
+    assert _parse_ppq_request_id("ppq:operation:not-a-timestamp:invoice") is None
```
jeroenubbink: Also vacuous: `"ppq:operation:not-a-timestamp:invoice"` splits into 4 parts, so `_parse_ppq_request_id` returns at the `len(parts) != 6` check and never reaches `int(parts[2])`. This tests arity, not expiry parsing. A 6-part id with a valid phase reaches it, e.g. `"ppq:operation:not-a-timestamp:claimed:invoice-1:none"`.

tests/unit/test_auto_topup.py
```diff
@@ -141,3 +159,346 @@ async def test_auto_topup_does_not_send_untracked_token() -> None:
...
+            "routstr.upstream.auto_topup.execute_bolt11_payment",
+            AsyncMock(return_value=(101, "https://mint-rich.test", "sat")),
+        ) as execute,
+        patch(
+            "routstr.upstream.auto_topup._record_ppq_invoice", AsyncMock()
+        ) as record,
+        patch(
+            "routstr.upstream.auto_topup._set_ppq_state_terminal", AsyncMock()
+        ) as terminal,
+        patch("routstr.upstream.auto_topup.sats_usd_price", return_value=0.001),
```
jeroenubbink, line 254: Nothing tests the USD cap. All three cap-adjacent tests use `amount=10` with `sats_usd_price -> 0.001` and `maximum_spend_sats=102`, so the ceiling is 11000 sats and the check never fires — ~100x of headroom. It is the only thing between a hostile or buggy PPQ invoice and an arbitrarily large melt, so it is worth a boundary case.

`Bolt11PaymentNotAttempted` also appears in no auto-topup test, leaving the release-and-retry branch uncovered.

tests/integration/test_ppq_auto_topup_claim.py
```diff
@@ -0,0 +1,516 @@
...
+
+import time
+from typing import Any
+from unittest.mock import AsyncMock, MagicMock, patch
+
+import pytest
+from sqlmodel import select
+
+from routstr.core.db import CashuTransaction, create_session
+from routstr.upstream.auto_topup import (
```
jeroenubbink, line 16: Nine private helpers are driven directly here, against the repo convention of testing through the public API. Tests bound to private helpers pin the implementation, so a behaviour-preserving refactor still breaks the suite. (Ian Cooper, "TDD, Where Did It All Go Wrong"; https://martinfowler.com/bliki/UnitTest.html)

routstr/upstream/auto_topup.py
```diff
@@ -156,47 +261,635 @@ async def _check_and_topup(row: UpstreamProviderRow) -> None:
...
+        return True
+
+    released = await _set_ppq_state_terminal(
+        row, claim.operation_id, collected=False, swept=True
+    )
+    return not released
+
+
+async def _ppq_provider_is_claimable(
+    session: object, provider_id: int | None
```
jeroenubbink: `session: object` plus the `# type: ignore[attr-defined]` below turns this into an unchecked call under `disallow_untyped_defs`. `AsyncSession` types it properly.

routstr/wallet.py
```diff
@@ -282,6 +283,203 @@ async def send_token(amount: int, unit: str, mint_url: str | None = None) -> str
...
+        # The mint is telling us it did not pay, so the proofs are ours again.
+        await plan.wallet.set_reserved_for_send(selected, reserved=False)
+        raise Bolt11PaymentNotAttempted("Cashu mint reported the melt as unpaid")
+
+    raise Bolt11PaymentAmbiguous(
+        f"Cashu melt did not reach a final state: {state or 'unknown'}"
+    )
+
+
+async def pay_bolt11_invoice(invoice: str) -> tuple[int, str, str]:
```
jeroenubbink: No production caller — only `tests/unit/test_wallet.py`. Same question for `Bolt11PaymentAmbiguous`: it is raised here but never caught by name, so `auto_topup.py` catches it via bare `except Exception` and a real mint ambiguity takes the same branch as a `TypeError`. Fail-safe, but the distinction the class exists to draw is unused.

Was more meant to land in this PR? If not, worth removing these along with their tests.

routstr/wallet.py
```diff
@@ -999,6 +1197,10 @@ async def _refund_sweep_once(cutoff: int) -> None:
             db.CashuTransaction.type == "out",
             db.CashuTransaction.collected == False,  # noqa: E712
             db.CashuTransaction.swept == False,  # noqa: E712
+            # NULL != 'x' is NULL in SQL, so legacy rows with no source would
+            # drop out of the sweep without the explicit IS NULL arm.
+            (db.CashuTransaction.source != "ppq_auto_topup")
```
jeroenubbink: The `IS NULL` arm is unreachable and the comment states the opposite of the schema: `cashu_transactions.source` is `nullable=False, server_default="x-cashu"` (migration `c3d4e5f6a7b8`). Harmless, but it should not assert what the schema forbids.

routstr/upstream/auto_topup.py
```diff
@@ -156,47 +261,635 @@ async def _check_and_topup(row: UpstreamProviderRow) -> None:
...
+                "it is no longer owned by this attempt",
+                extra={"provider_id": row.id, "invoice_id": invoice_id},
+            )
+
+
+async def _check_and_topup_ppq(row: UpstreamProviderRow, settings: dict) -> None:
+    threshold_usd = float(settings["topup_threshold"])
+    amount_usd = int(settings["topup_amount_limit"])
+    provider = PPQAIUpstreamProvider.from_db_row(row)
+    if provider is None or await _reconcile_ppq_state(row, provider):
```
jeroenubbink, line 890: For a provider with an active claim this is the second `_reconcile_ppq_state` of the cycle — `_reconcile_all_ppq_claims` already ran one at the top, each with its own session and its own PPQ call. Cheap to avoid, and it widens the window in the duplicate-top-up question above.

routstr/wallet.py
```diff
@@ -282,6 +283,203 @@ async def send_token(amount: int, unit: str, mint_url: str | None = None) -> str
...
+                    (balance_msats, wallet, proofs, quote, mint_url, unit)
+                )
+            except Exception as e:
+                logger.debug(
+                    "Cashu mint cannot fund BOLT11 invoice",
+                    extra={"mint_url": mint_url, "unit": unit, "error": str(e)},
+                )
+
+    if not candidates:
+        raise ValueError("No configured Cashu mint has enough balance to pay invoice")
```
jeroenubbink: Every per-mint failure is swallowed at `debug` above, so this message is what the operator gets even when the real cause was a mint outage or a rejected melt quote — and it is then wrapped in the generic "Auto top-up failed for provider". Worth collecting the per-mint reasons, logging them at `warning` when no candidate survives, and distinguishing "every mint refused" from "no mint had enough funds".

jeroenubbink (changes requested):
Five blockers, all inline. The claim design itself holds up — these are about what happens around it.

1. The melt applies no user-liability floor, unlike `_payout_mint_and_unit`.
2. A PPQ API error right after a successful melt loses the CRITICAL entirely.
3. A completed top-up leaves no operator-visible record anywhere.
4. Every local precondition runs after the PPQ invoice already exists.
5. Same-cycle duplicate top-up — needs an answer on PPQ's balance/status consistency.

Two tests also pass without exercising what they name.

Cross-PR: the ambiguous-melt handling here is the shape PR 597 needs. That one classifies melt outcomes by exception type, so a `wait_for` cancellation of a melt that did settle returns 503 "redeem at another mint"; this PR treats ambiguity as non-retryable and re-reserves by melt quote id, which `test_wallet_melt_restart.py` shows is the reservation that survives a restart.

PR #648, opened 2026-08-04
chore: remove unused dashboard code

ashen0x:
Removes 6,582 lines of ui/ that nothing references, plus the seven npm packages only they imported. Tracker: dashboard-dead-code.

## What this is

- The old shell: app-sidebar, site-header, nav-*, ui/sidebar. Replaced when every page moved to app-page-shell, but never deleted. AppSidebar had seven importers before that commit, zero after.
- The abandoned half of a provider refactor. The live copies sit one directory up with kebab-case names.
- Services, schemas and hooks for /api/users and /api/api-keys, endpoints the backend never got.
- The shadcn demo data file and the default Next starter SVGs.
- One functional edit: ui/app/providers.tsx (+4) unwraps AuthProvider now that it is gone.

## How dead was decided

Two checks had to agree before a file made the list:

1. Unreachable from the import graph of all ten entry pages.
2. None of its exported names appear anywhere in the surviving tree.

The second check matters because of lookalikes. The deleted providers/ProviderBalance.tsx shares its name with provider-balance.tsx one directory up, which is alive and stays.

Both checks were re-run after rebasing onto current main, since the autotopup work added new dashboard code in between. It imports nothing from this list.

## What stayed, and why

- The api-endpoint-tester chain (~1,900 lines) is also unreachable, but its tab was switched off in 3681fc8 with "no advanced testing for now". That reads like something parked, not something forgotten, so i didn't delete it over that commit. If it's not coming back I can remove it in a follow-up.
- The unused components/ui/ primitives, and the ~20 Radix packages only they reference. Whether the vendored shadcn catalogue stays as stock or counts as dead code is a convention question, so i left it for now. I'd lean towards trimming it.
- lib/services/__tests__/cost-validation.test.ts looks orphaned but tests live billing code. It's only unreachable because no test runner is wired up, which is a missing-setup problem, not dead code.

## Tested

Build, lint and format-check pass. Clicked through every dashboard page against a running node before and after, no differences.


9qeklajc:
LGTM

PR #651, opened 2026-08-06
Make withdrawn tokens recoverable

ashen0x:
Live preview: https://29380tojcqwoksuar2gg4xpowomw1up2ftvhvuhatveib1bttkpr-618be6ad.nsite.lol/ (built pre-split, so it also shows the pieces that moved out)

Rebased down to the recovery feature per the split. The clipboard hook and error helper commits sit underneath and drop off as those PRs merge.

## Why

Withdrawing hands you a cashu token, shown once. It is the money. The dialog promised a copy under Transactions, but that page never queried admin withdrawals, so it was nowhere. And nothing stopped the dialog closing with the token unsaved.

## What changed

- Withdrawals tab on Transactions (`source=admin`), reusing the existing table so row copy comes free.
- Rows show **Issued** rather than a forever-Pending: the node hands over a bearer token and never learns whether it was redeemed. The refund sweeper does eventually flip `collected` or `swept`, and those now win over Issued, so a reclaimed token is never displayed as outstanding.
- A successful withdrawal invalidates the tab's query; the 5 minute `staleTime` meant a fresh token was missing from the very tab the dialog names.
- Close guard: Escape, backdrop and the X all refused, so the X is gone; closing needs the "I have saved this token" checkbox. Download .txt added as a second exit.
- Tab strip scrolls on narrow screens; `justify-start` is load bearing, centered lists push the first tab into unreachable scroll space.
- Download .txt on withdrawal rows, shown only while the token is still redeemable. The token is never rendered in the table, so a blocked clipboard was otherwise the only way out.

## Known limitations

1. The silent-200 audit write got fixed on main by #654 while this was in review, so that earlier limitation is gone.
2. A failed fetch shows "No transactions found", all four tabs, pre-existing.
3. Filtering Pending returns withdrawals shown as Issued; they genuinely are `collected=false AND swept=false`.
4. No UI tests yet, the repo has no runner wired. Next on this branch: vitest plus tests for the invalidation, badge and close guard.

## How tested

| Withdrawal rows across all three states | Withdrawal dialog |
|---|---|
| Withdrawal rows: Issued keeps Copy and Download, Swept and Collected keep Copy only | Withdrawal dialog naming the Withdrawals tab, with no close X |

Real withdrawals against a local node on a test mint. The token recovered from the tab matched the dialog byte for byte, and Download .txt produced that same token from both the dialog and a history row. Rows in swept and collected states were checked to confirm they lose the Download action and stop reading Issued. Broke the wiring first to prove it: without the invalidation Refresh left this tab stale,without the page reset Clear Filters stranded it on page 2.

### DIFF
--- a/ui/app/transactions/page.tsx
+++ b/ui/app/transactions/page.tsx
@@ -66,6 +66,28 @@ import { toast } from 'sonner';
 
 const STORAGE_KEY = 'routstr-transaction-filters';
 
+const STATUS_BADGES: Record<
+  Transaction['status'],
+  { label: string; className: string }
+> = {
+  issued: {
+    label: 'Issued',
+    className: 'border-gray-500/20 bg-gray-500/10 text-gray-500',
+  },
+  collected: {
+    label: 'Collected',
+    className: 'border-green-500/20 bg-green-500/10 text-green-500',
+  },
+  swept: {
+    label: 'Swept',
+    className: 'border-orange-500/20 bg-orange-500/10 text-orange-500',
+  },
+  pending: {
+    label: 'Pending',
+    className: 'border-blue-500/20 bg-blue-500/10 text-blue-500',
+  },
+};
+
 function TransactionTable({
   transactions,
   copiedId,
@@ -197,10 +219,9 @@ function TransactionTable({
                         <Copy className='h-4 w-4' />
                       )}
                     </Button>
-                    {/* The sweeper marks a withdrawal collected or swept once
-                        it can no longer be redeemed, and the token is never
-                        rendered anywhere. */}
-                    {tx.source === 'admin' && !tx.swept && !tx.collected && (
+                    {/* The token is never rendered in the table, so the
+                        clipboard must not be the only way to get it out. */}
+                    {tx.status === 'issued' && (
                       <Button
                         variant='ghost'
                         size='icon'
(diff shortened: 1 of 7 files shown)


9qeklajc:
nice looks good... it would be great if u could split it:

 1. Clipboard hook and mechanical clipboard migrations.
 2. API-error helper and mechanical migrations.
 3. Login hardening.
 4. Withdrawal recovery, including backend sweep policy, persistence contract, truthful
    status/error UX, and tests.

ashen0x:
> nice looks good... it would be great if u could split it:
> 
> 1. Clipboard hook and mechanical clipboard migrations.
> 2. API-error helper and mechanical migrations.
> 3. Login hardening.
> 4. Withdrawal recovery, including backend sweep policy, persistence contract, truthful
>    status/error UX, and tests.

yupp, on it, will put up the clipboard hook, the error helper and the login bits as their own PRs, then rebase this one down to just the recovery, with the sweep/status semantics written out properly and tests on top of my manual runs.

ashen0x:
split's done, this one's rebased down to just the recovery now, rest is in the desc

9qeklajc:
> split's done, this one's rebased down to just the recovery now, rest is in the desc

@ashen0x nice thnx

ashen0x:
pushed an update, "collected/swept never flip for these" line in the desc was wrong, the sweeper does flip them: swept when it reclaims the token, collected when the mint says already-spent. so both now beat the Issued badge and a reclaimed token never shows as outstanding. desc fixed.

also added download .txt on the rows, gated the same way, since the token isn't rendered in the table and a blocked clipboard was otherwise the only way out.

9qeklajc:
rebase pls

ashen0x:
> rebase pls

done

9qeklajc:
rebase/merge to sync the new main

ashen0x:
> rebase/merge to sync the new main

synced to main

ui/app/transactions/page.tsx
```diff
@@ -508,6 +552,18 @@ export default function TransactionsPage() {
           Collected
         </Badge>
       );
+    // An outstanding withdrawal ends at "issued": the node hands over a
+    // bearer token and never learns whether it was redeemed, so it would
+    // otherwise read Pending forever.
+    if (tx.source === 'admin')
```
9qeklajc: I think instead of hard coding, we should make sure the backend return the state correctly, similar to x-cashu and api-key

9qeklajc (changes requested):

ui/app/transactions/page.tsx
ashen0x: yeah fair, but thing is x-cashu and api-key don't return a state either, they just send collected/swept and the ui works it out, difference is the node redeems those itself so it knows, a withdrawal is a bearer token it hands over and never hears about again, so both stay false till the sweeper hits it.

want me to just have the backend compute a status and send that? would fix the pending filter too

9qeklajc: yes plesae ... I also could think that at some point, provider could be able to topup the node's wallet to pay for ppq or another routstr node for example, so the direction `in` and `out` is also important here

ashen0x: done, backend computes the status now and the ui just renders it,also stopped pending filter returning withdrawals and added in/out check so an incoming topup won't read as issued

9qeklajc:
LGTM

9qeklajc (approved):

PR #656, opened 2026-08-10
fix: one clipboard hook that only claims success when the copy landed

ashen0x:
Fourteen places had their own copy-to-clipboard code; this collapses them into one hook, and the copy button stops saying "Copied!" when nothing was copied. Split out of #651.

## Why

On plain HTTP, which is what `compose.yml` gives you, the browser has no `navigator.clipboard`. Each copy spot dealt with that in its own way, and the fallbacks were broken in one of two ways. Some just threw, so the button silently did nothing. The rest used `execCommand('copy')` and trusted its return value, but that value only says a copy was attempted, not that it worked. Inside a dialog it usually didn't work, because the dialog's focus trap grabs the selection. So the button said "Copied!" while the clipboard kept its old contents. I hit this on withdrawal tokens, where a false "Copied!" can cost real money.

## What changed

- New `use-copy-to-clipboard.ts`: tries `navigator.clipboard` first. When that's unavailable it writes the text from a `copy` event listener and only reports success if that event actually fired, so it cannot claim a copy that didn't happen. It also owns the checkmark state (keyed, for components with several copy buttons) and shows a "copy manually" toast when both paths fail.
- Eleven files switch to it and delete their own versions.
- Small win along the way: the parent-key copy button in child-key-creator had no feedback and no accessible label, now it has both.

## How tested

Build, lint, format-check and tsc pass. The fallback was proven on a real plain-HTTP deployment: old code said "Copied!" with the clipboard unchanged, new code actually pastes, checked in a dialog and in a plain table. localhost cannot reproduce this, browsers treat it as a secure context so the fallback never runs there.

### DIFF
--- a/ui/hooks/use-copy-to-clipboard.ts
+++ b/ui/hooks/use-copy-to-clipboard.ts
@@ -1,0 +1,69 @@
+'use client';
+
+import { useCallback, useEffect, useRef, useState } from 'react';
+import { toast } from 'sonner';
+
+async function writeToClipboard(text: string): Promise<boolean> {
+  // navigator.clipboard is undefined on plain-HTTP deployments (non-secure context)
+  if (typeof navigator !== 'undefined' && navigator.clipboard) {
+    try {
+      await navigator.clipboard.writeText(text);
+      return true;
+    } catch {}
+  }
+  // execCommand reports that a copy ran, not that anything landed, so write the
+  // text from the copy event and treat that firing as the proof of delivery.
+  let delivered = false;
+  const onCopy = (event: ClipboardEvent) => {
+    event.clipboardData?.setData('text/plain', text);
+    event.preventDefault();
+    delivered = true;
+  };
+  try {
+    document.addEventListener('copy', onCopy);
+    return document.execCommand('copy') && delivered;
+  } catch {
+    return false;
+  } finally {
+    document.removeEventListener('copy', onCopy);
+  }
+}
+
+/**
+ * Clipboard copy with a success indicator that never lies: `copied` only
+ * turns true when the text actually reached the clipboard.
+ * `key` distinguishes multiple copy targets sharing one hook instance.
+ */
+export function useCopyToClipboard(resetMs = 2000) {
+  const [copiedKey, setCopiedKey] = useState<string | null>(null);
+  const timer = useRef<ReturnType<typeof setTimeout> | null>(null);
+  const latest = useRef(0);
+
+  const copy = useCallback(
+    async (text: string, key = 'default'): Promise<boolean> => {
+      // Copies can overlap; only the newest attempt owns the indicator, so a
+      // slow earlier failure cannot clear a later success.
+      const attempt = ++latest.current;
+      const ok = await writeToClipboard(text);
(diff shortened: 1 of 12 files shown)


ui/hooks/use-copy-to-clipboard.ts
```diff
@@ -0,0 +1,66 @@
...
+  const copy = useCallback(
+    async (text: string, key = 'default'): Promise<boolean> => {
+      const ok = await writeToClipboard(text);
+      if (timer.current) clearTimeout(timer.current);
+      if (ok) {
+        setCopiedKey(key);
+        timer.current = setTimeout(() => setCopiedKey(null), resetMs);
+      } else {
+        setCopiedKey(null);
+        toast.error('Copy failed. Select the text and copy it manually.');
```
9qeklajc: the message is missleading in terms that users wouldn't be able to copy the token as for now there is not modal that show the full token to select and copy .. would be nice to have something like onhover pop diablog where select and copy would work

9qeklajc (changes requested):

9qeklajc:
nice thnx.. some small changes are needed though

``` 
- ui/components/withdraw-modal.tsx:89
 - ui/components/providers/SimpleLightningTopup.tsx:48
 - ui/components/landing/key-info-display.tsx:34
 - ui/components/ecash-redeem.tsx:83
 - ui/components/landing/api-key-manager.tsx:86
```
these still have old copy style, could be also migrated

9qeklajc:
a side note, we do not really need to comment the code if it's clear

ui/hooks/use-copy-to-clipboard.ts
ashen0x: changed it, now it just says the browser blocked clipboard access instead of telling people to do something that isn't always possible.

for the token case specifically, the fix is really in #651, Its withdraw modal already shows the full token in a readonly textarea with copy and download next to it and i'm adding the same download to the history rows there so a blocked clipboard is never the only way out.

ashen0x:
four of those files (SimpleLightningTopup, key-info-display, ecash-redeem, api-key-manager) were already deleted from main in 91ac33e via #648, this branch was just cut before that landed, rebased so they're gone now. withdraw-modal is the only real one and it's converted in #651.

dropped the comment that only restated the code, kept the two explaining why the fallback looks the way it does.

also fixed one i found while in there: clicking two copy buttons quickly, the older one finishing could wipe the newer one's tick and fire a false "copy failed", three lines so the newest click wins.

on review order, these four are stacked so it's easiest as: #657 first (smallest, and both #658 and #651 contain its commit), then this one, then #658, then #651 last. #651 shows 18 files right now but 14 of those are just #656 and #657 riding along, it drops to 4 once they land.

ashen0x:
rebased this one too, it went into conflict with main the moment #657 landed, both touch the same import block in cli-tokens-settings

9qeklajc:
> withdraw-modal is the only real one and it's converted in https://github.com/Routstr/routstr-core/pull/651.

but why not here as this handle the migration ?

9qeklajc (approved):

PR #657, opened 2026-08-10
fix: surface backend error details instead of raw axios strings

ashen0x:
Error toasts and alerts were showing axios's generic "Request failed with status code 500" while the backend's `detail` field, which says what actually went wrong, got dropped on the floor. Split out of #651.

## What changed

- New `lib/api/errors.ts` with one helper, `getApiErrorMessage`. It prefers the backend's `detail` string; when detail is missing or not a string (FastAPI validation errors return arrays, for example) it falls back to the operation name plus the HTTP status; and when there is no response at all it says "Cannot reach the node. Is it running?".
- The providers page, CLI tokens, temporary balances and the detailed balance alert switch to it.

## How tested

Build, lint, format-check and tsc pass. The non-string detail shapes were checked specifically, so a validation error renders the fallback text instead of `[object Object]`.

### DIFF
--- a/ui/app/providers/page.tsx
+++ b/ui/app/providers/page.tsx
@@ -3,6 +3,7 @@
 import { Button } from '@/components/ui/button';
 import { Card, CardContent } from '@/components/ui/card';
 import { useQuery, useMutation, useQueryClient } from '@tanstack/react-query';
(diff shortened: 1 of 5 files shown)


9qeklajc:
LGTM

PR #658, opened 2026-08-10
fix: harden the admin login flow

ashen0x:
Login failures were a vanishing toast that always said "Invalid password", even when the real problem was that the node wasn't reachable. And a mistyped API URL stuck in config for every later attempt. Sits on the error-helper PR, so the tip commit is the review surface. Split out of #651.

## What changed

- Errors show inline under the form with `role="alert"` instead of a toast, and clear on the next attempt.
- Three cases get told apart: no response names the exact URL it tried and asks if the node is running, a 401 says the password is wrong, anything else goes through `getApiErrorMessage`.
- The manual base URL is still written before the attempt, since the login request has to target it, but it rolls back when the node turned out unreachable. A 401 keeps the URL on purpose: getting an HTTP response proves the URL was right and only the password was wrong.
- Inputs get labels, names and autocomplete attributes, so password managers offer to fill and save.

## How tested

Build, lint, format-check and tsc pass. Wrong-password, unreachable-node and success paths exercised against a local node, old code and new side by side:

| Before, node stopped | After, node stopped | After, wrong password |
|---|---|---|
| Old login blames the password while the node is completely down | New login names the unreachable node URL inline | Wrong password as a persistent inline error under a labeled input |

The first two columns are the same node fully stopped case. Old code blames the password anyway; new code says what's actually wrong and where it looked. The custom-URL rollback branch is not exercisable in this dev config (the URL field is config-gated), so that path is covered by review rather than a click.

### DIFF
--- a/ui/app/login/page.tsx
+++ b/ui/app/login/page.tsx
@@ -5,7 +5,13 @@ import type { ChangeEvent, FormEvent, ReactElement } from 'react';
 import { useRouter } from 'next/navigation';
 import { Button } from '@/components/ui/button';
 import { Input } from '@/components/ui/input';
+import { Label } from '@/components/ui/label';
 import { adminLogin } from '@/lib/api/services/auth';
+import {
+  getApiErrorMessage,
+  isNodeUnreachable,
+  isUnauthorized,
+} from '@/lib/api/errors';
 import { ConfigurationService } from '@/lib/api/services/configuration';
 import { toast } from 'sonner';
 import { AuthPageShell } from '@/components/auth-page-shell';
@@ -16,6 +22,7 @@ export default function AdminLoginPage(): ReactElement {
   const [password, setPassword] = useState<string>('');
   const [baseUrl, setBaseUrl] = useState<string>('');
   const [isLoading, setIsLoading] = useState<boolean>(false);
+  const [loginError, setLoginError] = useState<string>('');
 
   useEffect(() => {
     if (ConfigurationService.isTokenValid()) {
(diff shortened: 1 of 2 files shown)


ui/app/login/page.tsx
```diff
@@ -3,9 +3,12 @@
 import { useState, useEffect } from 'react';
 import type { ChangeEvent, FormEvent, ReactElement } from 'react';
 import { useRouter } from 'next/navigation';
+import { isAxiosError } from 'axios';
```
9qeklajc: please no direct denendency to axios into ui components.. imo. better and cleaner to do the logic similar to `getApiErrorMessage` function, but rather for axios

9qeklajc:
pls rebase

9qeklajc (changes requested):

ui/app/login/page.tsx
ashen0x: oh yeah , it was the only component importing axios , moved both checks into isNodeUnreachable and isUnauthorized in lib/api/errors.ts, and rebased.

ui/app/login/page.tsx
```diff
@@ -65,7 +76,18 @@ export default function AdminLoginPage(): ReactElement {
       router.push('/');
     } catch (error) {
       console.error('Login error:', error);
-      toast.error('Invalid password. Please try again.');
+      if (isNodeUnreachable(error)) {
+        setLoginError(
+          `Can't reach your node at ${ConfigurationService.getLocalBaseUrl()}. Is it running?`
+        );
+        if (allowCustomBaseUrl) {
+          ConfigurationService.setManualBaseUrl(previousBaseUrl);
```
9qeklajc, line 84: sorry missed this, I think the state shoul also be reset here ?

9qeklajc (changes requested):

ui/app/login/page.tsx
ashen0x: yeah stored one rolls back, you mean the input field right? left that on purpose, if like they didn't touch the field resetting changes nothing and if they did it throws away what they just typed. unreachable usually just means the node is off, so they'd start it and hit submit again, retyping the url every time. it resyncs from storage on reload anyway. can add it if you'd still rather they match though

9qeklajc:
LGTM

9qeklajc (approved):

PR #663, opened 2026-08-13
fix: handle Cashu tokens with short keyset IDs

ashen0x:
## What changed

Minibits' recent mint migration exposed an issue with Cashu TokenV4 tokens that use short keyset IDs.

These tokens are valid, but the short ID has to be matched with the mint's full keyset ID before the token can be redeemed. Core was passing the short ID into wallet operations that expected the full ID, which could cause valid tokens to fail with:

> A short keyset ID v2 was encountered, but got no keysets to map it to.

This change loads the source mint's current keysets, resolves short IDs using Cashu's existing helper, and keeps using the resolved proofs throughout the redemption flow.

It also makes keyset-loading failures happen before the token is spent, so the token can be returned safely and retried.

Nothing is hardcoded for Minibits, so this also works for future key rotations and other mints using short keyset IDs.

## Testing

Reproduced with a TokenV4 token generated by Minibits after the migration.

Added regression tests covering:

- short keyset IDs with multiple proofs
- same-mint redemption and cross-mint swaps
- legacy keyset IDs
- unknown and ambiguous short IDs
- missing or stale mint keysets
- failures happening before any wallet mutation

Verified with:

- `uv run pytest -q tests/unit/test_short_keyset_ids.py tests/unit/test_wallet.py tests/integration/test_swap_fee_retry.py` - 155 passed
- `make lint` - Ruff and mypy clean
- `make test-fast` - 1388 passed, with the same 3 pre-existing LiteLLM pricing-map failures as the base branch

## Related

The separate routstrd post-receive issue is handled in https://github.com/Routstr/routstrd/pull/68.

### DIFF
--- a/routstr/wallet.py
+++ b/routstr/wallet.py
@@ -299,6 +299,43 @@ async def get_balance(unit: str) -> int:
     return wallet.available_balance.amount
 
 
+async def _load_and_resolve_token_proofs(
+    wallet: Wallet, token_obj: Token, *, op_name: str
+) -> list[Proof]:
+    """Load mint keysets and return the proof list used for redemption.
+
+    ``TokenV4.proofs`` rebuilds its list on each access, so callers must reuse
+    this list after Cashu expands short keyset IDs in place.
+    """
+    # Cashu's load_mint() suppresses failures, which can leave cached keysets stale.
+    try:
+        await run_mint_operation(
+            lambda: wallet.load_mint_keysets(),
+            op_name=op_name,
+            mint_url=token_obj.mint,
+        )
+    except Exception as error:
+        if is_mint_connection_error(error) or is_mint_rate_limited(error):
+            raise
+        raise MintConnectionError("Cashu mint keysets are unavailable") from error
+    try:
+        await wallet.activate_keyset()
+    except Exception as error:
+        # Cashu raises plain Exception when no active keyset exists.
+        raise MintConnectionError("Cashu mint has no active keyset") from error
+
+    proofs = token_obj.proofs
+    try:
+        # Cashu's TokenV4 receive helper uses this resolver after load_mint().
+        await wallet._expand_short_keyset_ids(proofs)
+    except (KeyError, ValueError) as error:
+        raise ValueError(
+            "Cashu token references an unknown or ambiguous keyset"
+        ) from error
+
+    return proofs
+
+
 async def _redeem_same_mint(
     wallet: Wallet, token_obj: Token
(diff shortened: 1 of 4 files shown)


9qeklajc:
I think we should consider picking best parts from here and https://github.com/Routstr/routstr-core/pull/664

1. Swallow fixed only in redeem paths. get_wallet (line 1781) still calls upstream load_mint → rate-limit/outage still silently produces keyset-less cached wallets for every other flow
2. Fresh load_mint_keysets HTTP call on every redemption, even when nothing short and cache warm. 
3. Drops load_mint_info from redeem path entirely — behavior change nobody verified (split/melt may consult mint info for nut support).

ashen0x:
Thanks for taking a look at it, I went through both properly now. I like the wider `load_mint` change u added, that’s a nice catch, maybe we can keep that as a small follow-up since it touches all the `get_wallet` paths, not only this receive flow.

For the request part, same-mint was already calling `load_mint`, so this one actually saves the `/info` call there. Cross-mint does a refresh because after a rotation we need both the full id for the incoming proof and current active id for new outputs. checked the `mint_info` bit too and from what I can see `split` and `melt` don’t use it here and we don’t pass `amount_msat` to `melt_quote`. Let me know if I missed something there.

ashen0x:
Had one question on the cross mint path , i can see the proof ids get expanded correctly. Does `wallet.keyset_id` get updated too before the melt? Cashu uses that one when it creates fee change.
The test uses `fee_reserve=0`, so there aren’t any change outputs in that case. Did ur node test happen to use a non zero reserve? Just wanted to confirm that part too

ashen0x:
Also earlier found a couple of similar decode issues in sdk and chat while testing this. I’ll check those separately and put up fixes for them too.

9qeklajc:
> Had one question on the cross mint path , i can see the proof ids get expanded correctly. Does `wallet.keyset_id` get updated too before the melt? Cashu uses that one when it creates fee change. The test uses `fee_reserve=0`, so there aren’t any change outputs in that case. Did ur node test happen to use a non zero reserve? Just wanted to confirm that part too

yes `activate_keyset()` does that, which is used for melt .. 

for the `fee_reserve=0` a test would be great here ... we could do a follow up for the first point mentioned above after this is merged

ashen0x:
yupp added the non-zero fee reserve test now and yes we can keep the wider `load_mint` change for a followup.

9qeklajc:
LGTM

PR #669, opened 2026-08-17
feat(ehbp): surface tinfoil max_tokens via header for balance discounting

sh1ftred:
## Summary

EHBP / encrypted tinfoil requests have an **opaque, HPKE-sealed body**, so the proxy cannot read the client's completion cap to size the required balance. Previously the proxy always required the **full, context-window-sized `max_completion_cost`** regardless of what the client requested, which caused `402 insufficient_balance` errors for users who capped their tokens (e.g. "3188 sats required; 681 sats available").

This PR lets the client surface its `max_tokens` cap via a header on EHBP requests (mirroring the existing `X-Routstr-Model`), so the required balance scales down with the requested completion cap **without decrypting the encrypted body**.

## Changes

- **`routstr/proxy.py`**
  - On EHBP requests, read `X-Routstr-Max-Tokens` and pass it into `calculate_discounted_max_cost` as an explicit override (both the primary reservation and the failover re-reserve).
  - New `_validated_ehbp_max_tokens()` helper enforcing a **64,000-token minimum** (Tinfoil's enclave floor). A present header below 64k, or unparseable / `<=0`, returns a `400 invalid_request`. An absent header stays allowed (backwards compatible).
- **`routstr/payment/helpers.py`**
  - `calculate_discounted_max_cost()` accepts an optional `max_tokens` override; falls back to body `max_tokens`, then `max_completion_tokens` (also fixes the plain-request gap where only `max_tokens` was honored).
- **`routstr/upstream/ehbp.py`**
  - Added `X-Routstr-Max-Tokens` to the proxy-only headers so it is stripped before being forwarded to the enclave.

## Tests

- `test_coverage_proxy.py`: header parsing, 64k-floor validation, 400 responses, value flows into the discount, absent-header allowed.
- `test_payment_helpers.py`: override honored, `max_completion_tokens` fallback, override precedence over body.
- `test_tinfoil_integration.py`: updated proxy-only-header assertions.
- Full unit suite: **1050 passed, 1 skipped, 2 failed** — the 2 failures (`test_provider_slugs.py`) are pre-existing/environmental (they seed a `tinfoil` provider because `TINFOIL_API_KEY` is set in this env) and fail identically on `main`.

## Client-side note

The tinfoil SDK (`confidential-model-router`) should emit `X-Routstr-Max-Tokens: <max_tokens>` (>= 64000) alongside `X-Routstr-Model` on EHBP requests. That's a separate repo and not part of this PR.

### DIFF
--- a/routstr/payment/helpers.py
+++ b/routstr/payment/helpers.py
@@ -155,8 +155,15 @@ async def calculate_discounted_max_cost(
     max_cost_for_model: int,
     body: dict,
     model_obj: Any | None = None,
+    max_tokens: int | None = None,
 ) -> int:
-    """Calculate the discounted max cost for a request using model pricing when available."""
+    """Calculate the discounted max cost for a request using model pricing when available.
+
+    ``max_tokens`` is an explicit override used for opaque-body requests (e.g.
+    EHBP/encrypted tinfoil calls) where the proxy cannot read ``max_tokens``
+    from the request body. When ``None`` it falls back to the body's
+    ``max_tokens`` (then ``max_completion_tokens``).
+    """
     if settings.fixed_pricing:
         return max_cost_for_model
 
@@ -216,7 +223,12 @@ async def calculate_discounted_max_cost(
         if estimated_prompt_delta_sats > 0:
             adjusted = adjusted - math.floor(estimated_prompt_delta_sats * 1000)
 
-    max_tokens_raw = body.get("max_tokens", None)
+    if max_tokens is not None:
+        max_tokens_raw = max_tokens
+    else:
+        max_tokens_raw = body.get("max_tokens", None)
+        if max_tokens_raw is None:
+            max_tokens_raw = body.get("max_completion_tokens", None)
     if max_tokens_raw is not None:
         try:
             max_tokens_int = int(max_tokens_raw)
--- a/routstr/proxy.py
+++ b/routstr/proxy.py
@@ -273,9 +273,22 @@ async def _proxy(
     # raw encrypted body to the upstream's /private/ endpoint and stream the
     # encrypted response back untouched — the SDK's SecureClient decrypts it.
     is_ehbp = "ehbp-encapsulated-key" in headers
+    ehbp_max_tokens: int | None = None
     if is_ehbp:
+        # The body is HPKE-sealed/opaque, so the SDK also surfaces the client's
+        # completion cap ("max_tokens") via X-Routstr-Max-Tokens. This lets the
+        # proxy size the required balance down without decrypting the body.
+        # A present header below the enclave's floor is rejected outright.
         request_body_dict = {}
         model_id = headers.get("x-routstr-model", "")
+        try:
+            ehbp_max_tokens = _validated_ehbp_max_tokens(
+                headers.get("x-routstr-max-tokens", "")
+            )
+        except ValueError as exc:
+            return create_error_response(
+                "invalid_request", str(exc), 400, request=request
+            )
         if not model_id:
             return create_error_response(
                 "invalid_request",
@@ -364,7 +377,10 @@ async def _proxy(
         model=model_id, session=session, model_obj=model_obj
     )
     max_cost_for_model = await calculate_discounted_max_cost(
-        _max_cost_for_model, request_body_dict, model_obj=model_obj
+        _max_cost_for_model,
+        request_body_dict,
+        model_obj=model_obj,
+        max_tokens=ehbp_max_tokens,
     )
 
     check_token_balance(headers, request_body_dict, max_cost_for_model)
@@ -502,7 +518,10 @@ async def _proxy(
                 model=model_id, session=session, model_obj=model_obj
             )
             candidate_max = await calculate_discounted_max_cost(
-                candidate_max, request_body_dict, model_obj=model_obj
+                candidate_max,
+                request_body_dict,
+                model_obj=model_obj,
+                max_tokens=ehbp_max_tokens,
             )
             if candidate_max > max_cost_for_model:
                 await revert_pay_for_request(
@@ -829,6 +848,50 @@ async def get_bearer_token_key(
             },
         )
         raise
+
+
+def _parse_ehbp_max_tokens(raw: str) -> int | None:
+    """Parse the ``X-Routstr-Max-Tokens`` header into an integer (or ``None``).
+
+    ``None`` is returned for missing, non-integer, and non-positive values.
+    Range validation (e.g. a Tinfoil minimum completion cap) is the caller's
+    responsibility via :func:`_validated_ehbp_max_tokens`.
+    """
+    if not raw:
+        return None
+    try:
+        value = int(raw)
+    except (TypeError, ValueError):
+        return None
+    if value <= 0:
+        return None
+    return value
+
+
+# Tinfoil enclaves reject completion caps below this floor, so a client that
+# sends a lower ``X-Routstr-Max-Tokens`` is rejected rather than silently
+# priced off the (unreachable) smaller cap.
+_EHBP_MIN_MAX_TOKENS = 64_000
+
+
+def _validated_ehbp_max_tokens(
+    raw: str, min_tokens: int = _EHBP_MIN_MAX_TOKENS
+) -> int | None:
+    """Parse and range-check an ``X-Routstr-Max-Tokens`` header value.
+
+    Returns ``None`` when the header is absent (no constraint), or the parsed
+    integer when it is present and at least ``min_tokens``. Raises
+    ``ValueError`` when the header is present but invalid or below
+    ``min_tokens`` — callers should turn that into a 400.
+    """
+    if not raw:
+        return None
+    value = _parse_ehbp_max_tokens(raw)
+    if value is None or value < min_tokens:
+        raise ValueError(
+            f"X-Routstr-Max-Tokens must be an integer of at least {min_tokens}"
+        )
+    return value
 
 
 def extract_model_from_responses_request(request_body_dict: dict[str, Any]) -> str:
--- a/routstr/upstream/ehbp.py
+++ b/routstr/upstream/ehbp.py
@@ -52,6 +52,10 @@ logger = get_logger(__name__)
 _ENCLAVE_URL_HEADER = "X-Tinfoil-Enclave-Url"
 _REQUEST_USAGE_HEADER = "X-Tinfoil-Request-Usage-Metrics"
 _RESPONSE_USAGE_HEADER = "X-Tinfoil-Usage-Metrics"
+# Client supplies the completion cap via this header on EHBP (opaque-body)
+# requests so the proxy can size the required balance without decrypting the
+# body. It is stripped before the request is forwarded to the enclave.
+_MAX_TOKENS_HEADER = "X-Routstr-Max-Tokens"
 _TINFOIL_PROVIDER_TYPE = "tinfoil"
 _TINFOIL_ALLOWED_ENCLAVE_HOST_SUFFIX = ".tinfoil.sh"
 _TINFOIL_ALLOWED_ENCLAVE_HOSTS = frozenset({"tinfoil.sh"})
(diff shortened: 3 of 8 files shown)


routstr/payment/helpers.py
```diff
@@ -216,7 +223,12 @@ async def calculate_discounted_max_cost(
         if estimated_prompt_delta_sats > 0:
             adjusted = adjusted - math.floor(estimated_prompt_delta_sats * 1000)
 
-    max_tokens_raw = body.get("max_tokens", None)
+    if max_tokens is not None:
+        max_tokens_raw = max_tokens
+    else:
+        max_tokens_raw = body.get("max_tokens", None)
+        if max_tokens_raw is None:
+            max_tokens_raw = body.get("max_completion_tokens", None)
```
jeroenubbink, line 231: **Blocking.** This fallback reads a field nothing read before, and neither branch rejects non-positive or non-integer values. Running the branch with max cost 100000 msats and `min_request_msat` 1000:

```
{"max_completion_tokens": 0}          -> 1000
{"max_completion_tokens": -10000000}  -> 1000
{"max_completion_tokens": true}       -> 1000
{"max_completion_tokens": "80000"}    -> 80000   (a string; 400s as max_tokens)
```

The string case bypasses validation `max_tokens` gets via `parse_request_body_json`. There's no settlement clamp on the plain path, so `adjust_payment_for_tokens` drives the key balance negative. This one isn't EHBP-gated — it lands on every request.

routstr/payment/helpers.py
```diff
@@ -216,7 +223,12 @@ async def calculate_discounted_max_cost(
         if estimated_prompt_delta_sats > 0:
             adjusted = adjusted - math.floor(estimated_prompt_delta_sats * 1000)
 
-    max_tokens_raw = body.get("max_tokens", None)
+    if max_tokens is not None:
+        max_tokens_raw = max_tokens
+    else:
+        max_tokens_raw = body.get("max_tokens", None)
```
jeroenubbink, line 229: **Blocking.** When a body carries both fields we reserve on `max_tokens`, which is the one o-series upstreams reject. `strip_unsupported_param` matches that error, drops the field and retries the same upstream without recomputing `max_cost_for_model` — so `{"max_tokens": 500, "max_completion_tokens": 64000}` runs a 64000-token completion against a 500-token reservation.

Take the larger when both are present, and force a re-reserve after any body correction that removes a cap.

routstr/payment/helpers.py
```diff
@@ -216,7 +223,12 @@ async def calculate_discounted_max_cost(
         if estimated_prompt_delta_sats > 0:
             adjusted = adjusted - math.floor(estimated_prompt_delta_sats * 1000)
 
-    max_tokens_raw = body.get("max_tokens", None)
+    if max_tokens is not None:
+        max_tokens_raw = max_tokens
```
jeroenubbink, line 227: **Blocking (CI).** This narrowing makes `max_tokens_raw` an `int`, so the `body.get(...)` assignment below no longer type-checks:

```
helpers.py:229: error: Incompatible types in assignment (expression has type "Any | None", variable has type "int")
```

Plus two `None`-vs-`str` argument errors in the new test file, and a ruff I001 there. Both run in CI, so this can't go green as-is.

routstr/proxy.py
```diff
@@ -502,7 +518,10 @@ async def _proxy(
                 model=model_id, session=session, model_obj=model_obj
             )
             candidate_max = await calculate_discounted_max_cost(
-                candidate_max, request_body_dict, model_obj=model_obj
+                candidate_max,
+                request_body_dict,
+                model_obj=model_obj,
+                max_tokens=ehbp_max_tokens,
```
jeroenubbink, line 524: **Blocking (dead code).** `ehbp_max_tokens` is only ever set when `is_ehbp`, but the EHBP branch sets `request_body_dict = {}` and never repopulates it, and the guard above is `if i > 0 and request_body_dict:`. So this always receives `None` and the description's claim that the value flows into both the primary reservation and the failover re-reserve doesn't hold.

Worth filing separately, since it's pre-existing: an EHBP failover to a pricier second candidate still runs on the first candidate's reservation and then settles clamped to it, so the node under-bills on every EHBP failover.

routstr/proxy.py
```diff
@@ -831,6 +850,50 @@ async def get_bearer_token_key(
...
+        return None
+    if value <= 0:
+        return None
+    return value
+
+
+# Tinfoil enclaves reject completion caps below this floor, so a client that
+# sends a lower ``X-Routstr-Max-Tokens`` is rejected rather than silently
+# priced off the (unreachable) smaller cap.
+_EHBP_MIN_MAX_TOKENS = 64_000
```
jeroenubbink, line 874: A Tinfoil-specific constant in the provider-neutral proxy layer, applied to every EHBP request regardless of provider, with no settings knob. `ConfidentialInferenceProfile` is already how per-provider EHBP policy is carried — `proxy_only_headers` lives there.

routstr/proxy.py
```diff
@@ -831,6 +850,50 @@ async def get_bearer_token_key(
         raise
 
 
+def _parse_ehbp_max_tokens(raw: str) -> int | None:
```
jeroenubbink, line 853: This collapses absent, unparseable and non-positive into the same `None`, so `_validated_ehbp_max_tokens` has to repeat the emptiness check just to tell absent from invalid. It has no other caller. One function returning `int | None` and raising on invalid input covers both.

routstr/proxy.py
```diff
@@ -273,9 +273,22 @@ async def _proxy(
...
+    ehbp_max_tokens: int | None = None
     if is_ehbp:
+        # The body is HPKE-sealed/opaque, so the SDK also surfaces the client's
+        # completion cap ("max_tokens") via X-Routstr-Max-Tokens. This lets the
+        # proxy size the required balance down without decrypting the body.
+        # A present header below the enclave's floor is rejected outright.
         request_body_dict = {}
         model_id = headers.get("x-routstr-model", "")
+        try:
+            ehbp_max_tokens = _validated_ehbp_max_tokens(
```
jeroenubbink, line 285: This runs before the missing-`X-Routstr-Model` check below, so a request missing both gets only the max-tokens error, and the client needs a second round trip to discover the model header was missing all along. The model header is the harder precondition — validate it first.

routstr/upstream/ehbp.py
```diff
@@ -68,6 +72,7 @@ def _normalize_upstream_model_id(model_id: str | None) -> str:
 _PROXY_ONLY_HEADERS = frozenset(
     {
         "x-routstr-model",
+        _MAX_TOKENS_HEADER.lower(),
```
jeroenubbink, line 75: Stripped only on the EHBP forward path. `BaseUpstreamProvider.prepare_headers` removes host, content-length, refund-lnurl, key-expiry-time, x-cashu and auth — so on a plain request (the same SDK falling back to unencrypted mode) this header is forwarded verbatim to OpenRouter/OpenAI.

`x-routstr-model` has the same latent leak today; this adds a second proxy-only header without consolidating the two divergent lists into one source of truth.

routstr/upstream/ehbp.py
```diff
@@ -52,6 +52,10 @@
 _ENCLAVE_URL_HEADER = "X-Tinfoil-Enclave-Url"
 _REQUEST_USAGE_HEADER = "X-Tinfoil-Request-Usage-Metrics"
 _RESPONSE_USAGE_HEADER = "X-Tinfoil-Usage-Metrics"
+# Client supplies the completion cap via this header on EHBP (opaque-body)
+# requests so the proxy can size the required balance without decrypting the
+# body. It is stripped before the request is forwarded to the enclave.
+_MAX_TOKENS_HEADER = "X-Routstr-Max-Tokens"
```
jeroenubbink, line 58: `docs/ehbp-proxy-support.md` defines the EHBP request contract and `docs/tinfoil-direct-integration.md` enumerates what `_strip_proxy_headers()` removes. Neither mentions this header. Since the client-side note asks a separate repo to start emitting it, an SDK author following the docs will send the user's real `max_tokens` and get 400s.

tests/unit/test_coverage_proxy.py
```diff
@@ -7,6 +7,8 @@
 
 import pytest
 from fastapi import HTTPException
+from fastapi.responses import Response
+from unittest.mock import AsyncMock, MagicMock, patch
```
jeroenubbink, line 11: `ruff check` reports I001 here — stdlib `unittest.mock` after third-party `fastapi`.

tests/unit/test_coverage_proxy.py
```diff
@@ -64,6 +66,176 @@ def test_parse_json_rejects_non_integer_max_tokens() -> None:
     assert exc_info.value.status_code == 400
 
 
+# ===========================================================================
+# _parse_ehbp_max_tokens
+# ===========================================================================
+
+def test_parse_ehbp_max_tokens_valid() -> None:
+    """A numeric header value is parsed to an int."""
+    from routstr.proxy import _parse_ehbp_max_tokens
```
jeroenubbink, line 75: These import and assert on `_parse_ehbp_max_tokens`, `_validated_ehbp_max_tokens` and `_EHBP_MIN_MAX_TOKENS` directly. Repo convention is to test through the public API, and the behaviour is already reachable through the request path (as the 400 tests below show), so these pin an internal shape that breaks on any refactor of the two helpers into one.

The module docstring still says the file tests "request parsing, model extraction, and routing helpers", which no longer describes what's here.

tests/unit/test_coverage_proxy.py
```diff
@@ -64,6 +66,176 @@ def test_parse_json_rejects_non_integer_max_tokens() -> None:
...
+        patch("routstr.proxy.forward_ehbp_request", AsyncMock(return_value=Response(status_code=200))),
+    ):
+        response = await _proxy(
+            _ehbp_request({}),
+            "v1/chat/completions",
+            MagicMock(),
+        )
+
+    assert response.status_code == 200
+    assert seen["max_tokens"] is None
```
jeroenubbink, line 236: Two gaps worth closing while you're here: nothing exercises `forward_ehbp_x_cashu_request`, which consumes the same discounted `max_cost_for_model` and mints change against it; and nothing asserts the header is ignored on a non-EHBP request, which is the invariant keeping it from becoming a second billing input on the plain path.

jeroenubbink (changes requested):
The 402s are real and the diagnosis is right, but the cap has to be enforceable before it can
size the bill.

**The header sizes the reservation and nothing enforces it.** It's added to
`_PROXY_ONLY_HEADERS`, so it's stripped before the enclave — the sealed body's `max_tokens`
governs generation and we never see it. Settlement then clamps to
`min(actual, max_cost_for_model)`, so claiming the floor is strictly dominant: small hold,
bill capped at that hold, full-length generation served anyway. Roughly 20x on a 393k-context
model. The 64,000 floor is the only thing containing that.

I went looking for where 64,000 comes from and couldn't find it — nothing in
`tinfoilsh/confidential-model-router`, nothing in the EHBP spec (which defines only
`Ehbp-Encapsulated-Key` and `Ehbp-Response-Nonce`), nothing in our docs. Where's it from? If
it came out of a conversation with Tinfoil, that belongs in the description.

I also checked whether we could just shrink the hold without trusting the client. We can't:
`/v1/models` publishes no output cap, the enclaves serve `--max-model-len` equal to the
advertised window, and their `resolve_max_tokens` resolves an absent `max_tokens` to
`max_model_len - prompt_len`. The current full-window hold is arithmetically correct, so your
instinct is right — the cap has to come from the client. It just has to bind.

**Suggestion.** `forward_with_trailer` already takes `max_response_bytes` and aborts the
response when it's exceeded (fixed 25 MB default, no caller overrides it). Derive that
per-request from the claimed cap and the header becomes a contract we enforce ourselves:
claim 64k, generate 400k, connection dropped, still owe the discounted hold. Use a generous
bytes-per-token ratio so honest clients are never cut off.

Better version as an upstream ask: `resolve_max_tokens` already clamps with `.min()`, so a
proxy-supplied bound is one more argument, plumbed from a cleartext header the way
`manager/proxy.go` already handles `X-Tinfoil-Request-Usage-Metrics`. Truncation beats an
aborted connection, and there's precedent — PR 385 in `tinfoilsh/confidential-model-router`
added the served model to usage metrics for exactly this consumer.

Once it binds, the floor can become `min(claimed, model_max)`, which also fixes today's hard
400 for someone asking for 1024 tokens — the case this set out to help.

Four independent blockers and the smaller items are inline.

PR #672, opened 2026-08-17
feat: negative cache for terminal Cashu redemption failures

9qeklajc:
## Problem

Log analysis (08-12 → 08-17) showed 9 dead tokens generating thousands of redemption attempts per day. One token alone hit redemption ~3,300×/day. Root cause: when `credit_balance` fails in `validate_bearer_key`, the session rolls back — no `api_keys` row, no failure state, nothing persisted. The next request with the same dead token repeats the entire mint swap attempt from scratch. A polling client (e.g. `GET /v1/wallet/info`, which redeems via the auth dependency) plus one already-spent token = unbounded mint calls forever, concentrated on mints that are already flapping.

## Solution

In-memory negative cache (`routstr/redemption_cache.py`) keyed by the SHA-256 token hash:

- **Only terminal codes cached**: `cashu_token_already_spent`, `invalid_cashu_token`, `cashu_token_zero_value`, `cashu_token_swap_fees_exceed_amount`. These are permanent properties of the token — retrying can never succeed.
- **Transient failures never cached** (mint unreachable, rate-limited, cooldown) so a brief mint outage cannot poison valid tokens.
- **Cache hit returns the byte-identical error envelope** the original mint-backed failure produced (same status, type, code, message) — no client-visible behavior change beyond latency.
- **Bounded**: 10k entries LRU, 24h TTL, memory-only. Deliberately not a DB row — persisting failed tokens would let an attacker fill the database for free.

Wiring in `routstr/auth.py`:
- cache check after the existing-key lookup misses, before creating the token entry
- cache write in the `credit_balance` failure path, after classification


9qeklajc:
Moved the persistent cross-mint swap reconciliation work (d5528682) out of this PR into its own draft: #678. Force-push was not available in this session, so it is removed here via revert (01ecd2f2) — the PR diff is back to the negative-cache change only.

PR #678, opened 2026-08-18
feat: persistent reconciliation for interrupted cross-mint swaps

9qeklajc:
Split out of #672 — independent concern, its own review surface.

## Problem

A cross-mint swap that fails between melt dispatch and the balance credit surfaces as `TokenConsumedError` ("outcome requires reconciliation") and leaves real money stranded: the source proofs may be spent and the Lightning payment may still settle, but nothing ever completes the swap or credits the user.

### DIFF
--- a/migrations/versions/c5d7e9f1a3b5_add_pending_swaps_table.py
+++ b/migrations/versions/c5d7e9f1a3b5_add_pending_swaps_table.py
@@ -1,0 +1,46 @@
+"""add pending_swaps table
+
+Revision ID: c5d7e9f1a3b5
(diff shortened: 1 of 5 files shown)


9qeklajc:
Closing: superseded. Cross-mint swaps were removed entirely in cc55868e ("harden cashu path"); tokens from non-trusted mints are now rejected with UntrustedSourceMintError instead of being swapped, so the melt/mint reconciliation path this PR adds no longer has a target. The remaining ambiguous-outcome case (same-mint split() connection failure raising TokenConsumedError) is a different, smaller problem and would need its own PR.

PR #680, opened 2026-08-20
fix(payment): close prompt-estimation gaps for Responses API, tools, and original-detail images

sh1ftred:
## Problem

`calculate_discounted_max_cost()` — the routine that trims the reserved max cost based on the request's token caps — under-reserved against what the upstream actually bills, in two ways not covered by main's whole-body prompt estimation (`estimate_prompt_tokens`):

```text
image estimate
-   detail in {low, auto} only, 765-token cap
+   detail="original" -> 32x32px patch pricing, 30k-patch bound
+   Responses input_image now wired into the reservation (was 0 tokens)
completion discount
-   max_tokens only
+   largest of max_tokens | max_completion_tokens | max_output_tokens
```

## Fixes (per commit)

1. **Patch-based pricing for `detail="original"`** — the tiled estimator capped every non-low image at 765 tokens, while a 2048×2048 original-detail image bills ~4,900 tokens. Adds `_calculate_original_image_tokens` (patch count × 1.2, bounded by the 30,000-patch API rejection limit; conservative 36,000-token worst case for `file_id` refs whose dimensions can't be fetched).

2. **Honor `max_completion_tokens`** — modern chat clients (newer OpenAI SDKs) send `max_completion_tokens` instead of `max_tokens`; previously such requests got no completion discount and reserved the model's full worst-case `max_completion_cost`. Completion caps are now collected from all three spellings (`max_tokens`, `max_completion_tokens`, `max_output_tokens`) and the reservation is trimmed by the **largest** valid one: upstream precedence between the fields varies by provider, so reserving against a smaller declared cap could under-cover what the upstream bills.

3. **Wire Responses input images + fix original-detail fallbacks** (review follow-up):
   - `estimate_image_tokens_from_input` is now actually **called** on `body["input"]` in `calculate_discounted_max_cost` — previously it was defined but unreachable, so a Responses request with an `input_image` reserved **zero** image tokens.
   - Remote `original`-detail images are **fetched and measured** rather than blindly reserving the 36,000-token worst case (a 512×512 image reserved ~117× its real cost); the worst case is only a fallback for failed fetches / `file_id` refs.
   - Broken/undecodable data URLs now reserve the declared detail's worst case (36,000 for `original`) instead of the 85-token low-detail floor.

## Relationship to main

Main's `estimate_prompt_tokens` already counts prompt weight across the whole body (messages, Responses `input`, `instructions`, tools, `tool_calls`, replayed items) conservatively — so the prompt-side gaps this branch originally targeted are closed on main. This branch now only adds what main does **not** have: original-detail image patch pricing, Responses input-image wiring, and multi-field completion caps.

## Tests

`tests/unit/test_payment_helpers.py` — original-detail patch math, Responses input-image wiring regression, remote original-detail fetch, broken-data-URL fallback, completion-cap precedence, and the restored prompt-hiding security test.

Run: `.venv/bin/python -m pytest tests/unit/test_payment_helpers.py -q` (19 passed)


9qeklajc:
1. OpenAI defines `instructions` as context supplied to the model, but the code estimates only body["input"].
2. Valid Responses input items such as function_call_output are ignored
3. The estimator passes only image_url, dropping sibling detail and ignoring file_id.

jeroenubbink:
Small heads up: This PR touches the same lines as #669.

9qeklajc:
pls rebase

9qeklajc:
some small changes are needed:

## 1. The new function is never called

`estimate_image_tokens_from_input` is added at `routstr/payment/helpers.py:663`. Nothing in
`routstr/` calls it. Only the tests do.

Image counting only runs here (`helpers.py:234`):

```python
if isinstance(messages, list):
    image_tokens = await estimate_image_tokens_in_messages(messages)
```

Responses bodies have `input`, not `messages`. `/v1/responses` is a live route
(`routstr/proxy.py:398`). So a Responses request with an `input_image` still reserves zero
image tokens. That's the gap the PR title promises to close.

Fix: call it on `body.get("input")` in `calculate_discounted_max_cost`.

## 2. A security test was deleted

`test_discount_cannot_be_dodged_by_hiding_prompt_in_tools` is gone from
`tests/unit/test_payment_helpers.py`. Nothing replaces it.

It checked that a big prompt hidden in `tools`, `system`, or behind a fake `data:` prefix
still gets reserved for.

## 3. Description doesn't match the code

The PR body lists four fixes. Counting `instructions`, `tools`, `tool_calls`, and
`function_call` are already on main via `estimate_prompt_tokens`, not in this diff.

## 4. Remote `original` images over-reserve

`_estimate_input_image_tokens` returns the 36,000-token worst case for any non-data-URL with
`detail: "original"`, without trying to fetch it. Fine for `file_id`. For a plain https URL
it's worse than the existing chat path, which does fetch and measure. A 512x512 image
reserves ~117x its real cost.

Try the fetch first, fall back to 36,000.

## 5. Small stuff

- The `except` branch in `_estimate_input_image_tokens` returns `85` no matter the detail, so
  a broken data URL with `detail: "original"` gets low-detail pricing.

## litellm

litellm is already a dependency. It has some of this:

- `LiteLLMCompletionResponsesConfig._transform_response_input_param_to_chat_completion_message`
  turns a Responses `input` into chat messages. Run `input` through it and the existing
  `estimate_image_tokens_in_messages` just works — no second walker needed, and issue 1 goes away.
- `calculate_tiles_needed` and `resize_image_high_res` are the tile math here.
- `get_image_dimensions` reads dimensions from bytes and base64.
- `token_counter(model, messages, tools)` is a real tiktoken count, better than `chars // 3`.

It cannot do the main thing this PR is for. `_count_image_tokens` raises `ValueError` on any
detail other than `low`/`high`/`auto`, so `original` is unsupported.

So keep the patch pricing. Borrow the Responses transform and the tile math.

PR #694, opened 2026-08-25
fix(pricing): a rate that is not a number is not a price

jeroenubbink:
## What

`NaN`, `±inf`, a negative rate and an oversized JSON integer all reach this node as pricing,
and none of them is a price. This adds one definition of a usable rate and asks it on the paths
that read a rate from outside the node, at the one edge that stores a rate, and at the two
gates that decide whether a model is served and routed.

```python
def is_usable_rate(rate: float) -> bool:
    return math.isfinite(rate) and rate >= 0.0
```

Zero stays usable — free is a real price. This PR only asks whether a rate is *well-formed*.

## Why

Python's truthiness hides every one of these. `inf > 0` is true, so an infinite rate reads as
chargeable. `NaN > 0`, `NaN < 0` and `NaN == 0` are all false, so it slips past comparisons in
both directions. `json` parses the bare `NaN`/`Infinity` literals and overflows `1e999` to
`inf`, so an upstream catalog or a client request can carry one without trying.

What that cost, before this change:

- `NaN` reached the token math and raised `ValueError`; `inf` raised `OverflowError`. Both fire
  *after* the response is served, and the streaming handlers swallow them — the request went
  unbilled.
- A negative rate produced a negative cost, which settlement subtracts from the balance. It paid
  the caller to make requests.
- A non-finite upstream *cost* figure poisoned the proportional split (`inf / inf` is `NaN`).
  The broad handler absorbed it, so a request with a perfectly good total fell back to token
  estimation and was billed a fraction of what the upstream charged.
- An exchange quote of `0`, `NaN` or `-1` doesn't merely join the BTC/USD sample — it *wins* the
  `min()`, and prices every model on the node until the next refresh.
- An oversized integer raises `OverflowError`, not `ValueError`, which pydantic does not convert
  into a validation error: the admin edge answered a bad client value with a 500.

## Where the check now runs

| Surface | Behaviour |
|---|---|
| Catalog ingest | A malformed rate is not a resolved price, at either resolver rung. A bad cache rate falls back to the full input rate instead of riding along. |
| Admin write edge | 422, on every billable rate. One shared validator on the model write model, which is the only thing in the codebase that writes a stored price. |
| Served catalog | A row with a malformed stored rate is withheld. The operator's listing still shows it — that is the view it gets repaired from. |
| Routing map | The same rule when candidates are built, so a withheld model is not still routable. |
| Cost calculation | An unusable rate is no rate; an unusable upstream cost figure is no figure. (The path that bills an upstream's *reported total* is unaffected either way — rates only weight its breakdown.) |
| Exchange feed | An unusable quote is dropped, not adopted. If every quote is junk the node keeps the last rate it trusted. |

Two matching fixes for the same class of fault: one row whose stored JSON cannot be read no
longer takes out the whole catalog or the whole routing map, and a reply that *quotes* a
non-finite value now serializes instead of escaping as a 500 (`JSONResponse` encodes with
`allow_nan=False`). A 422 body gains a `request_id`, matching the other handlers.

## What is deliberately not here

Where a price *came from*, and whether a zero price should be billable, are a separate change.
This one never asks either question — a malformed rate is simply never used.

Two provider readers build a price from their own feed without asking (PPQ and Tinfoil). Both
are left alone here: PPQ's fix is entangled with how its price is labelled, and Tinfoil's
predates all of this. Neither can reach a request — the serve and route gates hold a malformed
price back wherever it was built.

Answers the malformed-rate review threads on #617, which this replaces along with a
provenance-only PR to follow.

## Testing

`make lint` clean · 1255 unit · 446 integration.

44 tests across the three rate-validation files plus 7 in the routing and generic-provider
suites. Written failing-first against real assertions, including two guard-rails against
over-rejection: a rate spelled as a numeric string is still accepted at the write edge, and a
zero-priced row is still served.

https://claude.ai/code/session_011cKHVF5LA7TR5QuYi6ErLM

### DIFF
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -12,6 +12,7 @@ from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from ..payment.models import (
+    REQUIRED_PRICING_FIELDS,
     _row_to_model,
     list_models,
 )
@@ -528,26 +529,32 @@ class ModelCreate(BaseModel):
     @field_validator("pricing")
     @classmethod
     def _validate_pricing(cls, value: dict[str, object]) -> dict[str, object]:
-        """Reject a malformed, non-finite or negative billable rate at the edge.
+        """Reject a rate that is malformed, non-finite, negative or not there.
 
         A present-but-invalid rate would otherwise slip through: a non-numeric
         string coerces to $0 on the read path (an unpriced-looking row), while a
         negative or ``NaN``/``inf`` value is truthy and reads back as a real
         price, so the model could be enabled and bill a nonsensical amount.
         Surfacing a 422 reports the client bug as a client bug instead of
-        persisting it. Absent rates and numeric strings (``"0.000005"``) stay
-        valid — the stored JSON accepts both.
+        persisting it. Numeric strings (``"0.000005"``) stay valid, and so does
+        an omitted auxiliary rate — the stored JSON accepts both.
         """
         for field in BILLABLE_PRICING_FIELDS:
-            raw = value.get(field)
-            if raw is None:
+            if field not in value:
+                # ``dict.get`` cannot tell this from an explicit ``null``, so
+                # both were skipped and a row that ``Pricing`` cannot parse was
+                # written — and then raised out of the response that reads it
+                # back, after the row had been committed.
+                if field in REQUIRED_PRICING_FIELDS:
+                    raise ValueError(f"{field} is required")
                 continue
             # The shared coercion also absorbs the OverflowError an oversized
             # integer raises, which pydantic does not convert into a validation
             # error — unhandled it escaped as a 500 for a bad client value.
-            if coerce_rate(raw) is None:
+            if coerce_rate(value[field]) is None:
                 raise ValueError(
-                    f"{field} must be a finite, non-negative number, got {raw!r}"
+                    f"{field} must be a finite, non-negative number, "
+                    f"got {value[field]!r}"
                 )
         return value
 
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -35,6 +35,7 @@ from .db import (
 from .db import (
     store_cashu_transaction_with_retry as store_cashu_transaction,
 )
+from .exceptions import json_compliant
 from .log_manager import log_manager
 from .logging import get_logger
 from .provider_slugs import allocate_unique_provider_slug
@@ -1273,8 +1274,13 @@ async def get_provider_models(provider_id: str) -> dict[str, object]:
                 "provider_type": provider.provider_type,
                 "base_url": provider.base_url,
             },
-            "db_models": [m.dict() for m in db_models],
-            "remote_models": [m.dict() for m in filtered_remote_models],
+            # This listing includes disabled models, so it is the one view that
+            # still carries a row the served-catalog backstop holds back —
+            # including one whose stored rate is not a usable number. The
+            # encoder would report that rate as `null`, indistinguishable from a
+            # missing one; show the operator the value that needs fixing.
+            "db_models": [json_compliant(m.dict()) for m in db_models],
+            "remote_models": [json_compliant(m.dict()) for m in filtered_remote_models],
         }
 
 
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -12,11 +12,10 @@ from sqlmodel import select
 from sqlmodel.ext.asyncio.session import AsyncSession
 
 from ..payment.models import (
-    BILLABLE_PRICING_FIELDS,
     _row_to_model,
-    is_usable_rate,
     list_models,
 )
+from ..payment.rates import BILLABLE_PRICING_FIELDS, is_usable_rate
 from ..proxy import refresh_model_maps, reinitialize_upstreams
 from ..wallet import fetch_all_balances, send_token, token_mint_url
 from . import vault
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -682,9 +682,13 @@ async def get_provider_model(provider_id: str, model_id: str) -> dict[str, objec
             raise HTTPException(
                 status_code=404, detail="Model not found for this provider"
             )
-        return _row_to_model(
-            row, apply_provider_fee=False, provider_fee=provider.provider_fee
-        ).dict()  # type: ignore
+        # Same duty as the listing this view is opened from: a stored rate that
+        # is not a usable number must be shown as it is, not encoded as `null`.
+        return json_compliant(  # type: ignore[return-value]
+            _row_to_model(
+                row, apply_provider_fee=False, provider_fee=provider.provider_fee
+            ).dict()
+        )
 
 
 @admin_router.delete(
--- a/routstr/core/admin.py
+++ b/routstr/core/admin.py
@@ -15,7 +15,7 @@ from ..payment.models import (
     _row_to_model,
     list_models,
 )
-from ..payment.rates import BILLABLE_PRICING_FIELDS, is_usable_rate
+from ..payment.rates import BILLABLE_PRICING_FIELDS, coerce_rate
 from ..proxy import refresh_model_maps, reinitialize_upstreams
 from ..wallet import fetch_all_balances, send_token, token_mint_url
 from . import vault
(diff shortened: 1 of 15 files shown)


routstr/payment/cost_calculation.py
```diff
@@ -267,9 +268,22 @@ async def calculate_cost(
...
+
+    # An unusable rate is not "no pricing" to Python's truthiness: `NaN` and a
+    # negative float are both truthy, so they sailed past this gate — the one
+    # guard meant to catch a rate that cannot be billed on — and reached the
+    # token math, which raises `ValueError` on `NaN` and `OverflowError` on
+    # `inf` *after* the response was served (the streaming handlers swallow
+    # that, so the request goes unbilled), while a negative produced a negative
+    # charge. Ask whether each rate is usable rather than whether it is truthy.
+    rates = (input_rate, output_rate, cache_read_rate, cache_creation_rate)
+    if not all(is_usable_rate(rate) for rate in rates) or not (
```
9qeklajc: This still contradicts the PR’s ‘zero remains usable’ rule. `has_usable_pricing()` admits zero-priced and partially free models, but the remaining truthiness check sends either zero component to `MaxCostData` and charges the full reservation.

I reproduced it with `prompt=0.0`, `completion=0.0`, 1,000/500 tokens, and `max_cost=1234`: the result was `MaxCostData(total_msats=1234)`.

Could this rely only on the explicit predicate?

```python
if not all(is_usable_rate(rate) for rate in rates):
    ...
```

Please add billing tests for `0/0`, `0/paid`, and `paid/0`; the current free-price test only proves the model is listed, not what the request is charged.

routstr/upstream/pricing_resolver.py
```diff
@@ -65,11 +65,23 @@ def estimate_context_length(model_id: str) -> int:
...
+    cleanly into a rate that credits the caller. An oversized integer raises
+    ``OverflowError`` rather than ``ValueError``, so that is caught too.
+    """
+    # Lazy, like the litellm lookup below: the resolver stays import-light.
+    from ..payment.models import is_usable_rate
+
     try:
-        return float(value)  # type: ignore[arg-type]
-    except (TypeError, ValueError):
+        parsed = float(value)  # type: ignore[arg-type]
```
9qeklajc: We need to reject booleans before coercion here. Python treats `bool` as numeric, so `_as_float(True)` currently returns `1.0`; the resulting `Pricing` also passes `has_usable_pricing()`.

```python
if isinstance(value, bool):
    return None
```

The LiteLLM branch has the same issue because `isinstance(True, (int, float))` is true. Ideally this coercion plus validation lives in one shared, dependency-neutral helper so OpenRouter, LiteLLM, native catalog values, and stored rows cannot drift.

routstr/payment/price.py
```diff
@@ -12,16 +12,68 @@
...
+            extra={
+                "error": str(e),
+                "error_type": type(e).__name__,
+                "exchange": exchange,
+                "quote": repr(raw),
+            },
+        )
+        return None
+
+    if not is_usable_rate(price) or price <= 0:
```
9qeklajc: This rejects exact zero, but an extremely small positive quote still passes and then underflows when converted to the per-sat price. Verified: `_parse_quote("1e-320", ...)` returns `1e-320`, and `1e-320 / 100_000_000` is `0.0`.

Because aggregation uses `min()`, one malformed feed wins over every healthy quote and can poison the node-wide conversion rate. At minimum, please validate that the derived sats price remains finite and positive; a plausible BTC/USD range or outlier check would make this safer still.

routstr/core/admin.py
```diff
@@ -520,6 +526,36 @@ class ModelCreate(BaseModel):
...
+        A present-but-invalid rate would otherwise slip through: a non-numeric
+        string coerces to $0 on the read path (an unpriced-looking row), while a
+        negative or ``NaN``/``inf`` value is truthy and reads back as a real
+        price, so the model could be enabled and bill a nonsensical amount.
+        Surfacing a 422 reports the client bug as a client bug instead of
+        persisting it. Absent rates and numeric strings (``"0.000005"``) stay
+        valid — the stored JSON accepts both.
+        """
+        for field in BILLABLE_PRICING_FIELDS:
+            raw = value.get(field)
```
9qeklajc: `value.get(field)` makes an absent key and an explicit `null` indistinguishable, and both are skipped. I verified that `ModelCreate` accepts both `{"prompt": null, "completion": 2e-6}` and a pricing object with `prompt` omitted, even though prompt/completion are required downstream.

Could we distinguish required fields from optional auxiliary fields here? Required `prompt`/`completion` should reject missing or `null`; optional fields may be absent, but an explicitly supplied `null` should not silently pass as a valid rate.

9qeklajc:
🔴 **Zero is accepted for serving/routing but still treated as missing during billing.** A free or partially free rate falls into `MaxCostData` and charges the full reservation. I reproduced this with a `0/0` model: `has_usable_pricing=True`, but a request with `max_cost=1234` returned `total_msats=1234`.

🔴 **Boolean catalog values are still accepted as prices.** Python converts `True` to `1.0`; I reproduced `_as_float(True) == 1.0`, and Pydantic preserves it as a usable rate. A malformed catalog field can therefore become `$1/token` instead of failing closed.

There are two smaller gaps worth closing while this code is already being hardened: a tiny positive exchange quote can underflow the derived sats price to zero, and missing/`null` required pricing fields pass the admin validator.

On cleanliness: the core idea is good, but the implementation is much larger and more repetitive than it needs to be. Most of the 1,725 added lines are tests, with repeated long explanations and bespoke HTTP fakes. I’d move the rate coercion/predicate into a small dependency-neutral module, parameterize more of the matrix, and keep the detailed failure history in one authoritative docstring rather than repeating it at each boundary.

9qeklajc:
good catch .. one thing that makes such a small PR too big are comments in code, in some places it's just redundant or shouldn't be that long, just a simple 1-2 line comment is enough if it's really needed .. code should mostly be self explained

routstr/payment/cost_calculation.py
jeroenubbink: Fixed in `337be48`. You're right — the truthiness half of the gate stayed and contradicted the PR's own rule; it now asks only `all(is_usable_rate(rate) for rate in rates)`.

Billing tests added for 0/0, 0/paid and paid/0.

routstr/upstream/pricing_resolver.py
jeroenubbink: Fixed in `dff164b`, on both rungs and in the catalog import filter.

Your shared-helper point is the right one: `is_usable_rate` lived in `payment/models.py` (httpx + fastapi + DB + settings), so all five consumers reached it through a deferred import. `payment/rates.py` now holds the predicate, the field tuple and one shared `coerce_rate` (`6318a3b`) — five lazy imports gone, booleans rejected once instead of three times. Worth noting `_parse_quote` and `_coerce_usd` already rejected them; that disagreement was the actual defect.

One deliberate widening: a numeric string is now a rate on the litellm rung too, which previously required a float outright.

routstr/payment/price.py
jeroenubbink: Fixed in `027bb3d`: the quote must be positive *after* the sats conversion, not just before it. `SATS_PER_BTC` is named now that it has two call sites.

I left the plausible-range check out. A hardcoded BTC/USD band is policy that ages badly, and an outlier test across quotes is a real idea but a different change — happy to file it.

routstr/core/admin.py
jeroenubbink: Fixed in `59bf8f4`, and it was worse than a silent pass: `Pricing` declares `prompt`/`completion` without a default, so the row is committed and then the response that reads it back raises — a 500 after the write has already landed.

Required rates now reject both missing and `null`; auxiliary rates may be absent but not `null`. Which are required is derived from `Pricing.__fields__` rather than restated, so the two can't drift.

jeroenubbink:
All four fixed, red-first, head `76d7b8f`. 1262 unit / 451 integration, lint clean, each commit type-checks alone.

On size: the coercion and the predicate moved to `payment/rates.py`, which imports only `math`, and the failure history now lives in that one module instead of at each boundary. The billing matrix is parameterised and the longest test docstrings are cut. I left the two HTTP fakes — they answer different endpoints, and merging them costs more indirection than it saves — and left 5–7 line docstrings alone, since that is already the style in `test_upstream_generic.py` and `test_algorithm.py`.

routstr/core/admin.py
```diff
@@ -520,6 +526,38 @@ class ModelCreate(BaseModel):
     enabled: bool = True
     forwarded_model_id: str | None = None
 
+    @field_validator("pricing")
+    @classmethod
+    def _validate_pricing(cls, value: dict[str, object]) -> dict[str, object]:
+        """Reject a rate that is malformed, non-finite, negative or not there.
+
+        A present-but-invalid rate would otherwise slip through: a non-numeric
```
9qeklajc, line 534: I was digging into the code lately and saw some of these in some places, it fine to have comments and docs in code but not in every line .. maybe that helps ai later to understand what that did, but that's why we have the tests, they could be more detailed  ..

routstr/core/admin.py
```diff
@@ -520,6 +526,38 @@ class ModelCreate(BaseModel):
...
+        string coerces to $0 on the read path (an unpriced-looking row), while a
+        negative or ``NaN``/``inf`` value is truthy and reads back as a real
+        price, so the model could be enabled and bill a nonsensical amount.
+        Surfacing a 422 reports the client bug as a client bug instead of
+        persisting it. Numeric strings (``"0.000005"``) stay valid, and so does
+        an omitted auxiliary rate — the stored JSON accepts both.
+        """
+        for field in BILLABLE_PRICING_FIELDS:
+            if field not in value:
+                # ``dict.get`` cannot tell this from an explicit ``null``, so
```
9qeklajc, line 544: for example this, ai do this in general way too offen and unnecessarly ... please only when needed

9qeklajc: and everywhere else

PR #696, opened 2026-08-26
Reduce request latency

9qeklajc:
## Summary

Reuse upstream connections, reduce reservation queries, instrument settlement latency, and queue file logging.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -905,6 +905,47 @@ async def pay_for_request(
             extra={"reservation_id": reservation.release_id},
         )
 
+    try:
+        await _validate_reservation_snapshot(key, reservation, session)
+    except BaseException:
+        released = False
+        try:
+            released = await _transition_reservation_to_released(
+                reservation,
+                session,
+                decrement_requests=True,
+                idempotent_success=True,
+            )
+        except BaseException:
+            try:
+                await session.rollback()
+            except BaseException:
+                pass
+
+        if not released:
+            try:
+                async with create_session() as cleanup_session:
+                    released = await _transition_reservation_to_released(
+                        reservation,
+                        cleanup_session,
+                        decrement_requests=True,
+                        idempotent_success=True,
+                    )
+            except BaseException:
+                logger.exception(
+                    "Failed to release invalid billing reservation",
+                    extra={"reservation_id": reservation.release_id},
+                )
+
+        if not released:
+            logger.error(
+                "Invalid billing reservation could not be released",
+                extra={"reservation_id": reservation.release_id},
+            )
+            await _stop_reservation_heartbeat(reservation.release_id)
+            _clear_current_reservation(reservation)
+        raise
+
     logger.info(
         "Payment processed successfully",
         extra={
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -1778,6 +1778,7 @@ async def adjust_payment_for_tokens(
 ) -> dict:
     """Settle payment while exposing latency for every import path."""
     started = time.perf_counter()
+    key_log_hash = key.hashed_key[:8] + "..."
     succeeded = False
     try:
         result = await _adjust_payment_for_tokens(
@@ -1795,7 +1796,7 @@ async def adjust_payment_for_tokens(
         logger.info(
             "Payment settlement finished",
             extra={
-                "key_hash": key.hashed_key[:8] + "...",
+                "key_hash": key_log_hash,
                 "model": response_data.get("model", "unknown"),
                 "settlement_duration_ms": round(
                     (time.perf_counter() - started) * 1000, 2
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -1311,7 +1311,7 @@ async def _charge_reservation_rows(
     return True
 
 
-async def adjust_payment_for_tokens(
+async def _adjust_payment_for_tokens(
     key: ApiKey,
     response_data: dict,
     session: AsyncSession,
@@ -1765,6 +1765,44 @@ async def adjust_payment_for_tokens(
             )
     # All calculate_cost variants are handled above.
     raise AssertionError("Unreachable: unhandled calculate_cost result")
+
+
+async def adjust_payment_for_tokens(
+    key: ApiKey,
+    response_data: dict,
+    session: AsyncSession,
+    deducted_max_cost: int,
+    model_obj: "Model | None" = None,
+    provider_fee: float | None = None,
+    reservation_snapshot: ReservationSnapshot | None = None,
+) -> dict:
+    """Settle payment while exposing latency for every import path."""
+    started = time.perf_counter()
+    succeeded = False
+    try:
+        result = await _adjust_payment_for_tokens(
+            key,
+            response_data,
+            session,
+            deducted_max_cost,
+            model_obj,
+            provider_fee,
+            reservation_snapshot,
+        )
+        succeeded = True
+        return result
+    finally:
+        logger.info(
+            "Payment settlement finished",
+            extra={
+                "key_hash": key.hashed_key[:8] + "...",
+                "model": response_data.get("model", "unknown"),
+                "settlement_duration_ms": round(
+                    (time.perf_counter() - started) * 1000, 2
+                ),
+                "settlement_succeeded": succeeded,
+            },
+        )
 
 
 async def periodic_key_reset() -> None:
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -642,8 +642,8 @@ async def get_billing_key(key: ApiKey, session: AsyncSession) -> ApiKey:
 
 async def pay_for_request(
     key: ApiKey, cost_per_request: int, session: AsyncSession
-) -> int:
-    """Process payment for a request."""
+) -> ReservationSnapshot:
+    """Reserve funds and return the durable identity for this request."""
     # Ensure cost_per_request is at least the minimum allowed request cost
     cost_per_request = max(cost_per_request, settings.min_request_msat)
 
@@ -929,7 +929,7 @@ async def pay_for_request(
         },
     )
 
-    return cost_per_request
+    return reservation
 
 
 async def revert_pay_for_request(
--- a/routstr/core/logging.py
+++ b/routstr/core/logging.py
@@ -28,7 +28,12 @@ DO NOT modify or remove these messages without updating the usage tracking logic
    - The 'max_cost_for_model' field is extracted for refund calculation
    - Must include 'max_cost_for_model' in extra dict
 
-6. Any ERROR level logs with "upstream" in the message
+6. "Payment settlement finished" (INFO) - routstr/auth.py and routstr/upstream/ehbp.py
+   - Emitted once per settlement attempt, including EHBP settlements
+   - Carries 'settlement_duration_ms' and 'settlement_succeeded'; the EHBP
+     emitter adds 'settlement_type'
+
+7. Any ERROR level logs with "upstream" in the message
    - Used to count upstream provider errors
    - Helps identify service reliability issues
 
@@ -44,6 +49,7 @@ import os
 import queue
 import re
 import sys
+import threading
 import time
 import tomllib
 from datetime import datetime
@@ -131,18 +137,24 @@ class DailyRotatingFileHandler(logging.handlers.TimedRotatingFileHandler):
 
 
 class QueuedDailyRotatingFileHandler(logging.Handler):
-    """Queue records so request handlers never perform rotating-file I/O."""
+    """Move rotating-file I/O off request threads.
+
+    When both locks are needed, acquire the logging module lock before the
+    handler lock to match ``dictConfig``.
+    """
 
     _queue: queue.Queue[logging.LogRecord]
     _target: DailyRotatingFileHandler
     _listener: logging.handlers.QueueListener
     _drain_timeout_seconds = 5.0
+    _reopen_backoff_seconds = 5.0
 
     def __init__(self, filename: str, **kwargs: Any) -> None:
         super().__init__()
         self._filename = filename
         self._kwargs = kwargs
         self._stopped = True
+        self._next_open_attempt = 0.0
         self._open()
 
     def _open(self) -> None:
@@ -163,26 +175,68 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
         self._target = target
         self._listener = listener
         self._stopped = False
-        handler_list = getattr(logging, "_handlerList")
-        if not any(reference() is self for reference in handler_list):
-            getattr(logging, "_addHandlerRef")(self)
+        self._closed = False
+        with getattr(logging, "_lock"):
+            handler_list = getattr(logging, "_handlerList")
+            # This wrapper owns the target's shutdown and lock ordering.
+            handler_list[:] = [
+                reference for reference in handler_list if reference() is not target
+            ]
+            if not any(reference() is self for reference in handler_list):
+                getattr(logging, "_addHandlerRef")(self)
+
+    def _reopen_locked(self) -> bool:
+        """Reopen using the lock order required by ``dictConfig``."""
+        with getattr(logging, "_lock"):
+            self.acquire()
+            try:
+                if not self._stopped:
+                    return True
+                if time.monotonic() < self._next_open_attempt:
+                    return False
+                try:
+                    self._open()
+                except Exception:
+                    self._next_open_attempt = (
+                        time.monotonic() + self._reopen_backoff_seconds
+                    )
+                    raise
+                return True
+            finally:
+                self.release()
 
     def setFormatter(self, fmt: logging.Formatter | None) -> None:
         super().setFormatter(fmt)
         self._target.setFormatter(fmt)
 
-    def emit(self, record: logging.LogRecord) -> None:
-        self.acquire()
-        try:
+    def handle(self, record: logging.LogRecord) -> bool:
+        if not self.filter(record):
+            return False
+
+        while True:
+            self.acquire()
             try:
-                # uvicorn's dictConfig closes us mid-startup; reopen like FileHandler does.
-                if self._stopped:
-                    self._open()
-                self._queue.put_nowait(copy.copy(record))
+                if not self._stopped:
+                    self.emit(record)
+                    return True
+            finally:
+                self.release()
+
+            try:
+                # Do not acquire the module lock while holding the handler lock.
+                if not self._reopen_locked():
+                    return False
             except Exception:
                 self.handleError(record)
-        finally:
-            self.release()
+                return False
+
+    def emit(self, record: logging.LogRecord) -> None:
+        try:
+            if not self._stopped:
+                self._queue.put_nowait(copy.copy(record))
+        except Exception:
+            # Handler.handle() does not catch exceptions raised by emit().
+            self.handleError(record)
 
     def flush(self) -> None:
         self.acquire()
@@ -195,7 +249,7 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
                 while self._queue.unfinished_tasks:
                     remaining = deadline - time.monotonic()
                     if remaining <= 0:
-                        return
+                        break
                     self._queue.all_tasks_done.wait(remaining)
             self._target.flush()
         finally:
@@ -212,19 +266,52 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
         self._listener._thread = None
         return True
 
+    def _close_retired_listener(
+        self,
+        listener: logging.handlers.QueueListener,
+        target: DailyRotatingFileHandler,
+    ) -> None:
+        def finish() -> None:
+            thread = listener._thread
+            if thread is not None:
+                thread.join()
+                listener._thread = None
+            try:
+                target.flush()
+            finally:
+                target.close()
+
+        threading.Thread(target=finish, daemon=True).start()
+
     def close(self) -> None:
-        # Held across the teardown so a concurrent emit cannot reopen us
-        # halfway through it. The listener thread never takes this lock.
+        # Stop the listener under the handler lock, then close the target outside
+        # it because FileHandler.close() also takes the logging module lock.
         self.acquire()
         try:
+            target = None
+            retired = None
             if not self._stopped:
+                listener = self._listener
+                current_target = self._target
+                stopped = self._stop_listener()
                 self._stopped = True
-                if self._stop_listener():
-                    self._target.flush()
-                    self._target.close()
+                if stopped:
+                    target = current_target
+                else:
+                    retired = (listener, current_target)
+                    sys.stderr.write(
+                        f"Logging listener for {self._filename} did not stop "
+                        f"within {self._drain_timeout_seconds}s; reopening on next record\n"
+                    )
         finally:
             self.release()
-            super().close()
+
+        if retired is not None:
+            self._close_retired_listener(*retired)
+        if target is not None:
+            target.flush()
+            target.close()
+        super().close()
 
 
 def get_package_version() -> str:
--- a/routstr/core/logging.py
+++ b/routstr/core/logging.py
@@ -130,16 +130,9 @@ class DailyRotatingFileHandler(logging.handlers.TimedRotatingFileHandler):
 
 
 class QueuedDailyRotatingFileHandler(logging.Handler):
-    """Queue records so request handlers never perform rotating-file I/O.
+    """Queue records so request handlers never perform rotating-file I/O."""
 
-    The handler reopens itself on the next record after being closed, the
-    way a plain ``FileHandler`` reopens its stream. That matters because
-    ``logging.config.dictConfig`` closes every live handler before applying a
-    new config, and uvicorn runs its own ``dictConfig`` *after* importing the
-    app module — so ``setup_logging``'s handlers are always closed once during
-    startup. Without the reopen the file log goes silent for the whole run.
-    """
-
+    _queue: queue.Queue[logging.LogRecord]
     _target: DailyRotatingFileHandler
     _listener: logging.handlers.QueueListener
 
@@ -147,16 +140,18 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
         super().__init__()
         self._filename = filename
         self._kwargs = kwargs
-        self._queue: queue.Queue[logging.LogRecord] = queue.Queue()
         self._open()
 
     def _open(self) -> None:
-        """Attach a fresh rotating file handler and start draining the queue."""
+        """Attach a fresh rotating file handler and start draining it."""
+        # A new queue per listener: QueueListener's stop sentinel is a shared
+        # singleton, so two listeners on one queue would steal each other's.
+        self._queue = queue.Queue()
         self._target = DailyRotatingFileHandler(self._filename, **self._kwargs)
         self._target.setFormatter(self.formatter)
         self._listener = logging.handlers.QueueListener(self._queue, self._target)
         self._listener.start()
-        self._closed = False
+        self._stopped = False
 
     def setFormatter(self, fmt: logging.Formatter | None) -> None:
         super().setFormatter(fmt)
@@ -165,33 +160,32 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
     def emit(self, record: logging.LogRecord) -> None:
         self.acquire()
         try:
-            if self._closed:
+            # uvicorn's dictConfig closes us mid-startup; reopen like FileHandler does.
+            if self._stopped:
                 self._open()
             self._queue.put_nowait(copy.copy(record))
         finally:
             self.release()
 
     def flush(self) -> None:
-        if self._closed:
+        if self._stopped:
             return
         self._queue.join()
         self._target.flush()
 
     def close(self) -> None:
+        # Held across the teardown so a concurrent emit cannot reopen us
+        # halfway through it. The listener thread never takes this lock.
         self.acquire()
         try:
-            if self._closed:
+            if self._stopped:
                 return
-            self._closed = True
-            listener, target = self._listener, self._target
+            self._stopped = True
+            self._listener.stop()
+            self._target.flush()
+            self._target.close()
         finally:
             self.release()
-
-        # Outside the lock: stop() waits for the listener thread, which needs
-        # to take the same lock on its way through the target handler.
-        listener.stop()
-        target.flush()
-        target.close()
 
 
 def get_package_version() -> str:
--- a/routstr/core/logging.py
+++ b/routstr/core/logging.py
@@ -130,15 +130,33 @@ class DailyRotatingFileHandler(logging.handlers.TimedRotatingFileHandler):
 
 
 class QueuedDailyRotatingFileHandler(logging.Handler):
-    """Queue records so request handlers never perform rotating-file I/O."""
+    """Queue records so request handlers never perform rotating-file I/O.
+
+    The handler reopens itself on the next record after being closed, the
+    way a plain ``FileHandler`` reopens its stream. That matters because
+    ``logging.config.dictConfig`` closes every live handler before applying a
+    new config, and uvicorn runs its own ``dictConfig`` *after* importing the
+    app module — so ``setup_logging``'s handlers are always closed once during
+    startup. Without the reopen the file log goes silent for the whole run.
+    """
+
+    _target: DailyRotatingFileHandler
+    _listener: logging.handlers.QueueListener
 
     def __init__(self, filename: str, **kwargs: Any) -> None:
         super().__init__()
-        self._target = DailyRotatingFileHandler(filename, **kwargs)
+        self._filename = filename
+        self._kwargs = kwargs
         self._queue: queue.Queue[logging.LogRecord] = queue.Queue()
+        self._open()
+
+    def _open(self) -> None:
+        """Attach a fresh rotating file handler and start draining the queue."""
+        self._target = DailyRotatingFileHandler(self._filename, **self._kwargs)
+        self._target.setFormatter(self.formatter)
         self._listener = logging.handlers.QueueListener(self._queue, self._target)
-        self._closed = False
         self._listener.start()
+        self._closed = False
 
     def setFormatter(self, fmt: logging.Formatter | None) -> None:
         super().setFormatter(fmt)
@@ -147,8 +165,9 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
     def emit(self, record: logging.LogRecord) -> None:
         self.acquire()
         try:
-            if not self._closed:
-                self._queue.put_nowait(copy.copy(record))
+            if self._closed:
+                self._open()
+            self._queue.put_nowait(copy.copy(record))
         finally:
             self.release()
 
@@ -164,15 +183,15 @@ class QueuedDailyRotatingFileHandler(logging.Handler):
             if self._closed:
                 return
             self._closed = True
+            listener, target = self._listener, self._target
         finally:
             self.release()
 
-        try:
-            self._listener.stop()
-            self._target.flush()
-            self._target.close()
-        finally:
-            super().close()
+        # Outside the lock: stop() waits for the listener thread, which needs
+        # to take the same lock on its way through the target handler.
+        listener.stop()
+        target.flush()
+        target.close()
 
 
 def get_package_version() -> str:
--- a/routstr/core/logging.py
+++ b/routstr/core/logging.py
@@ -37,9 +37,11 @@ If you need to modify these messages, ensure you also update the parsing logic i
 - routstr/core/log_manager.py
 """
 
+import copy
 import logging.config
 import logging.handlers
 import os
+import queue
 import re
 import sys
 import tomllib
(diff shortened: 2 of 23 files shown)


routstr/upstream/http_client.py
```diff
@@ -0,0 +1,38 @@
...
+    loop = asyncio.get_running_loop()
+    if (
+        _upstream_http_client is None
+        or _upstream_http_client.is_closed
+        or _upstream_http_client_loop is not loop
+    ):
+        previous = _upstream_http_client
+        _upstream_http_client = httpx.AsyncClient(
+            transport=httpx.AsyncHTTPTransport(retries=1),
+            timeout=None,
```
jeroenubbink: **Blocker.** `timeout=None` gives `pool=None` and `read=None`, and the transport inherits `DEFAULT_LIMITS` — `max_connections=100`, pool-wide, not per-origin.

So the 101st concurrent call hangs on the semaphore rather than raising `PoolTimeout`, holding a claimed reservation and renewing its lease while it waits. And with no read timeout, an upstream that goes quiet without closing keeps its slot forever.

Suggest an explicit `Limits(...)` for the worker's concurrency, a finite `pool` timeout so exhaustion is a 503 rather than a hang, and a finite `read` timeout.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,38 @@
...
+    """Return the process-wide client so upstream origins reuse pooled connections."""
+    global _upstream_http_client, _upstream_http_client_loop
+    loop = asyncio.get_running_loop()
+    if (
+        _upstream_http_client is None
+        or _upstream_http_client.is_closed
+        or _upstream_http_client_loop is not loop
+    ):
+        previous = _upstream_http_client
+        _upstream_http_client = httpx.AsyncClient(
```
jeroenubbink: Note. One client, one cookie jar: `_send_single_request` extracts `Set-Cookie` and `build_request` merges it back. A `__cf_bm` or sticky-LB cookie from one request is replayed on every later request to that origin, for every user, and the jar grows for the process lifetime. `cookies=None` doesn't opt out — clear the jar, or pass `cookies={}` per request.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,38 @@
...
+        if previous is not None and not previous.is_closed:
+            loop.create_task(previous.aclose())
+    return _upstream_http_client
+
+
+async def close_upstream_http_client() -> None:
+    """Close the shared connection pool during application shutdown."""
+    global _upstream_http_client, _upstream_http_client_loop
+    client = _upstream_http_client
+    _upstream_http_client = None
```
jeroenubbink: Blocker, small. The globals are cleared before the await, so anything calling `get_upstream_http_client()` during `aclose()` builds a client nothing will close — the periodic tasks are still winding down at that point. Clear after the await, or gate on a shutdown flag.

routstr/upstream/base.py
```diff
@@ -3161,7 +3199,6 @@ async def forward_request(
                             requested_model=original_model_id,
                             model_obj=model_obj,
                             reservation_snapshot=reservation_snapshot,
-                            client=client,
                             request_body=request_body,
                         )
```
jeroenubbink: **Blocker.** The messages branch got `except BaseException: await response.aclose(); raise` at line 3117; this branch and the Responses one didn't.

`handle_streaming_chat_completion` raises before returning a `StreamingResponse` on real paths — `get_reservation_snapshot` on an empty parsed body, any DB error in `create_session()`. Nothing closes `response`, so the connection leaves the pool permanently. Per-request that died with the request; against a 100-slot shared pool it accumulates.

routstr/upstream/base.py
```diff
@@ -3440,7 +3468,6 @@ async def forward_responses_request(
                         requested_model=original_model_id,
                         model_obj=model_obj,
                         reservation_snapshot=reservation_snapshot,
-                        client=client,
                         request_body=transformed_body,
                     )
```
jeroenubbink: Same gap — no guard, so a raise from `handle_streaming_responses_completion` leaks the connection.

routstr/upstream/base.py
```diff
@@ -3575,66 +3598,70 @@ async def forward_get_request(
...
+            response_headers.pop("content-length", None)
+            background_tasks = BackgroundTasks()
+            background_tasks.add_task(response.aclose)
+            return StreamingResponse(
+                response.aiter_bytes(),
+                status_code=response.status_code,
+                headers=response_headers,
+                background=background_tasks,
+            )
+        except Exception as exc:
```
jeroenubbink: **Blocker.** `except Exception` misses `CancelledError` (a `BaseException` since 3.8), and the `async with` that used to guarantee teardown is gone. A disconnect during `forward_upstream_error_response`'s `aread()` leaves `response` checked out for good.

This is the unauthenticated GET path — proxy.py routes both the Tinfoil attestation GETs and the fall-through unauthenticated branch here — so it drains the pool with no credentials. `try/finally`, or `except BaseException`.

routstr/upstream/base.py
```diff
@@ -2965,10 +3007,7 @@ async def forward_request(
             },
         )
 
-        client = httpx.AsyncClient(
-            transport=httpx.AsyncHTTPTransport(retries=1),
-            timeout=None,
-        )
+        client = get_upstream_http_client()
```
jeroenubbink: Worth investigating rather than a finding. Pooled reuse adds a failure the per-request clients couldn't hit: httpx holds idle connections for `keepalive_expiry=5.0`, and an upstream that drops an idle socket first makes the next `send()` raise `RemoteProtocolError` — which lands below as a 502 plus a failover.

`retries=1` doesn't cover that: it's consumed only in `_connect()`, for `ConnectError`/`ConnectTimeout`, never on an established stream. So as set it does nothing for the failure that reuse actually introduces. Worth measuring against the real upstreams, then either retrying `RemoteProtocolError` at the send site or dropping `retries` — better nothing than a setting that reads as protection and isn't.

routstr/upstream/base.py
```diff
@@ -97,6 +101,44 @@ async def _finalize_and_close_stream(
             await _aclose_if_needed(client)
 
 
+async def adjust_payment_for_tokens(
```
jeroenubbink: Note on shape. This shadows `..auth.adjust_payment_for_tokens` (imported here as `_adjust_payment_for_tokens`), so calls in this file get the wrapper and a caller importing from `..auth` silently gets none. All 15 call sites are here today, so nothing is missing — it just ties the instrumentation to the import path rather than the mechanism. Moving the span into `auth.adjust_payment_for_tokens` drops the alias, the shadow and the duplicated signature.

Also: EHBP settles via `finalize_ehbp_actual_cost_payment` / `finalize_ehbp_max_cost_payment`, neither of which passes through here — so encrypted-body requests contribute nothing to the new metric.

routstr/core/logging.py
```diff
@@ -127,6 +129,65 @@ def _cleanup_old_files(self) -> None:
...
+        self._target.setFormatter(self.formatter)
+        self._listener = logging.handlers.QueueListener(self._queue, self._target)
+        self._listener.start()
+        self._stopped = False
+
+    def setFormatter(self, fmt: logging.Formatter | None) -> None:
+        super().setFormatter(fmt)
+        self._target.setFormatter(fmt)
+
+    def emit(self, record: logging.LogRecord) -> None:
```
jeroenubbink: **Blocker.** `_open()` opens a file and starts a thread, and `emit` has no `try/except -> handleError`. `Handler.handle()` doesn't catch what `emit` raises, so a full disk or a missing `logs/` surfaces at the `logger.info(...)` call site — one of which is the settlement wrapper's `finally` in `upstream/base.py`, where it would replace the exception already propagating. Every stdlib file handler routes emit failures to `handleError` for this reason.

Also `_open()` sets `_stopped = False` last, so a raise on the line above wedges the handler permanently.

routstr/core/logging.py
```diff
@@ -127,6 +129,65 @@ def _cleanup_old_files(self) -> None:
...
+        finally:
+            self.release()
+
+    def flush(self) -> None:
+        if self._stopped:
+            return
+        self._queue.join()
+        self._target.flush()
+
+    def close(self) -> None:
```
jeroenubbink, line 286: **Blocker.** `close()` never chains to `super().close()`, which matters in exactly the case this class exists for. `dictConfig` -> `_clearExistingHandlers()` runs `shutdown()` then `del logging._handlerList[:]`. `emit` reopens, but `_open()` doesn't re-register (only `Handler.__init__` calls `_addHandlerRef`), so at exit `shutdown()` walks a list this handler isn't in — no flush — and `QueueListener` sets `daemon = True`, so the thread dies with the queue non-empty.

What's lost is what this file's docstring declares a contract: the `RESERVE` / `Payment processed successfully` lines that `usage_analytics_store.py` and `log_manager.py` parse.

routstr/core/logging.py
```diff
@@ -127,6 +129,65 @@ def _cleanup_old_files(self) -> None:
...
+        self.acquire()
+        try:
+            # uvicorn's dictConfig closes us mid-startup; reopen like FileHandler does.
+            if self._stopped:
+                self._open()
+            self._queue.put_nowait(copy.copy(record))
+        finally:
+            self.release()
+
+    def flush(self) -> None:
```
jeroenubbink, line 241: `flush()` touches `_stopped` and `_target` without `acquire()`, while `emit()` and `close()` hold it. `shutdown()` calls `flush()` then `close()`, so a flush blocked in `join()` can resume after `_target` is closed (`ValueError: I/O operation on closed file`), or join the old queue after `emit()` swapped in a new one and flush nothing.

`join()` is also unbounded: if the listener dies, `task_done()` stops and this hangs at exit. A timeout makes that a slow exit instead. (The unbounded queue on line 149 is the stdlib shape — leave it.)

jeroenubbink (changes requested):
Reviewed at `690dff0` against `main` `6d422ad`.

Right direction — an `AsyncClient` per request meant a TLS handshake per call. But a process-wide client moves the failure boundary from the request to the node, and the surrounding code still assumes the old isolation.

**1. The pool is a node-wide cap with no backpressure.** The transport inherits `DEFAULT_LIMITS` (`max_connections=100`, pool-wide, not per-origin) and `timeout=None` gives `pool=None` and `read=None`. The 101st concurrent call hangs on the semaphore instead of raising `PoolTimeout` — holding a claimed reservation and renewing its lease throughout. A stalled-but-open upstream stream never returns its slot, so ~100 of them make every provider unreachable, healthy ones included.

**2. Connections escape the pool three ways**, which turns that cap into an outage: the `except BaseException: aclose()` guard added to the messages branch at line 3117 is missing from the other two streaming branches, `forward_get_request` traded `async with` for an `except Exception` that misses `CancelledError`, and `close_upstream_http_client` clears its globals before awaiting.

**3. The log handler raises into billing, and drops records at exit.** `emit()` does file I/O with no `handleError`, and `Handler.handle()` doesn't catch — so a failed write surfaces at the `logger.info(...)` call site, one of which is the settlement wrapper's `finally`. Separately `close()` never chains to `super().close()`, so after `dictConfig` the handler is off `logging._handlerList`, `shutdown()` never flushes it, and the daemon listener dies with the queue non-empty — losing the records this file's docstring declares a parsing contract.

Inline notes: the shared cookie jar, whether `retries=1` is covering anything, and EHBP contributing nothing to the new metric.

9qeklajc:
The backend-test (3.11) job is failing on static type-check (mypy), not runtime tests:

- tests/unit/test_upstream_http_client.py:37 — `AsyncBaseTransport` has no attribute `_pool` [attr-defined]
- tests/unit/test_upstream_http_client.py:66 — function missing type annotation [no-untyped-def]
- routstr/core/logging.py:166,167 — module has no attribute `_handlerList` / `_addHandlerRef` [attr-defined]
- tests/unit/test_queued_logging.py:37,39,40,45 — missing attrs / annotations

`mergeable_state` is blocked. The PR builds and runs, but needs these type errors resolved before it can merge. Likely refactor to avoid reaching into httpcore internals (`. _pool`) and stdlib logging internals (`_handlerList`), and add the missing annotations in the new tests.

routstr/core/logging.py
```diff
@@ -127,6 +130,103 @@ def _cleanup_old_files(self) -> None:
...
+        super().setFormatter(fmt)
+        self._target.setFormatter(fmt)
+
+    def emit(self, record: logging.LogRecord) -> None:
+        self.acquire()
+        try:
+            try:
+                # uvicorn's dictConfig closes us mid-startup; reopen like FileHandler does.
+                if self._stopped:
+                    self._open()
```
jeroenubbink: **Lock-order inversion — deadlocks the whole process.**

`emit()` holds the handler lock and then takes `logging._lock`, via `DailyRotatingFileHandler.__init__` → `_addHandlerRef`, and again explicitly at `:168`. `dictConfig` takes them the other way: it holds `logging._lock` across `_clearExistingHandlers()` → `logging.shutdown()` → `h.acquire()`.

Any log call racing a `dictConfig` deadlocks, and since the handler lock is then never released, **every** subsequent log call in the process blocks — request coroutines park holding claimed reservations. The reopen branch runs whenever `_stopped` is True, which the comment at `:178` says is the normal startup state, so the window is every boot. The race test at `test_queued_logging.py:88` only exercises the handler lock, so it passes.

Rule to hold: never take the module lock while holding the handler lock.

routstr/upstream/base.py
```diff
@@ -3180,14 +3180,12 @@ async def forward_request(
                         )
                     finally:
                         await response.aclose()
-                        await client.aclose()
 
             if reservation_snapshot is None:
                 reservation_snapshot = await get_reservation_snapshot(key, session)
 
             background_tasks = BackgroundTasks()
             background_tasks.add_task(response.aclose)
```
jeroenubbink: **The reservation's only settlement lives in `background=`, which Starlette can skip.**

Starlette runs background tasks only when `stream_response` returns normally — a client disconnect or a `ReadTimeout` inside `aiter_bytes` skips both tasks, so the reservation is never settled or released and the connection is never returned to the pool. Under `timeout=None` the iterator could not raise; the new `read=300` makes this reachable.

Both siblings already guard it from inside the generator: `base.py:1333` and `:1776`, `finally: await asyncio.shield(_finalize_and_close_stream(...))`, commented *"Shielded so a client disconnect cannot cancel billing finalization or leak the upstream connection."*

Same block at `:3467`.

routstr/upstream/base.py
```diff
@@ -3067,18 +3063,21 @@ async def forward_request(
...
+                                requested_model=original_model_id,
+                                model_obj=model_obj,
+                                reservation_snapshot=reservation_snapshot,
+                                request_body=request_body,
+                            )
+                        except BaseException:
+                            await response.aclose()
+                            raise
                         background_tasks = BackgroundTasks()
                         background_tasks.add_task(response.aclose)
```
jeroenubbink: **`/v1/messages` streams have no close that survives a disconnect.**

This is the only `aclose` for the stream, and Starlette skips background tasks on a client disconnect. The generator's own `finally` at `:2274` settles the bill and closes nothing — no `aclose`, no shield, unlike `:1333` and `:1776`.

httpcore only reaps connections that are closed, expired or idle, so an ACTIVE one with an unread stream is held forever: 100 aborted streams exhaust the pool node-wide. Pre-PR this leaked a per-request client, which was GC-reclaimable.

routstr/upstream/base.py
```diff
@@ -3236,19 +3233,25 @@ async def forward_request(
...
                 error_message = "Network error while connecting to upstream service"
+                status_code = 502
             else:
                 error_message = f"Error connecting to upstream service: {error_type}"
+                status_code = 502
 
-            raise UpstreamError(error_message, status_code=502)
+            raise UpstreamError(error_message, status_code=status_code)
 
         except Exception as exc:
```
jeroenubbink, line 3521: **Deleted `client.aclose()` with nothing put back.**

This arm and `except httpx.RequestError` at `:3218` lost their `await client.aclose()` without gaining `await response.aclose()`, so any exception after `client.send()` strands a shared-pool connection permanently. Previously, closing the per-request client tore down its private pool; with a process-wide client the response close is the only reclaim.

Same at `:3497` and `:3533`. These two methods don't pre-initialise `response`, so a fix needs `response: httpx.Response | None = None` as at `:3583`.

routstr/upstream/ehbp.py
```diff
@@ -513,6 +513,35 @@ async def _release_failed_ehbp_charge(
...
+    succeeded = False
+    try:
+        result = await operation
+        succeeded = True
+        return result
+    finally:
+        logger.info(
+            "Payment settlement finished",
+            extra={
+                "key_hash": key.hashed_key[:8] + "...",
```
jeroenubbink: **`key` may be expired here, and the raise destroys the return value.**

`_claim_reservation_for_charge` calls `session.rollback()` and returns False (`auth.py:1246`), so the wrapped operation returns 0 normally — and this `finally` then lazy-loads `hashed_key` from a coroutine → `MissingGreenlet`. Being a bare `finally`, the raise **discards the `return 0`** and propagates to `proxy.py:711`'s `except Exception`, which reverts a reservation that was already charged.

The fix is already the pattern in this PR: the sibling wrapper captures `key_log_hash` before the `try` (`auth.py:1781`), and `finalize_ehbp_actual_cost_payment` captures `key_hash` at `:560`.

routstr/upstream/base.py
```diff
@@ -3575,66 +3579,80 @@ async def forward_get_request(
...
+        try:
+            response = await client.send(
+                client.build_request(
+                    request.method,
+                    url,
+                    headers=headers,
+                    content=request.stream(),
+                    params=self.prepare_params(path, request.query_params),
+                ),
+                stream=True,
```
jeroenubbink: **Unauthenticated GETs now hold a shared pool slot.**

`forward_get_request` is reached only from `proxy.py:439` (Tinfoil attestation) and `proxy.py:571` (the no-auth branch, which 401s anything but GET) — no key, no reservation, no inference.

Streaming these small metadata bodies buys nothing, but it converts a buffered read on a throwaway client into a shared pool slot held for as long as the caller takes to read, on an unauthenticated endpoint with no rate limit in front of it. A slow or aborted reader starves the pool that paid inference shares. Suggest keeping this path buffered.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,81 @@
...
+    loop = asyncio.get_running_loop()
+    if (
+        _upstream_http_client is None
+        or _upstream_http_client.is_closed
+        or _upstream_http_client_loop is not loop
+    ):
+        previous = _upstream_http_client
+        client = httpx.AsyncClient(
+            transport=httpx.AsyncHTTPTransport(
+                limits=httpx.Limits(
```
jeroenubbink: **Was dropping `retries=1` deliberate?**

`httpx.Limits(max_connections=100, max_keepalive_connections=20)` is byte-identical to `httpx.DEFAULT_LIMITS`, so this argument changes nothing. The only behavioural delta in the constructor is the loss of connect-phase retries (`connect_tcp` + TLS handshake, with backoff), which absorbed a DNS blip or a TCP RST while an upstream rolls a deployment. Those now surface as `ConnectError` → 502 or a failover hop.

`base.py:4289`, `:4595` and `tinfoil.py:127` still pass `retries=1`, so the hot path is the only weakened one.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,81 @@
...
+                    max_connections=UPSTREAM_MAX_CONNECTIONS,
+                    max_keepalive_connections=UPSTREAM_MAX_KEEPALIVE_CONNECTIONS,
+                ),
+            ),
+            timeout=httpx.Timeout(
+                connect=30.0,
+                read=UPSTREAM_READ_TIMEOUT_SECONDS,
```
jeroenubbink: **`read` bounds the wait for response headers, not just chunk gaps.**

httpcore's `_receive_response_headers` uses `timeouts.get('read')`. A non-streaming reasoning completion emits nothing until generation finishes, so this caps how long any completion may take: at 300 s it 502s after the upstream has already generated and billed, and `proxy.py` then burns the full budget again on the next candidate.

Suggest splitting the two waits — leave the header wait generous and bound the inter-chunk stall on the stream loop that already runs the heartbeat — making the value operator-configurable rather than a module constant, and not failing over on a `ReadTimeout`, which isn't candidate-specific. `write=30` is the matching bound for large multimodal uploads.

routstr/upstream/gemini_messages.py
```diff
@@ -302,11 +303,10 @@ async def _post_and_stream(
...
-) -> tuple[httpx.AsyncClient, httpx.Response]:
-    """POST to upstream chat-completions and return (client, response) for
-    streaming. Caller is responsible for closing both."""
+) -> httpx.Response:
+    """POST to upstream chat-completions and return a streaming response."""
     url = f"{base_url.rstrip('/')}/chat/completions"
-    client = httpx.AsyncClient(timeout=httpx.Timeout(120.0, read=120.0))
+    client = get_upstream_http_client()
     try:
         request = client.build_request(
```
jeroenubbink, line 357: **This replaces all four timeout fields, including `pool`.**

httpx does `extensions = dict(**extensions, timeout=timeout.as_dict())` with no per-field merge, and `httpx.Timeout(120.0, read=120.0)` expands to `pool: 120.0`. So this path waits 120 s on the shared pool instead of the client's 5 s — holding its reservation and renewing the lease throughout — then surfaces via `except Exception` at `:322` as a 502, so pool exhaustion is misreported as an upstream fault and never reaches the new 503 branch.

Pass only the field you mean to change.

routstr/core/logging.py
```diff
@@ -127,6 +130,103 @@ def _cleanup_old_files(self) -> None:
...
+        self.acquire()
+        try:
+            if self._stopped:
+                return
+
+            deadline = time.monotonic() + self._drain_timeout_seconds
+            with self._queue.all_tasks_done:
+                while self._queue.unfinished_tasks:
+                    remaining = deadline - time.monotonic()
+                    if remaining <= 0:
```
jeroenubbink, line 251: **Both teardown paths drop queued records, and the records are the billing ones.**

`flush()` returns here without ever reaching `self._target.flush()` at `:200`, and `close()` sets `_stopped = True` at `:219` before `_stop_listener()` reports, skipping the flush/close on a join timeout. The `QueueListener` thread is a daemon, so anything still queued is discarded at exit with nothing on stderr — measured **197 of 400 records lost on a normal exit**. Those are the `Payment processed successfully` / `Max cost payment finalized` lines this module declares a parsing contract for.

The join-timeout path is worse than lossy: the listener and its fd are orphaned with `_stopped` already True, so the next `emit()` opens a **second** handler on the same file — two buffered streams appending, each running `doRollover` → `_cleanup_old_files` against files the other holds open.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,81 @@
...
+    if (
+        _upstream_http_client is None
+        or _upstream_http_client.is_closed
+        or _upstream_http_client_loop is not loop
+    ):
+        previous = _upstream_http_client
+        client = httpx.AsyncClient(
+            transport=httpx.AsyncHTTPTransport(
+                limits=httpx.Limits(
+                    max_connections=UPSTREAM_MAX_CONNECTIONS,
```
jeroenubbink: **The cap is inherited, global, and not operator-configurable.**

This is the httpx default, and httpcore applies it to one global connection list with no per-origin bucketing — so 100 slow streams to one provider starve every other provider, and no bulkhead is expressible on a single `Limits`. Worth considering a client per upstream origin, which makes the cap per-provider for free.

Two related points: `PoolTimeout` is raised as `UpstreamError(503)` into the candidate loop, which `continue`s against the same exhausted pool (5×N seconds, reservation held) — it's a local resource condition, not the candidate's fault, so failing fast is better. And `keepalive_expiry` defaults to 5.0 s, so below roughly 1 req/5 s per origin every request still pays a fresh handshake, while inheriting the cap, the 5 s `PoolTimeout`, `read=300` and the lost retries. Neither number is configurable, unlike the DB pool.

routstr/proxy.py
```diff
@@ -610,9 +609,10 @@ async def _proxy(
 
     reservation_snapshot: ReservationSnapshot | None = None
     if is_ehbp or request_body_dict:
-        await pay_for_request(key, max_cost_for_model, session)
-        reservation_snapshot = await get_reservation_snapshot(key, session)
-        # Snapshot validation performs SELECTs after pay_for_request commits.
+        reservation_snapshot = await pay_for_request(
```
jeroenubbink, line 612: **This removes no queries — it removes the validation.**

Returning the snapshot from `pay_for_request` is a good simplification, but both `session.get()` calls in `_validate_reservation_snapshot` are identity-map hits: sessions are `expire_on_commit=False` (`core/db.py:1039`) and `pay_for_request` committed the `ReservationRelease` in this same session, so zero SQL is emitted (measured on SQLAlchemy 2.0.42 + aiosqlite). The deleted comment's premise — *"validation performs SELECTs"* — is what's incorrect; the saving is two dict lookups.

What's actually gone is the key-ownership / billing-key-ownership / `status == 'active'` / `reserved_msats` cross-check at reserve time. An orphan child key makes `get_billing_key` fall back at `auth.py:638`, and that mismatch used to raise **before** the upstream call; it now first raises at settlement, after the tokens are paid for and consumed.

Suggest keeping the validation call.

routstr/core/logging.py
```diff
@@ -127,6 +130,103 @@ def _cleanup_old_files(self) -> None:
...
+        try:
+            listener.start()
+        except Exception:
+            target.close()
+            raise
+
+        self._queue = record_queue
+        self._target = target
+        self._listener = listener
+        self._stopped = False
```
jeroenubbink, line 177: **`_stopped = False` is assigned last, so a failed open wedges the handler permanently.**

If `DailyRotatingFileHandler(...)` at `:153` raises — missing `logs/`, full volume — `_stopped` stays True and every subsequent record retries a file open plus a thread start, with `handleError` writing a traceback to stderr each time (measured: 50 attempts for 50 records). In a container that's the Docker log, so each of the ~8 log lines per request becomes an `open()` syscall and a synchronous multi-line stderr write on the event-loop thread — self-amplifying exactly when the node is already degraded.

Separately, `_open()` never resets `self._closed = False` after `super().close()` set it, so a live handler reports closed.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,81 @@
...
+                pool=UPSTREAM_POOL_TIMEOUT_SECONDS,
+            ),
+        )
+        # AsyncClient's public setter copies into a concrete Cookies jar.
+        # Replace the backing jar so response cookies are never retained.
+        client._cookies = _StatelessCookies()
+        _upstream_http_client = client
+        _upstream_http_client_loop = loop
+        if previous is not None and not previous.is_closed:
+            loop.create_task(previous.aclose())
```
jeroenubbink: **Fire-and-forget close, on a loop that is usually already dead.**

No reference is kept and no exception handler is installed, and the only path here is the loop-mismatch branch — so `previous`'s connections belong to a closed loop. Reproduced: `RuntimeError('Event loop is closed')` from the loop exception handler, then `ResourceWarning: unclosed socket`. httpcore empties `_connections` before the awaits that raise, so the socket is never closed while `previous.is_closed` flips to True — the flag lies. CPython also holds tasks only weakly, so it can be GC'd mid-close.

At minimum, keep a reference and attach a done-callback that logs the failure, so a leaked pool is visible rather than silent.

routstr/upstream/base.py
```diff
@@ -2965,10 +2966,7 @@ async def forward_request(
             },
         )
 
-        client = httpx.AsyncClient(
-            transport=httpx.AsyncHTTPTransport(retries=1),
-            timeout=None,
-        )
+        client = get_upstream_http_client()
```
jeroenubbink: Raised outside the `try:` on the next line, so `RuntimeError("Upstream HTTP client is shutting down")` escapes as a 500 rather than the 503 the guard intends. Same at `:3344` and `:3582`.

routstr/upstream/base.py
```diff
@@ -3153,17 +3150,20 @@ async def forward_request(
...
-                            max_cost_for_model,
-                            background_tasks,
-                            requested_model=original_model_id,
-                            model_obj=model_obj,
-                            reservation_snapshot=reservation_snapshot,
-                            client=client,
-                            request_body=request_body,
-                        )
+                        try:
+                            return await self.handle_streaming_chat_completion(
```
jeroenubbink: `client` is `None` at every call site now — pre-PR it was passed here and at `:3443` to close the per-request client. Suggest deleting the parameter at `:1024`, `:1510` and `:89` along with `_aclose_if_needed(client)`: left in place, `client=get_upstream_http_client()` is the obvious next edit and would close the shared client out from under every other in-flight stream, with mypy unable to object.

routstr/upstream/base.py
```diff
@@ -3575,66 +3579,80 @@ async def forward_get_request(
...
+                    "error": str(exc),
+                    "error_type": type(exc).__name__,
+                    "method": request.method,
+                    "url": url,
+                    "path": path,
+                    "query_params": dict(request.query_params),
+                    "traceback": tb,
+                },
+            )
+            return create_error_response(
```
jeroenubbink, line 3932: The GET path maps only `PoolTimeout`; `ConnectError`, `ReadTimeout` and `NetworkError` all fall through to this generic 500, while `proxy.py:572` fails over on `[502, 429]` only — so an unreachable upstream on a GET never triggers failover, unlike the POST paths.

routstr/upstream/base.py
```diff
@@ -3575,66 +3579,80 @@ async def forward_get_request(
...
+            return StreamingResponse(
+                response.aiter_bytes(),
+                status_code=response.status_code,
+                headers=response_headers,
+                background=background_tasks,
+            )
+        except BaseException as exc:
+            if response is not None:
+                await response.aclose()
+            if isinstance(exc, httpx.PoolTimeout):
```
jeroenubbink: Returns before any logging, and nothing anywhere logs the effective limits, so pool exhaustion is invisible on this path.

routstr/core/main.py
```diff
@@ -247,6 +248,8 @@ async def lifespan(_: FastAPI) -> AsyncGenerator[None, None]:
                 "Error stopping background tasks",
                 extra={"error": str(e), "error_type": type(e).__name__},
             )
+        finally:
+            await close_upstream_http_client()
```
jeroenubbink, line 252: The only unguarded await in an otherwise defensively-wrapped shutdown: an `aclose()` error turns a clean exit into "Application shutdown failed" after the operator was told background tasks stopped cleanly.

routstr/upstream/http_client.py
```diff
@@ -0,0 +1,81 @@
...
+            timeout=httpx.Timeout(
+                connect=30.0,
+                read=UPSTREAM_READ_TIMEOUT_SECONDS,
+                write=30.0,
+                pool=UPSTREAM_POOL_TIMEOUT_SECONDS,
+            ),
+        )
+        # AsyncClient's public setter copies into a concrete Cookies jar.
+        # Replace the backing jar so response cookies are never retained.
+        client._cookies = _StatelessCookies()
```
jeroenubbink: `set_cookie_header` is never called — httpx builds a fresh merged `Cookies` per request — so that override is dead code, and the jar stays empty only because `extract_cookies` is a no-op. Writing the private `client._cookies` is also silently reverted by the public setter.

routstr/auth.py
```diff
@@ -1767,6 +1767,45 @@ async def _accumulate_fee(total_cost_msats: int) -> None:
...
+    started = time.perf_counter()
+    key_log_hash = key.hashed_key[:8] + "..."
+    succeeded = False
+    try:
+        result = await _adjust_payment_for_tokens(
+            key,
+            response_data,
+            session,
+            deducted_max_cost,
+            model_obj,
```
jeroenubbink, line 1830: This "Payment settlement finished" carries a different field set from the `ehbp.py:530` emitter, and the message isn't listed in `logging.py`'s CRITICAL LOG MESSAGES contract.

tests/unit/test_upstream_http_client.py
```diff
@@ -0,0 +1,87 @@
...
+        await close_upstream_http_client()
+
+
+@pytest.mark.asyncio
+async def test_upstream_http_client_has_bounded_pool_wait_and_read_timeout() -> None:
+    client = get_upstream_http_client()
+    try:
+        assert client.timeout.pool == 5.0
+        assert client.timeout.read == 300.0
+        assert getattr(client._transport, "_pool")._max_connections == 100
```
jeroenubbink: This asserts the httpx default: it passes against a bare `AsyncHTTPTransport()`, and with `UPSTREAM_MAX_CONNECTIONS` deleted. It never references the constant it is meant to pin. It also reaches into `client._transport._pool`, against CLAUDE.md's public-API rule.

tests/unit/test_queued_logging.py
```diff
@@ -0,0 +1,117 @@
...
+
+    assert any(reference() is handler for reference in handler_list)
+
+    log_path = next(tmp_path.glob("app_*.log"))
+    assert "after close" in log_path.read_text()
+
+    handler.close()
+
+
+def test_queued_file_handler_contains_reopen_failures(
```
jeroenubbink, line 76: No assertion here. With `raiseExceptions` forced False the only property checked is that `logger.info()` doesn't raise, which the bare `except Exception: self.handleError(record)` guarantees regardless of whether the handler is wedged, dropping records, or duplicating them — and this is the sole test of the disk-failure path the payment log depends on.

tests/unit/test_queued_logging.py
```diff
@@ -0,0 +1,117 @@
...
+def test_queued_file_handler_survives_close_racing_with_emit(tmp_path: Path) -> None:
+    """Closing while another thread logs must not hang or orphan the queue."""
+    handler = QueuedDailyRotatingFileHandler(
+        str(tmp_path / "app.log"), when="midnight", backupCount=1
+    )
+    handler.setFormatter(logging.Formatter("%(message)s"))
+
+    logger = logging.Logger("queued-file-race-test")
+    logger.addHandler(handler)
+
```
jeroenubbink: `spam()` exits only via `done`, which the main thread cannot set while the closer holds the handler lock — so when the deadlock this test names actually occurs, both joins time out, the asserts fail, and `threading._shutdown()` then blocks forever joining the live non-daemon threads. CI reports a hung job rather than a failure. `done.set()`/`handler.close()` are also outside a `finally`, so a raise from the 5 s `flush()` leaks the listener and its fd into every later test.

tests/unit/test_queued_logging.py
```diff
@@ -0,0 +1,117 @@
...
+def test_queued_file_handler_keeps_logging_after_close(tmp_path: Path) -> None:
+    """dictConfig closes live handlers; uvicorn runs one after app import."""
+    handler = QueuedDailyRotatingFileHandler(
+        str(tmp_path / "app.log"), when="midnight", backupCount=1
+    )
+    handler.setFormatter(logging.Formatter("%(message)s"))
+
+    logger = logging.Logger("queued-file-reopen-test")
+    logger.addHandler(handler)
+    logger.info("before close")
```
jeroenubbink, line 56: Asserts on `handler._closed` and `logging._handlerList`, and `auth._adjust_payment_for_tokens` is monkeypatched elsewhere in the suite — CLAUDE.md asks for assertions through the public API.

tests/unit/test_queued_logging.py
```diff
@@ -0,0 +1,117 @@
+import logging
```
jeroenubbink, line 1: No module docstring here or in the two other new test files; the convention in this repo is a module docstring describing what the tests cover. `test_queued_file_handler_contains_reopen_failures` also takes `monkeypatch` unannotated while its siblings annotate.

jeroenubbink (changes requested):
Re-reviewed at `761ff58`. Three of the previous round's blockers are genuinely fixed: `emit()` routes failures to `handleError`, `close()` chains to `super().close()` and re-registers in `_handlerList`, and the shared client has explicit limits, a finite `pool`/`read`, and clears its globals behind a closing flag. The shared-pool direction is right and the latency goal is worth having.

What follows is what those fixes left open or newly introduced. Six are money- or availability-affecting; all are latent, and the last group explains why four of them are invisible to the suite (`make lint` clean, 1307 unit + 456 integration green).

Two are worth calling out because the fix is itself the defect. The `limits=` added in response to blocker 1 is byte-identical to `httpx.DEFAULT_LIMITS`, while the same rewrite silently dropped `retries=1` — the only behavioural change in that constructor is the regression. And the `read=300` added alongside it is what makes the generic-streaming settlement hole reachable for the first time.

Verified on the PR head: the logging deadlock (twice, with faulthandler), 197 of 400 records dropped at a normal exit, 50 open attempts for 50 records after one disk error, the EHBP `MissingGreenlet` after rollback, the cross-loop socket leak, and measured `retries == 0`, `keepalive_expiry == 5.0`, `Limits(100,20) == DEFAULT_LIMITS`, `Timeout(120.0, read=120.0).as_dict()['pool'] == 120.0`.

Two notes that had no line in the diff to hang off:

- **`base.py:4289` and `:4595`** — the X-Cashu paths still build per-request clients, so the browser-PWA path gets none of the pooling and the 100-connection cap doesn't describe the node's real socket count. Intentional?
- **`tests/unit/test_log_secret_redaction.py:43`** — the fixture docstring "configured exactly like the production file handler" is no longer true after this change.

PR #700, opened 2026-08-29
fix(ehbp): raise upstream timeout to 60s and return 504 on timeout

sh1ftred:
## Summary

Fixes the `TimeoutError` reported in [tinfoilsh/confidential-model-router#464](https://github.com/tinfoilsh/confidential-model-router/issues/464), where EHBP (Tinfoil) requests intermittently failed with a bare `500` instead of a meaningful timeout.

Routstr's EHBP forwarding buffers the encrypted response through an h11 client (`tinfoil_trailer.py`) to capture the `X-Tinfoil-Usage-Metrics` trailer, and imposed a hard-coded **30-second** inactivity timeout. Slow time-to-first-token (long prompts / queueing) on models like Kimi-K3 would trip it.

## Changes

- **Bump the default EHBP timeout from 30s → 60s** (`_DEFAULT_TIMEOUT_SECONDS`).
- **Introduce `EhbpTimeoutError`** (subclass of `UpstreamError`, `code=UPSTREAM_TIMEOUT`, `status_code=504`) raised from connect / send / read timeouts in the h11 forwarding client.
- **Bearer auth** now surfaces a proper `504` (via the existing `UpstreamError` handling) instead of a generic `500`.
- **X-Cashu** requests now refund the redeemed amount and return a `504` with the refund token, so a timeout can't eat user funds.

## Tests

- `test_tinfoil_trailer.py`: connect-timeout and read-timeout both raise `EhbpTimeoutError`; exception metadata (`504` / `UPSTREAM_TIMEOUT`).
- `test_ehbp_timeout.py`: X-Cashu timeout refunds the full amount and returns `504` with `X-Cashu` set.

`ruff check`, `ruff format --check`, and `mypy` all pass on the changed files.

### DIFF
--- a/routstr/core/exceptions.py
+++ b/routstr/core/exceptions.py
@@ -34,6 +34,22 @@ class UpstreamError(Exception):
         super().__init__(message)
 
 
+class EhbpTimeoutError(UpstreamError):
+    """Raised when an EHBP upstream times out waiting for a response.
+
+    Distinct from a generic :class:`UpstreamError` so callers can map the
+    failure to a ``504 Gateway Timeout`` with a stable ``UPSTREAM_TIMEOUT``
+    code instead of a misleading ``500`` internal server error.
+    """
+
+    def __init__(self, message: str, status_code: int = 504):
+        super().__init__(
+            message,
+            status_code=status_code,
+            code="UPSTREAM_TIMEOUT",
+        )
+
+
 async def http_exception_handler(request: Request, exc: Exception) -> JSONResponse:
     """Handle HTTP exceptions and include request ID in response."""
     request_id = getattr(request.state, "request_id", "unknown")
--- a/routstr/upstream/ehbp.py
+++ b/routstr/upstream/ehbp.py
@@ -32,7 +32,7 @@ from ..core.db import (
 from ..core.db import (
     store_cashu_transaction_with_retry as store_cashu_transaction,
 )
-from ..core.exceptions import UpstreamError
+from ..core.exceptions import EhbpTimeoutError, UpstreamError
 from ..core.settings import settings
 from ..payment.cost_calculation import (
     CostData,
@@ -1041,6 +1041,46 @@ async def forward_ehbp_x_cashu_request(
             )
         except Exception:
             raise
+
+    except EhbpTimeoutError as e:
+        logger.warning(
+            "EHBP X-Cashu upstream timed out",
+            extra={
+                "error": str(e),
+                "path": path,
+                "method": request.method,
+                "redeemed": redeemed,
+            },
+        )
+
+        if redeemed and amount > 0:
+            try:
+                refund_token = await send_cashu_refund(amount, unit, mint, request_id)
+                error_response = create_error_response(
+                    "upstream_timeout",
+                    str(e),
+                    504,
+                    request=request,
+                    code="UPSTREAM_TIMEOUT",
+                )
+                error_response.headers["X-Cashu"] = refund_token
+                return error_response
+            except Exception as refund_error:
+                logger.error(
+                    "Failed to refund EHBP X-Cashu token after timeout",
+                    extra={
+                        "error": str(refund_error),
+                        "original_error": str(e),
+                    },
+                )
+
+        return create_error_response(
+            "upstream_timeout",
+            str(e),
+            504,
+            request=request,
+            code="UPSTREAM_TIMEOUT",
+        )
 
     except Exception as e:
         error_message = str(e)
--- a/routstr/upstream/tinfoil_trailer.py
+++ b/routstr/upstream/tinfoil_trailer.py
@@ -20,11 +20,12 @@ from urllib.parse import urlsplit
 import h11
 
 from ..core import get_logger
+from ..core.exceptions import EhbpTimeoutError
 
 logger = get_logger(__name__)
 
 _READ_BUFSIZE = 65536
-_DEFAULT_TIMEOUT_SECONDS = 30.0
+_DEFAULT_TIMEOUT_SECONDS = 60.0
 _DEFAULT_CLOSE_TIMEOUT_SECONDS = 1.0
 _DEFAULT_MAX_RESPONSE_BYTES = 25 * 1024 * 1024
 _HOP_BY_HOP_HEADERS = {
@@ -100,10 +101,15 @@ async def forward_with_trailer(
     headers = _strip_hop_by_hop_headers(headers)
 
     ssl_ctx = ssl.create_default_context()
-    reader, writer = await asyncio.wait_for(
-        asyncio.open_connection(host, port, ssl=ssl_ctx),
-        timeout=timeout_seconds,
-    )
+    try:
+        reader, writer = await asyncio.wait_for(
+            asyncio.open_connection(host, port, ssl=ssl_ctx),
+            timeout=timeout_seconds,
+        )
+    except asyncio.TimeoutError as exc:
+        raise EhbpTimeoutError(
+            f"EHBP upstream {host} timed out after {timeout_seconds:g}s connecting"
+        ) from exc
 
     try:
         # Build HTTP/1.1 request
@@ -126,7 +132,13 @@ async def forward_with_trailer(
             request_data += body
 
         writer.write(request_data)
-        await asyncio.wait_for(writer.drain(), timeout=timeout_seconds)
+        try:
+            await asyncio.wait_for(writer.drain(), timeout=timeout_seconds)
+        except asyncio.TimeoutError as exc:
+            raise EhbpTimeoutError(
+                f"EHBP upstream {host} timed out after "
+                f"{timeout_seconds:g}s sending request"
+            ) from exc
 
         # Parse response with h11
         conn = h11.Connection(h11.CLIENT)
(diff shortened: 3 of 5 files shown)


routstr/upstream/ehbp.py
```diff
@@ -1042,6 +1042,46 @@ async def _stream_body_xcashu() -> AsyncIterator[bytes]:
...
+                "redeemed": redeemed,
+            },
+        )
+
+        if redeemed and amount > 0:
+            try:
+                refund_token = await send_cashu_refund(amount, unit, mint, request_id)
+                error_response = create_error_response(
+                    "upstream_timeout",
+                    str(e),
```
jeroenubbink, line 1061: **Item 1 — the upstream hostname now reaches the client body.**

`str(e)` here (and again at line 1079, and on the bearer path via `create_upstream_error_response`) carries the trailer client's message, which embeds `{host}`:

```
{"error": {"message": "EHBP upstream inference.private-infra.internal timed out after 60s connecting",
           "type": "upstream_timeout", "code": "UPSTREAM_TIMEOUT"}}
```

`create_error_response` does redact, but only `redact_org_ids`, which matches `org-<base62>` and nothing hostname-shaped. Both immediate siblings avoid this — line 941 answers "Error forwarding EHBP request to upstream" and line 1102 "EHBP request failed after token redemption" — and `main`'s generic 500 didn't expose it either, so this is new client-visible detail. Tinfoil's host is public, but the same code path serves any EHBP provider.

Suggest a fixed client message and `str(e)` kept in the log `extra` (it's already there at line 1049).

tests/unit/test_ehbp_timeout.py
```diff
@@ -0,0 +1,81 @@
...
+    request.state.request_id = "req-123"
+    request.method = "POST"
+    request.query_params = {}
+    request.headers = {}
+    request.body = AsyncMock(return_value=b"opaque")
+    return request
+
+
+@pytest.mark.asyncio
+async def test_x_cashu_timeout_refunds_and_returns_504(
```
jeroenubbink: **Item 2 — the bearer path has no test.**

The three new tests cover the trailer client and `forward_ehbp_x_cashu_request`. The bare-500 that motivates the PR is on the *bearer* path, and what turns the timeout into a 504 there is one line: `ehbp.py:860` `except UpstreamError: raise`, sitting above the `except Exception` that flattens everything to 500.

`test_ehbp_timeout_error_metadata` does pin the `UpstreamError` subclassing, so that half is covered. The handler isn't — deleting those two lines from `forward_ehbp_request` restores the exact reported behaviour and all 7 new tests still pass, as does the whole ehbp unit selection (37 passed).

A test asserting `forward_ehbp_request` propagates 504 / `UPSTREAM_TIMEOUT` when `forward_with_trailer` times out would close it.

routstr/upstream/ehbp.py
```diff
@@ -1042,6 +1042,46 @@ async def _stream_body_xcashu() -> AsyncIterator[bytes]:
...
+                "method": request.method,
+                "redeemed": redeemed,
+            },
+        )
+
+        if redeemed and amount > 0:
+            try:
+                refund_token = await send_cashu_refund(amount, unit, mint, request_id)
+                error_response = create_error_response(
+                    "upstream_timeout",
```
jeroenubbink, line 1060: The same failure answers two different `error.type` values: `upstream_timeout` here, and `upstream_error` on the bearer path, because `create_upstream_error_response` hardcodes the type while forwarding `status_code` and `code`. Same 504, same `UPSTREAM_TIMEOUT`, different `type` depending on how the caller authenticated. Worth picking one.

routstr/upstream/tinfoil_trailer.py
```diff
@@ -20,11 +20,12 @@
 import h11
 
 from ..core import get_logger
+from ..core.exceptions import EhbpTimeoutError
 
 logger = get_logger(__name__)
 
 _READ_BUFSIZE = 65536
-_DEFAULT_TIMEOUT_SECONDS = 30.0
+_DEFAULT_TIMEOUT_SECONDS = 60.0
```
jeroenubbink, line 28: Two things about 60 as a constant.

It's the only timeout in the node that isn't a setting — `settings.py` has `mint_operation_timeout_seconds`, `stale_reservation_timeout_seconds`, `refund_sweep_claim_timeout_seconds`, `database_pool_timeout`, `database_busy_timeout`. Neither `forward_with_trailer` call site passes `timeout_seconds`, so this default is the whole policy, and the next model that exceeds it needs another release.

It also changes worst-case hold time. EHBP requests reserve the maximum: `proxy.py` passes `request_body_dict = {}` because the body is sealed, so `calculate_discounted_max_cost` sees no prompt tokens and no `max_tokens` and applies no discount. `pay_for_request` is taken before the candidate loop, and a timeout fails over, so a provider-wide stall now pins a full-ceiling reservation for up to N candidates x 60s where it was N x 30s.

routstr/upstream/tinfoil_trailer.py
```diff
@@ -140,10 +152,16 @@ async def forward_with_trailer(
             event = conn.next_event()
 
             if event is h11.NEED_DATA:
-                data = await asyncio.wait_for(
-                    reader.read(_READ_BUFSIZE),
-                    timeout=timeout_seconds,
-                )
+                try:
+                    data = await asyncio.wait_for(
+                        reader.read(_READ_BUFSIZE),
```
jeroenubbink, line 157: The `wait_for` wraps each individual `reader.read()`, so this is an inactivity timeout, not a deadline — an upstream that sends one byte every 59s keeps the connection, the worker and the reservation indefinitely. The actual bound on a slow-but-alive response is `_DEFAULT_MAX_RESPONSE_BYTES`.

That's a reasonable shape for streaming, but it means raising this constant doesn't bound the case it looks like it bounds. A separate connect budget and inactivity window would say that explicitly.

routstr/upstream/ehbp.py
```diff
@@ -1042,6 +1042,46 @@ async def _stream_body_xcashu() -> AsyncIterator[bytes]:
...
+                    str(e),
+                    504,
+                    request=request,
+                    code="UPSTREAM_TIMEOUT",
+                )
+                error_response.headers["X-Cashu"] = refund_token
+                return error_response
+            except Exception as refund_error:
```
jeroenubbink, line 1068: If `send_cashu_refund` raises, this logs and falls through to line 1077, so the client gets a 504 with no `X-Cashu` header and nothing distinguishing it from a timeout that refunded cleanly — the redeemed amount is unrecoverable client-side. Same hole exists in the twin at line 1113, so this isn't new, but it's now a third site. A distinct code on that branch would at least make it greppable.

jeroenubbink (commented):
Verified the fix end-to-end rather than from the diff: on `main` a timing-out `forward_with_trailer` surfaces as `500` / `code=null` / "An unexpected server error occurred"; on this branch the bearer path surfaces `504` / `UPSTREAM_TIMEOUT`, via `ehbp.py:860` `except UpstreamError: raise` into `proxy.py:833`. The 504 is atomic — `forward_with_trailer` buffers the whole response, so a timeout can only fire before any client byte is sent.

On the X-Cashu path the timeout already refunded and answered 502 before this change, so the new handler moves no money; it changes the status and the message. Note it also returns rather than raises, so an X-Cashu client gets no failover to the next candidate while a bearer client does — consistent with every other EHBP X-Cashu error, just worth knowing.

Items 1 and 2 below are the two I'd want resolved before merge; the rest are follow-up material.

routstr/core/exceptions.py
```diff
@@ -34,6 +34,22 @@ def __init__(
...
 
+class EhbpTimeoutError(UpstreamError):
+    """Raised when an EHBP upstream times out waiting for a response.
+
+    Distinct from a generic :class:`UpstreamError` so callers can map the
+    failure to a ``504 Gateway Timeout`` with a stable ``UPSTREAM_TIMEOUT``
+    code instead of a misleading ``500`` internal server error.
+    """
+
+    def __init__(self, message: str, status_code: int = 504):
```
jeroenubbink: `status_code` is a parameter no caller passes, and the signature drops `details`, which `UpstreamError` accepts and `create_upstream_error_response` reads via `getattr(error, "details", None)` — so a timeout can never carry details even though the response builder would forward them. `def __init__(self, message: str, details: dict | None = None)` with a fixed 504 would match how it's used.

tests/unit/test_ehbp_timeout.py
```diff
@@ -0,0 +1,81 @@
...
+    monkeypatch.setattr(
+        ehbp_module,
+        "recieve_token",
+        AsyncMock(return_value=(1000, "msat", None)),
+    )
+    monkeypatch.setattr(
+        ehbp_module, "store_cashu_transaction", AsyncMock(return_value=None)
+    )
+    send_cashu_refund_mock = AsyncMock(return_value="refund-token")
+    monkeypatch.setattr(
```
jeroenubbink: `ruff format` wants this call on one line, and the assertion at line 79 likewise — `ruff format --check` reports this file as needing reformatting at `0662dc5`, where it was clean at `b15ab59`. CI won't catch it: `test.yml` runs `ruff check .` and `mypy .`, never `ruff format --check`.

PR #701, opened 2026-08-30
docs: propose the direction for analytics v2

ashen0x:
Analytics are scraped out of log files, republished whenever the payload changes on a 15 minute loop, and have grown past what most relays will store. A page reading them cannot tell a quiet day from a day nobody reported, and cannot reliably tell two nodes apart.

statpipe

The document proposes what replaces that, and makes four decisions rather than leaving them open:

1. Revenue stays public, as daily settled totals per node and per model on the closed week.
2. Provider identity is the signed `(pubkey, provider_d)` coordinate, not the pubkey alone.
3. A completed request counts once, after a successful terminal state is settled and committed.
4. Public model identity is `canonical_slug or id` from whichever candidate actually served.

Reasoning for each is in the document, including the arguments against. Say so if you disagree with any and I will change it before the format is frozen. Three of the four cannot be undone once a node publishes. Revenue is the exception and it only goes one way: it can be added later, never withdrawn, so if you want it node-local this is the moment.

Rollout: analytics v2 is default-off behind its own setting. While v2 is off, v1 publishes unchanged. Enabling v2 switches publishing from v1 to v2, so they never publish at the same time.
24h view: v2 uses the last complete UTC day and does not publish partial-day data, preserving daily privacy resolution.

### DIFF
--- a/docs/protocol/analytics-v2.md
+++ b/docs/protocol/analytics-v2.md
@@ -1,0 +1,171 @@
+# Analytics v2, direction
+
+**This is a direction document, not a specification.** The wire format is written and
+follows in a second PR, because each decision below changes what it has to carry.
+
+In short: v2 records durable terminal outcomes in the node, publishes daily facts inside
+weekly events, waits for relay acceptance, and leaves view windows to consumers.
+
+Nothing here is implemented. v1 stays live until a node upgrades.
+
+Building a cost or benchmark page? Per-request cost already works today and needs none of
+this, see decision 1. What v2 adds is public historical aggregates.
+
+## What is wrong with v1
+
+Each of these is observed in shipped code or measured on live relays on 2026-08-29.
+
+**Events exceed the limits of relays the node already targets.** A production node
+publishes a 241,929 byte frame, which exceeds the 131,072 byte `max_message_length` that
+both `nos.lol` and `relay.routstr.com` advertise. On 2026-08-29 it was retrievable from
+`relay.damus.io`, which permits 1,000,000, and a later oversized revision was retrievable
+from `relay.ditto.pub`, which permits 4,000,000. Delivery therefore depends on whichever
+large-limit relays happen to hold it. The bulk is the `windows` object at 202,055 bytes, of which the
+`3m` window alone is 80,079. Nothing in the publisher checks size before sending
+(`routstr/nostr/analytics.py:376-393`).
+
+**Public resolution is hourly, and the publish times are themselves a signal.**
+`model_usage_mix` is copied verbatim into the event
+(`routstr/nostr/analytics.py:219`) and the 24h window is defined at one hour intervals
+(`routstr/nostr/analytics.py:35-43`), so it is an hour by hour series. The
+publisher wakes every 15 minutes and publishes whenever the semantic payload changed
+(`routstr/nostr/analytics.py:313-318`, `:376-379`), so the timing of events reveals when
+public aggregates moved.
+
+**A snapshot cannot say what it covers.** v1 emits rolling window metadata but no
+coverage-start or covered-through boundary (`routstr/nostr/analytics.py:222-229`), and it
+drops zero buckets rather than emitting them
+(`routstr/core/usage_analytics_store.py:1207-1212`). A quiet interval and an unreported
+interval are indistinguishable.
+
+**Usage is inferred from log files.** The store tails `app_*.log` and tracks byte offsets
+(`routstr/core/usage_analytics_store.py:334`, `:453`), and reads the model name off a log
+entry (`:523`). A 2026-08-29 source search found no use of `canonical_slug` anywhere in
+the analytics publisher or store, so rows cannot be joined to the model catalog identity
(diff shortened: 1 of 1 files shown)


jeroenubbink:
Read the doc and the sketch. The direction is right, and the split between the node publishing facts and the consumer owning policy is the part I'd defend hardest. Six notes, checked against `main` at `1715ff3`. Two of them change what the format has to carry.

**1. Publish gross settled, and do not carry `net_revenue` forward.** v1's event exposes both: the per-model rows and the time buckets are gross (`routstr/nostr/analytics.py:163-171`), while the summary adds `net_revenue_msats` / `net_revenue_sats` (`:194-196`). That net is not a net of anything. `revenue_msats` accumulates settled charges on *successful* requests (`routstr/core/log_manager.py:649-650`); `refunds_msats` accumulates the reverted *max-cost reservation* of *failed* ones (`:483-487`); `:501` subtracts the second from the first. Disjoint populations — a failed request never contributed revenue, so what is subtracted is a hold that was never income, and the hold is worst-case sized, so the figure can go negative on a node having a bad day with long-context models. The stats page already plots the gross bucket fields, so gross daily totals are the continuous choice and dropping the net field costs nothing.

**2. Presence needs three states, and "missing" is rarer than it looks.** When an upstream omits its usage trailer, the node does not record an absence — it fabricates a count. `MissingUsageEstimator.billing_data` returns measured usage when present and a locally computed estimate otherwise (`routstr/upstream/count_tokens.py:162-182`), and the non-streaming path substitutes the same way (`routstr/upstream/base.py:1405-1410`). The estimate is stamped `"estimated": true` (`count_tokens.py:209`, `:223`), that flag reaches the client, and it is dropped before billing: `NormalizedUsage` (`routstr/payment/usage.py`) is four integers with no provenance. So at settlement today, an upstream-reported count and a count we computed ourselves are indistinguishable, and a two-state present/absent flag would record the fabricated one as present. Suggest observed / zero-filled / estimated per bucket, captured in `adjust_payment_for_tokens` (`routstr/auth.py:1314`) — the debit and its commit are already in that function (`:1421-1431`), so the ledger write has a transaction to join.

**3. One live path settles at exactly zero, under the model name `"unknown"`.** Generic opaque streams have no request/response token seam, so the node deliberately finalizes with `{"model": "unknown", "usage": None}` (`routstr/upstream/base.py:1955-1975`), `normalize_usage` returns `None`, and `calculate_cost` returns `_empty_cost` (`routstr/payment/cost_calculation.py:109-126`) — the reservation is released and the request is served free. Under decision 3 it counts, so it lands in the daily row with zero revenue and no joinable model identity under decision 4. Worth deciding explicitly whether those requests are counted, excluded, or counted in their own bucket, rather than inheriting the answer from whichever code path writes the ledger.

**4. Per-model revenue mixes real prices with one the node invented.** When a model carries no `input_cache_read`, cache-read tokens bill at the full prompt rate — configured path (`routstr/payment/cost_calculation.py:404-405`) and litellm path (`:437-445`). Measured against the bundled cost map on 2026-09-01: of 2580 chat models, 873 (34%) carry a cache-read rate and 280 (11%) a cache-write rate. Published per-model daily revenue is explicitly meant for cross-node comparison, and this makes two nodes serving identical traffic report different revenue depending on whether the operator happened to configure a rate. Either the row carries whether the price was inferred, or the doc states that revenue is not comparable across nodes at model granularity. I have a change in flight that surfaces exactly this per model, so the flag is cheap to provide if you want it in the format.

**5. `canonical_slug or id` cannot be joined across nodes.** `ModelRow.canonical_slug` is nullable (`routstr/core/db.py:446`) and generic upstreams set it to `None` unconditionally (`routstr/upstream/generic.py:214`). So the same model published by a node fronting a generic OpenAI-compatible endpoint and by a node fronting OpenRouter carries two different public identities, and the consumer counts two models. The fallback is a property of the node's upstream, not of the model. Carrying both columns keeps the join possible.

**6. The consumer cannot assume a requested limit was honoured.** Measured on 2026-09-01: `relay.routstr.com` and `nos.lol` advertise `max_limit: 500` in NIP-11 (and `max_message_length: 131072`, matching your size finding), while `relay.redsh1ft.com` advertises no limitation block at all and returns exactly 100 events for both `-l 300` and `-l 500`, with nothing marking the truncation. At 52 weekly addresses per node per year, a consumer reading a year across ten nodes is already past 500 on the relays that do advertise, and past 100 silently on one that does not — and a truncated read is indistinguishable from missing coverage, which is the failure v2 exists to abolish. Worth specifying that the consumer slices by `since`/`until` rather than raising a limit.

One thing in the other direction: the ledger has a node-local consumer waiting. We are building a per-upstream diagnostics report, and the one row we could not build — how often real traffic got estimated usage rather than reported usage — is parked precisely because it needs this stack. That argues for the ledger being queryable on the node, not only a source for publishing.

On decision 1 itself: keeping revenue public looks right to me for the supply-side signal, with the caveat in item 1 about which number it is.

PR #702, opened 2026-09-01
test(integration): assert admin price edits reach the served catalogue

jeroenubbink:
## Why

The served `/v1/models` catalogue is assumed to track the database, but nothing
asserts it. The closest existing coverage puts its model assertions inside
`for model in data["data"]` and checks only field presence and types — it passes
on an empty catalogue, and its pricing check is
`isinstance(model[field], (dict, type(None)))`, true of any value at all.

## What

Three integration tests that seed a provider, create a model through the admin
API, and assert: a price edit changes the served price, the served price is
fee-adjusted while the admin read-back is raw, and a disabled model drops out of
`/v1/models` while its row stays in the DB.

Each test reaches the served model by id rather than iterating the catalogue, so
an empty catalogue fails them rather than skipping them.

## Tested

`make test-integration` green; lint, format and type-check clean. Every
assertion is on a concrete number, not a type.

### DIFF
--- a/tests/integration/test_model_price_propagation.py
+++ b/tests/integration/test_model_price_propagation.py
@@ -1,0 +1,221 @@
+"""Cover that an admin price write reaches the served catalogue (GET /v1/models).
+
+A price edit changes the served price; the served price is fee-adjusted while the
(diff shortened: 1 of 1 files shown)


9qeklajc:
LGTM

PR #703, opened 2026-09-01
refactor(models): extract the deterministic USD builder from _row_to_model

jeroenubbink:
## Why

Comparing a configured price against the served price requires calling the same
code the serve path uses — minus the sats conversion, which depends on a live
BTC feed and would report exchange-rate noise as pricing drift.

## What

Split `_row_to_model` at that seam. Everything deterministic (request floor,
litellm cache backfill, provider-fee multiply, max-cost recompute) moves to
`_build_model_from_row`; `_row_to_model` becomes that call plus the sats
conversion. Pure mechanical move — no behaviour change, no signature change, the
sats-conversion swallow untouched.

## Tested

Characterization tests over the three public surfaces that reach this code —
`GET /v1/models`, the admin single-model read-back, and the admin provider model
listing — pass unchanged before and after the extraction. They cover the plain
model, litellm backfill, cache-rate preservation, the request floor, the fee
flag against recomputed max costs, survival of a failed sats conversion, the
full serialised dict field for field, and agreement between the two admin views.

Removing a single field from the serialised model fails the full-dict test and
the listing-agreement test while every value-level test still passes — which is
why both are there. Unit + integration suites green; lint, format and type-check
clean.

### DIFF
--- a/routstr/payment/models.py
+++ b/routstr/payment/models.py
@@ -241,9 +241,10 @@ async def async_fetch_openrouter_models(source_filter: str | None = None) -> lis
         return []
 
 
-def _row_to_model(
+def _build_model_from_row(
     row: ModelRow, apply_provider_fee: bool = False, provider_fee: float = 1.01
 ) -> Model:
+    """The deterministic USD view of a stored model row, before the sats conversion."""
     architecture = json.loads(row.architecture)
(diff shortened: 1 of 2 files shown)


9qeklajc:
LGTM

PR #704, opened 2026-09-01
feat(payment): record whether each cache rate was configured or inferred

jeroenubbink:
## Why

When a model has no `input_cache_read` or `input_cache_write`, billing falls back
to the full prompt rate for those tokens. The rate that results is
indistinguishable from a real one — both arrive as a float — so nothing
downstream can tell a rate the node was given from one it invented.

The settlement path already logs `Applied model-specific pricing` with a
`pricing_source` field on every request. That label covers the prompt and
completion rates only, so it reads `configured` on a request whose cache tokens
were priced by the fallback. This PR makes that record say what it knows.

**The fallback itself is unchanged and no charged amount moves.** That is
deliberate: there is no better number to fall back to.

- No upstream API publishes cache pricing. `/v1/models` has no caching field
  anywhere in the OpenAI-compatible shape, and Anthropic's `count_tokens` never
  consults the cache. Cache rates live in human-facing pricing pages.
- Both catalogue sources are already consulted before a rate can be called
  inferred — the provider feed, then litellm via `backfill_cache_pricing`.
  litellm carries a cache-read rate for 873 of 2580 chat models; for the rest
  there is no source left to try, and those are disproportionately the custom
  and self-hosted upstreams a node operator configures by hand.
- The only way to observe a real cache rate is to cause a cache hit and read the
  cost back — which requires the upstream to report a USD cost for the request.
  When it does, that reported figure is what we bill and the cache rate only
  moves the breakdown, not the total. So the measurement is available exactly
  where the rate does not matter, and unavailable exactly where it does.

That leaves one real fix, and it already ships: the operator sets the rate.
`input_cache_read` is in `BILLABLE_PRICING_FIELDS` and accepted by the admin
model PATCH. What was missing is any signal that it needs setting.

## What

Two commits.

1. `_get_pricing_rates` returned four bare floats, unpacked positionally at two
   call sites. Two of the four are cache rates that routinely fall back to the
   prompt rate, so a transposed pair does not crash — it bills a
   plausible-looking wrong amount. It now returns a `PricingRates` named tuple.
   No behaviour change.
2. `PricingRates` carries `cache_read_inferred` / `cache_write_inferred`, and
   the existing pricing record gains `cache_read_source` and
   `cache_write_source` — each `configured`, `litellm` or `inferred`.

No new log lines and no new severity: one record that already fired on every
settlement gains two fields. The two rates are tracked separately, because a
catalogue that knows the read rate routinely omits the write one.

`fixed_pricing` short-circuits before any rate is resolved, so it emits no
pricing record at all — nothing is sourced and nothing is claimed.

## Tested

Seven unit tests, through `calculate_cost`: both rates missing, both configured,
one of each, no cache tokens in the request, the litellm branch labelling its own
rates, the `fixed_pricing` exemption, and billing invariance pinning the exact
msats breakdown rather than that it is positive.

Each was checked against a sabotaged implementation: dropping the two fields
fails five of them, labelling everything with the catalogue source fails four,
and letting the read flag stand in for the write flag fails the independence
test specifically.

Both commits are independently green — lint, type-check and the 1301/1308 unit
tests pass at each — so the branch stays bisectable when merged unsquashed.

https://claude.ai/code/session_01Bhvy9xFm54E2i5yMevUpHG

### DIFF
--- a/routstr/payment/cost_calculation.py
+++ b/routstr/payment/cost_calculation.py
@@ -1,5 +1,5 @@
 import math
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, NamedTuple
 
 from pydantic.v1 import BaseModel
 
@@ -46,6 +46,20 @@ class MaxCostData(CostData):
 class CostDataError(BaseModel):
     message: str
     code: str
+
+
+class PricingRates(NamedTuple):
+    """Per-1k-token rates for the four billable token categories.
+
+    The four arrived as a bare tuple, unpacked positionally at two call sites.
+    Two of them are cache rates that routinely fall back to the prompt rate,
+    which makes a transposition read as a plausible bill rather than a crash.
+    """
+
+    input_1k: float
+    output_1k: float
+    cache_read_1k: float
+    cache_write_1k: float
 
 
 def _empty_cost(cls: type[CostData] = CostData) -> CostData:
@@ -187,7 +201,7 @@ async def calculate_cost(
             output_usd = _coerce_usd(cost_details.get("output_cost")) or _coerce_usd(
                 cost_details.get("upstream_inference_completions_cost")
             )
-            cache_pricing_rates: tuple[float, float, float, float] | None = None
+            cache_pricing_rates: PricingRates | None = None
             if cache_read_tokens > 0 or cache_creation_tokens > 0:
                 try:
                     cache_pricing_rates = _get_pricing_rates(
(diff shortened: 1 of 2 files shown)


9qeklajc:
not sure about this, it just add logs that we would never evaluate .. if we do not get the cache rate, a warning wouldn't really help, it feels verbose, I think it's better to try to fallback and handle it instead of logging..

jeroenubbink:
> not sure about this, it just add logs that we would never evaluate .. if we do not get the cache rate, a warning wouldn't really help, it feels verbose, I think it's better to try to fallback and handle it instead of logging..

Instead of introducing a new logline i added the cache_read_source and cache_write_source to an already existing line so this is just additional information. It is good to surface this though because cache reads (deepseek) get overcharged while cache writes (anthropic) get undercharged. 

There is not really a generic fallback available to fix this. No provider publishes cache pricing in any API and litellm only has a cache-read rate for about a third of chat models. The only way to measure a real cache rate is to make an upstream report a USD cost for a cache hit, but when an upstream reports a cost we bill that number and the cache rate no longer moves the total. So it's measurable exactly where it doesn't matter.

The operator can already fix it per model: `input_cache_read` is in `BILLABLE_PRICING_FIELDS` and accepted by the admin model override endpoint. So the fix exists, nothing was telling anyone to apply it.

9qeklajc:
hav to get my head around this, not sure if this is the right way

9qeklajc:
re-open this when we have a use for the inferred fields

PR #705, opened 2026-09-01
feat(payment): keep the upstream reported USD cost beside the billed one

jeroenubbink:
## Why

The provider fee is multiplied into the same local that holds the upstream's
reported USD cost, destroying it in place. `total_usd` is post-fee and nothing
keeps the pre-fee figure, so a later cost check cannot compare "what the
upstream said" against "what we charged". Dividing the total back out does not
recover it: on a token-priced request the total is the node's own arithmetic,
so the division compares that number to itself.

## What

Add `upstream_usd` to `CostData`, captured before the fee multiply, with
`Field(default=0.0, exclude=True)` so it never reaches a client — published
beside `total_usd` it would spell out the node's margin as a ratio. It stays on
the object and does not survive `.dict()`, which is what the outbound cost
metadata and response headers are built from.

Token-priced requests leave it at 0.0, which doubles as the discriminator for
whether an upstream reported a cost at all. No persistence, no migration, no
charged-amount change.

## Tested

Unit tests assert the pre-fee figure survives the multiply, the token path
leaves it zero, it never serialises through `.dict()` or `.json()`, and the
billed total is reproducible from it through fee, rate and rounding. Unit +
integration suites green; lint, format and type-check clean.

### DIFF
--- a/routstr/payment/cost_calculation.py
+++ b/routstr/payment/cost_calculation.py
@@ -1,7 +1,7 @@
 import math
 from typing import TYPE_CHECKING
 
-from pydantic.v1 import BaseModel
+from pydantic.v1 import BaseModel, Field
 
 from ..core import get_logger
 from ..core.settings import settings
@@ -37,6 +37,7 @@ class CostData(BaseModel):
     cache_creation_msats: int = 0
     # Actual debit after finalization; None means settlement has not run yet.
     charged_msats: int | None = None
+    upstream_usd: float = Field(default=0.0, exclude=True)
 
 
 class MaxCostData(CostData):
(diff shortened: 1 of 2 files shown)


routstr/payment/cost_calculation.py
```diff
@@ -37,6 +37,10 @@ class CostData(BaseModel):
     cache_creation_msats: int = 0
     # Actual debit after finalization; None means settlement has not run yet.
     charged_msats: int | None = None
+    # The upstream's own reported USD cost, before the provider fee is applied.
```
9qeklajc: the comment is not really needed

jeroenubbink: Removed. Amended into the single commit rather than adding a follow-up, since upstream merges without squashing — `c2c2c97` → https://github.com/Routstr/routstr-core/pull/705/commits/b8c5a6e0f4ab1041adef2ea701d5d9589ead9d5f

9qeklajc:
LGTM

PR #708, opened 2026-09-03
fix(auth): clamp overrun charge to a delegated child's balance limit

jeroenubbink:
## The bug

A delegated child key's `balance_limit` is enforced when a request **reserves** funds, but not when it **settles**. If the response costs more than was reserved, the overrun path charges whatever the parent's balance allows and never re-checks the child's limit — so a child key can spend past its limit by overrunning, repeatedly.

## The fix

In the overrun branch of `adjust_payment_for_tokens`, re-read the child row and clamp the charge to its remaining limit (`balance_limit - total_spent`). The same condition is passed down as a SQL guard on the child `UPDATE`, so a concurrent writer cannot slip in between the read and the write; a failed guard falls into the existing retry loop.

## Concessions

- **The node absorbs the overrun above the limit.** The clamp reduces the charge on the parent row too, so the excess is not billed to anyone. This matches the balance clamp directly above it, which already caps at what the parent can pay.
- **The guard is opt-in, not unconditional.** Putting it inside `_charge_reservation_rows` for every caller is simpler but wrong: if an admin lowers a child's limit while a request is in flight, an ordinary in-authorization settlement would silently fail the guard, exhaust the retry loop, and release the reservation without charging anything. `test_a_midflight_limit_drop_does_not_free_the_settlement` pins that case.
- `with_for_update()` is a no-op on SQLite. Correctness comes from the SQL guard plus the retry loop, not from row locking — the comment says so rather than implying a lock that isn't there.

### DIFF
--- a/routstr/auth.py
+++ b/routstr/auth.py
@@ -1255,6 +1255,7 @@ async def _charge_reservation_rows(
     reserved_msats: int,
     charge_msats: int,
     extra_billing_guards: tuple = (),
+    extra_child_guards: tuple = (),
 ) -> bool:
     """Release the reserved amount and record the charge on the billing row
     (and the child row when different) inside the caller's transaction.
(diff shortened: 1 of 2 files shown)


9qeklajc:
thnx but child key feature was removed completely

PR #714, opened 2026-09-05
fix(ehbp): pass through enclave key-config 422 as problem+json

sh1ftred:
## Problem

Tinfoil rotated their enclave HPKE keys (~Sep 4). Long-running SDK clients (e.g. the routstrd daemon, up since Aug 29) kept sealing requests to the old key. The enclave returned `422 application/problem+json` with `type=urn:ietf:params:ehbp:error:key-config`, but the proxy wrapped it into its own `application/json` error envelope via `create_upstream_error_response()`. This destroyed the content type and body shape that EHBP clients use to detect key-config mismatch and trigger re-attestation. Result: **every request with a stale key failed permanently** — the recovery loop never fired.

```
enclave:  422 application/problem+json  {"type":"urn:...key-config", ...}
             ↓  proxy wraps
client:   422 application/json          {"error":{"message":"EHBP upstream..."}}
             ↓  SDK checks content-type: NOT problem+json → no match
          → no KeyConfigMismatchError → no re-attestation → repeat forever
```

## Fix

When the upstream EHBP response is a key-config 422, pass it through **unmodified** — original status, `application/problem+json` content type, and raw body. This is what the tinfoil SDK and routstr-sdk already handle natively (re-attest + retry once).

### Bearer auth path (`forward_ehbp_request`)
```
forward_ehbp_request
  forward_with_trailer → resp
  if resp.status != 200:
+   if is_key_config(resp):
+     return passthrough_response(resp)   ← was: raise UpstreamError
    raise UpstreamError(...)
```

### X-Cashu path (`forward_ehbp_x_cashu_request`)
Same passthrough, plus full token refund before returning (request was never processed by the enclave).

## What changed

```
routstr/upstream/ehbp.py
+ _is_ehbp_key_config_response()     # 422 + problem+json + key-config URN
+ _passthrough_key_config_response() # raw passthrough with EHBP headers
  forward_ehbp_request()             # use passthrough on key-config
  forward_ehbp_x_cashu_request()     # same + refund

tests/unit/test_tinfoil_integration.py
+ TestIsEhbpKeyConfigResponse        # 9 tests
+ TestPassthroughKeyConfigResponse   # 3 tests
```

## Testing

56 tests pass (12 new). Key scenarios covered:
- Genuine key-config 422 detected
- Non-422, wrong content-type, wrong URN, empty/invalid body all correctly rejected
- Passthrough preserves status, content-type, body, and EHBP headers
- Passthrough filters out upstream-internal headers (server, x-request-id)

## Companion PR

routstr-sdk side: [link to SDK PR] — broadens the SDK's detection to also match the proxy-wrapped body as a fallback, covering clients behind older proxies that don't have this fix yet.

### DIFF
--- a/routstr/upstream/ehbp.py
+++ b/routstr/upstream/ehbp.py
@@ -46,7 +46,7 @@ from ..wallet import (
     recieve_token,
     send_token,
 )
-from .tinfoil_trailer import forward_with_trailer
+from .tinfoil_trailer import TrailerResponse, forward_with_trailer
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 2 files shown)


sh1ftred:
Companion SDK PR: https://github.com/Routstr/routstr-sdk/pull/52 — broadens the SDK's key-config detection to also match proxy-wrapped bodies, covering clients behind proxies that don't have this fix yet.

9qeklajc:
first one is I think the most important to fix and the tests

### forwarding `Ehbp-Response-Nonce` breaks the stock tinfoil SDK

`_passthrough_key_config_response()` forwards the nonce header on purpose:

```python
if k.lower() in ("ehbp-response-nonce", "content-length"):
    passthrough_headers[k] = v
```

But the `ehbp` client checks for the nonce _before_ it checks for a key-config error:

```js
// ehbp@0.3.2/dist/esm/client.js
static async shouldDecryptResponse(response) {
    if (response.headers.has(PROTOCOL.RESPONSE_NONCE_HEADER)) {
        return true;   // goes to decrypt, key-config check never runs
    }
    await Transport.checkKeyConfigMismatch(response);
```

So if the enclave ever sends a nonce on the 422, the tinfoil SDK tries to decrypt the plaintext error, fails, and never throws `KeyConfigMismatchError`. No re-attestation. Same permanent-failure loop, different error.

routstr-sdk is fine (it checks key-config first, `TinfoilSecure.ts:318`). But the tinfoil SDK is the one the PR says it also fixes.

The nonce has no meaning on an unencrypted error body. Just drop it:

```python
passthrough_headers: dict[str, str] = {"content-type": "application/problem+json"}
```

And flip `test_ehbp_nonce_header_forwarded` to assert the header is absent.

---

### Content-type check is a substring match

```python
if "application/problem+json" not in ct:
```

The SDK and the `ehbp` lib both parse the media type exactly. This version also accepts `text/html; x=application/problem+json`.
Since this check decides whether we send the raw upstream body to the client, tighten it:

```python
media_type = ct.split(";", 1)[0].strip()
if media_type != "application/problem+json":
    return False
```

---

### No tests at the two call sites

The helper tests are good. But the actual behaviour the fix depends on isn't covered:

- **Bearer path:** the early `return` skips the `UpstreamError` handler. The reservation is still released, but only because `proxy.py:791` doesn't have 422 in its `should_retry` list.
  Nothing pins that. A test asserting the reservation is released on a key-config 422 would protect it.
- **X-Cashu path:** nothing asserts the refund is sent and `X-Cashu` is set on the passthrough response.

PR #737, opened 2026-09-17
fix: rewrite rejected developer message role to system on retry

sh1ftred:
## What we did

Some upstream routers reject `role: "developer"` messages (Codex-style system prompts) with a hard 400, while others accept/normalize them — causing intermittent failures for identical requests depending on which upstream the balancer picked.

**The problem, seen in captured traffic:**

```
client ── developer role ──> routstr ──balancer──┬──> upstream A (OpenRouter)   ──> 200 OK
                                                 └──> upstream B (strict serde) ──> 400
                                                      messages[0].role: unknown variant `developer`,
                                                      expected one of `system`, `user`, `assistant`,
                                                      `tool`, `latest_reminder`
```

**The fix — one new reactive corrector in the existing request-correction pipeline:**

```diff
  DEFAULT_CORRECTORS: tuple[Corrector, ...] = (
    strip_unsupported_param,
+   demote_unknown_message_role,
  )
```

- Keys off the upstream's own error wording (`messages[N].role: unknown variant \`developer\``, plus looser wordings like "role \`developer\` is not supported")
- Renames every `developer` message to `system` — lossless, `developer` is OpenAI's newer spelling of `system`
- Same-upstream retry via the existing correction path; label `role-developer-system` guarantees it applies at most once
- Roles with no safe fallback (e.g. `latest_reminder`) decline and the original error propagates

**Why reactive, not proactive:** the corrector only engages on a rejected body (nothing was billed), so the retry is cheap — and upstreams that do understand `developer` keep the original body untouched.

## Files

- `routstr/upstream/request_correction.py` — new corrector + role-detection helpers
- `tests/unit/test_request_correction.py` — 9 tests: exact captured error text, multi-message rewrite, no-mutation, no-op declines, end-to-end via `correct_request`

## Verification

- 28/28 unit tests pass in `test_request_correction.py`
- ruff + mypy clean
- Reproduced from real captured traffic: 16 requests to the strict upstream all 400'd; every request to normalizing upstreams returned 200


sh1ftred:
fixed client side.

PR #738, opened 2026-09-17
feat: identify generic upstream host in provider field and add masked provider_url

9qeklajc:
## Summary
- Generic upstreams stamped `provider` as bare `"generic"`, unlike OpenRouter's `openrouter:<sub-provider>`. Now they report `generic:<upstream host>` (upstream-reported `provider` still wins).
- New `provider_url` response field on every provider, carrying the upstream base URL.
- Both go through `public_provider_url()` (same masking as model paths), so private IPs and explicit ports surface as `localhost`.


sh1ftred:
LGTM 🚀

- Generic override → host → base-class namespacing composes cleanly, and re-stamping stays idempotent via the `generic:` prefix check (important for streaming).
- `provider_url` is covered on both paths: base `_apply_provider_field` (generic via `super()`) and OpenRouter's override.
- Masking via `public_provider_url()` keeps private IPs / explicit ports from leaking through either field, consistent with model paths.
- Nice test coverage: precedence, idempotency, and the masking case.

PR #739, opened 2026-09-17
feat: cap single Lightning payout at MAX_PAYOUT_SAT

9qeklajc:
Adds a `MAX_PAYOUT_SAT` setting (default `250000`) that caps the amount sent in a single periodic Lightning payout. Balances above the cap are paid out over multiple cycles.

- New `max_payout_sat` setting in `routstr/core/settings.py`
- `_payout_mint_and_unit` sends `min(available_balance, max_payout_sat)` (msat-converted for msat units)
- Unit test for the cap
- Documented in `docs/provider/configuration.md`

PR #741, opened 2026-09-17
perf: cut mint requests per payout cycle

9qeklajc:
Periodic payout was hitting mint rate limits (seen with Minibits). Two reductions:

- Reload only proofs from the local DB before payout instead of `force_reload=True`, which re-fetched `/v1/info`, `/v1/keysets` and every `/v1/keys/{id}` from the mint each cycle. Keysets stay on their normal refresh interval. New `get_wallet(force_reload_proofs=True)`.
- Skip the wallet entirely (no `checkstate`, no settle delay) when unreserved proofs sum to `<= MIN_PAYOUT_SAT`; previously only an empty proof list short-circuited.


9qeklajc:
Closing at the maintainer’s request: this PR’s changes have been incorporated into a combined local Stage 1 fix for #761 alongside #739. The replacement includes the payout cap, proof-only reload and low-balance short-circuit, plus bounded adaptive read-only checkstate batching for HTTP 413/500, strict proof-state validation, and per-mint configuration. 303 relevant unit tests passed; lint and changed-source type checks passed. Important: the replacement is intentionally still uncommitted/unpushed, not merged or deployed. Automatic consolidation and the separate recovery redesign are deferred; existing melt recovery is unchanged. Keep #761 open until the replacement is reviewed and merged.

9qeklajc:
The combined replacement is now available in #763, which supersedes this PR.

PR #742, opened 2026-09-18
stats: record completed request usage after settlement

ashen0x:
Completed request statistics need a durable source after payment settlement. This adds a private outcome ledger and records completed requests, charged revenue, model identifiers and token usage with reported, estimated or missing provenance. It includes zero-charge completions and keeps pricing, refunds and payment ordering unchanged.

This is Core stats 1/4. It provides the ledger and settlement capture; the later runtime PR activates collection. Public reports and operator controls are separate layers.

Validation: 350 focused ledger, migration, settlement and failover tests passed. Whole-tree Ruff and Mypy passed, along with diff whitespace checks.

Known boundary: a generic X-Cashu stream that disconnects before successful completion still has no completion record, even when prepaid value is retained. This PR preserves that payment behavior; treatment of those interrupted streams remains a release decision. A backwards clock change combined with a crash is also an unresolved coverage edge case.


9qeklajc:
nice.. moving the stats to db is the right decision 👍 

just from a quick check, it would be great to split some code for example `routstr/core/terminal_outcomes.py` it makes it easier to check .. 
and also not sure about why there is `*_observed`, `*_source`, `..`, in the outcome table ? is it not possible to.. 
input_observer and inputs > 0 fells like it does mean the same thing ?
not sure but are there cases that we get input is present but not output ?

will test more tomorrow

ashen0x:
> nice.. moving the stats to db is the right decision 👍
> 
> just from a quick check, it would be great to split some code for example `routstr/core/terminal_outcomes.py` it makes it easier to check .. and also not sure about why there is `*_observed`, `*_source`, `..`, in the outcome table ? is it not possible to.. input_observer and inputs > 0 fells like it does mean the same thing ? not sure but are there cases that we get input is present but not output ?
> 
> will test more tomorrow

yupp, good point. will remove the extra flags and keep one label per count saying whether it came from provider, was estimated by us, or wasn’t provided, we still need that because “provider reported 0” and “provider didn’t give us a count” are different, and > 0 can’t tell them apart. input and output counts don’t always come together, e.g embeddings only report input usage. will also see if large files can be split cleanly so its easier to review and maintain.

ashen0x:
pushed the changes. removed the extra flags like i said, so now each count only has one label saying whether it came from provider, was estimated by us, or wasn't provided. also split terminal_outcomes.py, the background writer part is now in its own file terminal_outcome_writer.py so its easier to review, the code itself is same. rebased everything on latest main too.

migrations/versions/c8e4a1f2b3d5_add_terminal_outcome_ledger.py
```diff
@@ -0,0 +1,133 @@
+"""add terminal outcome ledger
```
9qeklajc: migration should be recreated

routstr/core/db.py
```diff
@@ -788,6 +800,104 @@ class ReservationRelease(SQLModel, table=True):  # type: ignore
     created_at: int = Field(default_factory=lambda: int(time.time()))
 
 
+class TerminalOutcome(SQLModel, table=True):  # type: ignore
```
9qeklajc: Blocker: this makes the ledger non-private.

`outcome_id` is the request id (`base.py:184` <- `middleware.py:105`). It is also returned to the client as `x-routstr-request-id` and logged on every line. The writer uses `create_session`, so this table is in the same DB as `cashu_transactions`, which stores `request_id` next to `api_key_hashed_key`.

```sql
SELECT o.*, c.api_key_hashed_key FROM terminal_outcomes o
JOIN cashu_transactions c ON c.request_id = o.outcome_id;
```

Fix: `outcome_id=uuid.uuid4().hex`, or HMAC the request id if correlation is needed. Cheaper now than after rows exist.

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+            self._loss_day = previous_loss_day or _utc_day_from_ms(self._now_ms())
+            self._clear_runtime()
+            logger.critical("Terminal outcome writer could not start", exc_info=True)
+            return False
+        self._accepting = True
+        self._task = self._new_task()
+        return True
+
+    async def stop(self, *, timeout: float = 5.0, close_coverage: bool = False) -> bool:
+        closed_day = _utc_day_from_ms(self._now_ms()) - timedelta(days=1)
```
9qeklajc: Start and stop on the same UTC day gives `end < start`.

`_ensure_epoch` (:306) sets `coverage_start_day` to tomorrow, this closes at yesterday. Observed on a fresh DB: `(epoch 1, start 2026-09-24, end 2026-09-22)`. No CHECK forbids it (`db.py:848-857`).

Fix: clamp `closed_day` to `coverage_start_day - 1`, or skip the close when `closed_day < coverage_start_day`.

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+        session: AsyncSession,
+        current: TerminalOutcomeEpoch,
+        lost_day: date,
+    ) -> int | None:
+        transition = await session.exec(  # type: ignore[call-overload]
+            update(TerminalOutcomeEpoch)
+            .where(col(TerminalOutcomeEpoch.epoch) == current.epoch)
+            .where(col(TerminalOutcomeEpoch.current_slot) == 1)
+            .values(
+                coverage_end_day=lost_day - timedelta(days=1),
```
9qeklajc: Two bugs here.

1. `coverage_end_day = lost_day - 1` is not clamped to `coverage_start_day`.
2. The UPDATE below rewrites every closed epoch with `coverage_end_day >= lost_day`, truncating already-covered days. `declare_loss` keeps the minimum day (:255), so the oldest stale day rewinds prior epochs.

Repro: closed epoch 0 = 08-20..08-28, current epoch 1 start 08-31, then `_rotate_epoch(lost_day=2026-08-25)` -> epoch 0 truncated to 08-24, epoch 1 left at start 08-31 / end 08-24.

Fix: clamp, and bound the UPDATE to epochs that overlap `lost_day`.

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+
+    async def _reconcile(self, queued: _QueuedOutcome) -> _PersistResult:
+        try:
+            async with self._session_factory() as session:
+                existing = await session.get(TerminalOutcome, queued.outcome_id)
+        except asyncio.CancelledError:
+            raise
+        except Exception:
+            return _PersistResult.RETRY
+        if existing is None:
```
9qeklajc: `_reconcile` returns RETRY when the row is absent and `_persist` retries on a 1s sleep with no bound. Missing table -> permanent loop, queue never drains, `stop()` burns its timeout and returns `clean=False`, no loss declared. Bound the attempts and escalate to `declare_loss`.

Two more in this file, multi-worker only:
- `:136` `start()` runs `_recover_unattended_coverage` before `_create_run`, so worker B checks liveness inside A's window, sees no live run and rotates away valid coverage. Insert the run row first.
- `:478` after losing the `_stage_rotation` race the loop re-reads the new current epoch and rotates it with the same `lost_day`: inverted window plus one extra epoch per process.

Untested: `:312`, `:492`, `:536`, `:586`, heartbeat `:693`, done-callback `:747`. A non-terminating retry would ship green.

routstr/core/terminal_outcomes.py
```diff
@@ -0,0 +1,150 @@
...
+    input_source: str | None = None
+    output_source: str | None = None
+    cache_read_source: str | None = None
+    cache_creation_source: str | None = None
+
+
+terminal_outcome_writer = TerminalOutcomeWriter()
+
+
+def record_terminal_outcome(
```
9qeklajc: Paths that settle without recording or declaring a loss: `base.py:1410`, `base.py:264`, `ehbp.py:899`, `ehbp.py:897` (`released is False`), `auth.py:1260`, `:1300`, `:1384`, `:1447`, `:1551`.

Per-path `else` will not hold and the failure is silent by construction. Make `_TerminalOutcomeState` a context manager that declares a loss on exit when neither fired.

routstr/upstream/base.py
```diff
@@ -102,6 +111,222 @@ async def _finalize_and_close_stream(
...
+    try:
+        async for chunk in stream:
+            yield chunk
+        state.mark_success()
+    except BaseException:
+        state.mark_transport_failure()
+        raise
+
+
+async def _track_x_cashu_generic_stream(
```
9qeklajc: Same as `:1410`, and this path never refunds, so the full redeemed ecash is retained. `settlement_context(require_success=True)` returns `None` on disconnect: outcome dropped, no loss declared.

routstr/upstream/base.py
```diff
@@ -1180,6 +1407,7 @@ async def finalize_db_only() -> None:
                                 model_obj,
                                 self.provider_fee,
                                 reservation_snapshot,
+                                terminal_outcome=outcome_state.settlement_context(),
```
9qeklajc: Charged, but no row and no loss marker.

`mark_transport_failure()` (:1659) runs before the shielded `finally`, so `settlement_context()` returns `None` here. `adjust_payment_for_tokens` still charges the key. The day still reads as fully covered.

The PR notes the missing row. The missing marker is separate: a known gap is reconcilable, a silent one is not.

```python
except BaseException:
    outcome_state.mark_transport_failure()
    mark_terminal_outcome_loss("stream_transport_failure")
    raise
```

routstr/upstream/base.py
```diff
@@ -2622,19 +2911,28 @@ async def _forward_x_cashu_messages_via_litellm(
             self._fold_cache_into_input_tokens(response_json["usage"])
 
         response_headers: dict[str, str] = {}
+        refund_amount_sent = 0
```
9qeklajc: Not a funds bug. Recording the check so it does not get re-raised.

This single `if` guards the refund, the ledger write and the loss marker. If `get_x_cashu_cost` returned `None` the ecash would be kept with no row and no marker, while siblings at `:4434`, `:4610`, `:5563` declare a loss.

It cannot return `None`: `calculate_cost` returns `CostData | MaxCostData | CostDataError`, the `match` at `:4160-4191` covers all three, and the error arm returns `unpriced_cost(...)`, a truthy zero-msat `CostData`. Verified with a litellm response carrying no `usage`: full refund, header set, row written.

So `return None` at `:4191` is unreachable. Delete it, otherwise widening that return type later turns this nesting into a real funds bug.

Separately, line 2901 is unguarded: a raise out of `calculate_cost` would retain the ecash with no marker. No known trigger.

routstr/upstream/ehbp.py
```diff
@@ -743,7 +883,29 @@ async def finalize_ehbp_max_cost_payment(
...
+            reservation.reserved_msats,
+            idempotent_success=False,
+        )
+    except BaseException:
+        if terminal_outcome is not None:
+            mark_terminal_outcome_loss("ehbp_release_commit_ambiguous")
+        raise
+    if not released:
+        await _stop_reservation_heartbeat(reservation.release_id)
+    if released and terminal_outcome is not None:
```
9qeklajc: This passes the bare `terminal_outcome`; the sibling at `:812` wraps it first. A request with a valid `X-Tinfoil-Usage-Metrics` header that fails the priced path writes real `input_tokens`/`output_tokens` with `input_source`/`output_source`/`pricing_source` = `"missing"`. Consumers filtering `source="reported"` undercount.

```python
record_terminal_outcome(
    _context_with_presence(terminal_outcome, _tinfoil_usage_presence(usage_data)),
    ...
)
```

Reproduced: 120 input / 30 output tokens with both sources `None`.

routstr/core/db.py
9qeklajc: I'd also move this to a different file

routstr/core/db.py
```diff
@@ -788,6 +800,104 @@ class ReservationRelease(SQLModel, table=True):  # type: ignore
...
+        ),
+    )
+
+    epoch: int = Field(primary_key=True, sa_type=BigInteger)
+    coverage_start_day: date = Field(sa_type=Date)
+    coverage_end_day: date | None = Field(default=None, nullable=True, sa_type=Date)
+    current_slot: int | None = Field(default=None, nullable=True)
+
+
+class TerminalOutcomeWriterRun(SQLModel, table=True):  # type: ignore
```
9qeklajc: same https://github.com/Routstr/routstr-core/pull/742/changes#r4087521099

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+            except asyncio.QueueEmpty:
+                return
+            else:
+                self.declare_loss(
+                    "queued terminal outcome discarded after continuity loss",
+                    queued.terminal_day,
+                )
+                self._queue.task_done()
+
+    async def _recover_loss(self) -> None:
```
9qeklajc: Blocker: one routine per-request error discards up to 4096 unrelated valid outcomes.

`_recover_loss()` opens with `self._discard_queue()` (:650), which drains the **entire** pending queue (`_QUEUE_SIZE` = 4096) and declares each drained item lost. `declare_loss` is a process-global switch: `_accepting = False`, day degraded, epoch rotated.

It is reached from ordinary per-request failures. `base.py:4440`:

```python
except Exception as e:
    settlement_failed = True
    if outcome_state.settlement_context() is not None:
        mark_terminal_outcome_loss("X-Cashu streaming settlement failed")
```

One mint timeout on one user's refund throws away every already-settled outcome still queued, belonging to unrelated requests, and marks the whole UTC day untrustworthy. Mint and network timeouts are routine; the blast radius is a day of ledger data.

Fix: drain and persist the proven rows before rotating, and scope a per-request settlement failure to a single-row gap marker. "This request's outcome is unknown" and "this process lost continuity" are currently the same switch and shouldn't be.

routstr/auth.py
```diff
@@ -1188,8 +1201,48 @@ async def _accumulate_fee(total_cost_msats: int) -> None:
...
+            context = replace(
+                terminal_outcome,
+                pricing_source=cost.pricing_source,
+                input_source=cost.input_source,
+                output_source=cost.output_source,
+                cache_read_source=cost.cache_read_source,
+                cache_creation_source=cost.cache_creation_source,
+            )
+            if terminal_usage is not None:
+                recorded_usage = {**(recorded_usage or {}), **terminal_usage}
```
9qeklajc: Blocker: merging two vendor usage dialects writes a token count matching neither fragment.

Executed against the real normalizer:

```
normalize_usage({'input_tokens': 50, 'output_tokens': 7, 'estimated': True,
                 'prompt_tokens': 120, 'cache_read_input_tokens': 100})
  -> input_tokens=20 output_tokens=7 cache_read_tokens=100
```

The `prompt_tokens - cache` branch wins and the explicit `input_tokens=50` is discarded. So on the litellm-converted messages streaming path a row can carry `input_tokens=20` (from neither fragment) beside a `revenue_msats` billed from a different count, making the row's implied per-token rate wrong. Silent: no error, no gap flag.

The presence labels are meanwhile computed from `terminal_usage` alone, so provenance and counts describe different dicts.

Fix: don't overlay dialects. Normalize one fragment and record its own `NormalizedUsage`, falling back per bucket rather than by dict spread.

routstr/core/terminal_outcomes.py
```diff
@@ -0,0 +1,150 @@
...
+    terminal_at_ms: int | None = None,
+    usage: object = None,
+) -> None:
+    """Submit a settled outcome without awaiting storage or raising.
+
+    ``usage`` is the raw upstream usage. It replaces the token counts and is
+    parsed here, so a malformed value can only mark a gap.
+    """
+    try:
+        if usage is not None:
```
9qeklajc: Two competing token inputs; the four explicit kwargs are dead whenever `usage` is passed.

`record_terminal_outcome` takes `input_tokens`/`output_tokens`/`cache_read_input_tokens`/`cache_creation_input_tokens`, then overwrites all four from `normalize_usage(usage)`. `auth.py:1230` and `base.py:239` pass both; the three `ehbp.py` sites pass no `usage`. So the two branches differ by accident rather than design.

`base.py:234` documents the overwrite as intentional ("Stats keep what upstream reported, even where billing did not parse it"), so this is API shape, not discarded billing.

Fix: one input. Every caller already holds `NormalizedUsage`/`CostData` - take the normalized counts and drop the raw `usage` parameter. Where stats want the lenient read, the caller picks that normalizer.

Also: the `from ..payment.usage import ...` on the line below is the only non-top-level import in the file. There's no cycle to dodge - `payment/usage.py` imports nothing from `routstr` (just `math`, `dataclasses`, `typing`, `pydantic.v1`). Hoist it.

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+        self._task = None
+        self._queue = None
+        self._idle = None
+        self._wake = None
+        self._run_id = None
+        self._enabled = False
+        self._accepting = False
+        self._stopping = False
+
+    async def _ensure_epoch(self, now: int) -> None:
```
9qeklajc: Altitude question on the ~400 lines from here to :597.

Fifteen private methods (`_ensure_epoch`, `_rotate_epoch`, `_claim_stale_runs`, `_recover_unattended_coverage`, `_void_served_coverage`, ...) exist to tell a future reader which days are trustworthy, and none of the entry points is exercised: `serving=` is only ever the default `False` (`terminal_outcomes.py:141`), `close_coverage=` likewise (:146), `flush()` (:217) has no production caller.

The repo already solves this shape: `auth.py:734-836` uses a single heartbeat timestamp column plus the stale sweeper task in `main.py` - no status enum, no epoch table.

Worth deciding this before fixing the epoch-window bugs flagged elsewhere in this review: if the machinery is replaced by the reservation-lease pattern, those bugs go away with it rather than needing individual clamps. Suggest keeping queue-to-persist in this PR and adding the epoch table when a reader exists.

routstr/upstream/ehbp.py
```diff
@@ -213,6 +224,45 @@ def parse_tinfoil_usage_metrics(header_value: str | None) -> dict | None:
...
+
+
+def _context_with_presence(
+    context: TerminalOutcomeContext,
+    presence: UsageFieldPresence,
+) -> TerminalOutcomeContext:
+    return replace(context, **presence.sources_dict())
+
+
+def _context_with_served_model(
```
9qeklajc: `_context_with_served_model` pops three keys out of its caller's dict.

The helper returns a new frozen context (good) but `pop()`s `actual_model_identifier`, `served_model_identifier` and `actual_model_unresolved` out of the caller's `cost_info`. It runs at :1101, and `finalize_ehbp_actual_cost_payment` calls it **again** at :815 on that same dict - the second call is a no-op *only because* the first destroyed the keys. Idempotence resting on a hidden mutation crossing two functions.

The pops are load-bearing twice over: :1112 spreads the same `cost_info` into response headers, which is why the keys get stripped. Reorder the calls or add a third and model attribution silently changes.

Fix: :1100 already pops `actual_model` explicitly one line above - pop the stats keys there too, matching the local idiom, make this helper read-only (`.get`), and call it once.

routstr/payment/cost_calculation.py
```diff
@@ -35,6 +40,12 @@ class CostData(BaseModel):
     cache_creation_input_tokens: int = 0
     cache_read_msats: int = 0
     cache_creation_msats: int = 0
+    # Settlement-only metadata must not expand the client cost contract.
+    pricing_source: str = Field(default="missing", exclude=True)
```
9qeklajc: Provenance round-trips through the client cost contract.

These four `*_source` fields plus the new `usage_presence` param threaded through `calculate_cost`, `_empty_cost`, `unpriced_cost`, `_calculate_from_usd_cost` and `_calculate_from_tokens` carry data cost calculation neither produces nor uses. Every recording site already holds the presence object - `auth.py:1223` builds `UsageFieldPresence(input_source=cost.input_source, ...)`, reading straight back out what it passed in.

Fix: keep `pricing_source` (genuinely derived in `_get_pricing_rates`), drop the four `*_source` fields and the `usage_presence` params, and pass presence directly to `record_terminal_outcome` at the three record sites.

routstr/upstream/base.py
```diff
@@ -102,6 +111,222 @@ async def _finalize_and_close_stream(
 CostMetadata = CostData | MaxCostData | dict[str, Any]
 
 
+@dataclass
```
9qeklajc: These 215 lines want their own module.

The seven new top-level defs from here to :328 (`_TerminalOutcomeState`, `_terminal_outcome_context`, `_event_usage_presence`, `_record_x_cashu_terminal_outcome`, `_track_generic_terminal_stream`, `_track_x_cashu_generic_stream`, `_observe_terminal_sse_bytes`) reference nothing else in this file except the `CostMetadata` alias (:111) and the 7-line pure `_cost_field` (:330). No `self.`, no `logger.`, no `calculate_cost`. `_observe_terminal_sse_bytes` is a generic SSE byte-framing parser with zero provider knowledge.

base.py is already 6201 lines; this is the cheapest 215 it will ever shed. Suggest `routstr/upstream/terminal_outcome_tracking.py`, and promoting `_cost_field`/`CostMetadata` into `payment/cost_calculation.py` (where `CostData` lives) so the new module needs no base.py import at all.

It also pays for itself immediately: `ehbp.py:967` and `:1231` currently rebuild `_terminal_outcome_context` inline, `ehbp.py:1408-1418` hand-rolls `_record_x_cashu_terminal_outcome`, and `ehbp._context_with_presence`/`_cost_info_presence` parallel `_event_usage_presence`. All three become imports.

routstr/upstream/base.py
```diff
@@ -3896,6 +4270,7 @@ async def handle_x_cashu_streaming_response(
         request_id: str | None = None,
         model_obj: Model | None = None,
         request_body: bytes | None = None,
+        record_outcome: bool = True,
```
9qeklajc: `record_outcome: bool` is a second channel for a decision the PR already expresses another way.

Eight sibling signatures say "don't record" as `terminal_outcome: TerminalOutcomeContext | None = None` (1350, 1705, 1858, 2159, 2367, 2684, 2798, 2973) with 21 pass-throughs. These three methods (:4273, :4503, :4692) instead take a bool and duplicate `_terminal_outcome_context(request_id, model_obj) if record_outcome else None` at :4309 and :4524. Two mechanisms for one concept in one class.

The bool genuinely varies at exactly one call site, :4952 (`not path.endswith("messages/count_tokens")`).

Fix: drop the param from all three; :4952 passes `terminal_outcome=None if path.endswith("messages/count_tokens") else _terminal_outcome_context(...)`.

routstr/core/terminal_outcomes.py
```diff
@@ -0,0 +1,150 @@
+from __future__ import annotations
+
+import time
+from dataclasses import dataclass
+
+from .logging import get_logger
+from .terminal_outcome_writer import (
```
9qeklajc: The 150/756 split is at the wrong seam.

This facade imports three underscore-private names from its sibling, which is the tell: the two files are one module cut at a non-boundary. It also re-derives at :62-65 a source dict that `UsageFieldPresence.sources_dict()` already builds (`payment/usage.py:64`, used at 8 other sites), and holds `cashu_retained_msats` - Cashu unit arithmetic living in `core/`, duplicating the `unit == "msat" else * 1000` switch at `payment/helpers.py:101` and `wallet.py:109`.

Fix: merge the two modules (each is the other's only consumer) or make the shared helpers public, and move the Cashu math next to the other Cashu amount handling in `payment/`.

routstr/core/db.py
```diff
@@ -788,6 +800,104 @@ class ReservationRelease(SQLModel, table=True):  # type: ignore
...
+        CheckConstraint(
+            "terminal_at_ms >= 0 AND input_tokens >= 0 "
+            "AND output_tokens >= 0 AND cache_read_input_tokens >= 0 "
+            "AND cache_creation_input_tokens >= 0 AND revenue_msats >= 0",
+            name="ck_terminal_outcomes_nonnegative",
+        ),
+    )
+
+    outcome_id: str = Field(primary_key=True)
+    terminal_at_ms: int = Field(sa_type=BigInteger)
```
9qeklajc: These four enum-like columns have no CHECK, though the sibling table added in this same PR does.

The code restricts them to `{reported, estimated, missing}` (`terminal_outcomes.py:79`), and `terminal_outcome_writer_runs` constrains its enum at :869 (`status IN (...)`). Add the matching constraint here and in `migrations/versions/c8e4a1f2b3d5_...py`, in the style already used for `status`.

Two more on this table while it's being recreated:
- `terminal_day` (:821) is a pure function of `terminal_at_ms` via `_utc_day_from_ms`, with nothing enforcing agreement, and the index it anchors serves no query yet.
- The `_at_ms` millisecond columns diverge from every other model in this file, which stores int epoch **seconds** via `default_factory` (cf. `ReservationRelease.created_at`, :788). None of the three new tables has a `created_at`. Internally consistent, so convention rather than bug - worth a line in the model docstring either way.

routstr/core/terminal_outcome_writer.py
```diff
@@ -0,0 +1,756 @@
...
+            return _PersistResult.RETRY
+        if existing is None:
+            return _PersistResult.RETRY
+        return (
+            _PersistResult.STORED
+            if _same_outcome(existing, queued)
+            else _PersistResult.CONFLICT
+        )
+
+    async def _persist_once(self, queued: _QueuedOutcome) -> _PersistResult:
```
9qeklajc: The queue never actually batches.

`_persist_once` opens a session and commits one row per outcome, and `_run` (:713) pops one at a time, so the 4096-deep `asyncio.Queue` buys latency, not batching.

`_heartbeat` (:692) additionally runs `_recover_stale_runs` - that's `_claim_stale_runs` + `_recover_pending_runs`, each its own session and table scan - every 15s even when completely idle, and `_persist` repeats the same pair on every 1s retry (:639) while head-of-line blocking the queue.

Fix: drain up to N items with `get_nowait` and insert them in one session/transaction - `core/usage_analytics_store.py:608 _apply_updates_locked` is the existing batched-write pattern in this repo. Run stale-run recovery on startup, or on a much longer interval than the heartbeat, rather than on every tick and every retry.

9qeklajc (changes requested):

